PR #41, opened 2014-04-20
Added _updated and _error attributes to jobs.

lperrin:
Hey Manuel,

We're working on a small web interface to admin bull jobs (like kue). It will be a totally independent module so bull code can stay minimal. We just need a few more things in bull:
- `job._updated`: last time the job was moved to a different state.
- `job._error`: error message of failed jobs
- `job.retry()`: move the job back to the "wait" queue.


manast:
will close this PR since it does not seem to be relevant anymore

PR #72, opened 2015-01-08
Retry jobs

albertjan:
Adds the ability to retry a job has something to do with #8. I know about #47. This is a much simpler it just adds a function to retry a job. Because it's harder to retry a job when using oxen due to the message.

PR #75, opened 2015-01-12
Adds retry test back in and fixes race conditions

albertjan:
in before hooks. See comments in #72

### DIFF
--- a/test/test_job.js
+++ b/test/test_job.js
@@ -3,17 +3,17 @@ var Queue = require('../lib/queue');
 var expect = require('expect.js');
 
 
-describe('Job', function(){
+describe('Job', function () {
   var queue;
 
-  before(function(done){
+  before(function (done) {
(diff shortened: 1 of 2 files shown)


albertjan:
Hmm the tests fail but I'm quite sure there's not much wrong with it.

manast:
now that we have delayed functionality, retry should delay the job at every retry attempt, maybe with some time exponential function.

albertjan:
Automatically? Or still triggered? Or maybe with an option?

manast:
I think an option, like retryTime or something similar. or an object for the whole retry functionality:
```javascript
retry: {
 delay: 1000,
 attempts: 5,
}
```

manast:
I think delayed retries is one of the major features missing in bull. If you dont have time I can give this pull request a shot soon.

manast:
forget my previous comment, this is about the test :).

PR #77, opened 2015-01-22
Move the auto-restart of queue inside process.

### LINKED ISSUE #76: Cannot call method 'bind' of undefined
Error append in https://github.com/OptimalBits/bull/blob/25480022f1c01b782afb0b81c68bad65b2933615/lib/queue.js#L287.

One of the root cause if : https://github.com/OptimalBits/bull/blob/25480022f1c01b782afb0b81c68bad65b2933615/lib/queue.js#L58

Step to produce error:
- Create a queue in order to add a new job to process.
- Do not call queue.process() (worker that process job are separated from code that create job)

- Close connection between redis and client.
- This line https://github.com/OptimalBits/bull/blob/25480022f1c01b782afb0b81c68bad65b2933615/lib/queue.js#L68 will detect connection end and will try to rerun the queue. But this last one have no handler on client side, it's only use to push new job.

lchenay:
Move the auto-restart of queue inside process.
This will prevent calling run() when no handler in case of redis error.
Fix #76

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -54,18 +54,10 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
 
   var _this = this;
 
-  var runQueueWhenReady = function(){
-    _this.bclient.once('ready', _this.run.bind(_this));
-  };
-
   // bubble up Redis error events
   [this.client, this.bclient].forEach(function (client) {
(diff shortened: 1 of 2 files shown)


manast:
Thanks. Good catch and solution.

PR #79, opened 2015-01-23
Fix #78, wrong function called on type delayed and completed

### LINKED ISSUE #78: Since delayed feature, payload of job are not deleted on job.remove().
Since delayed feature, payload of job are not deleted on job.remove().

And error appear:
"ERR Error running script (call to f_c674f148868d034fdefc258b87ec4bf9201c0919): @user_script:8: WRONGTYPE Operation against a key holding the wrong kind of value"

lchenay:
Fix #78


manast:
thanks. Seems like a unit test for remove jobs is missing, I will add it ASAP.

PR #81, opened 2015-01-25
Add priority queue fonctionality

lchenay:
Adding priority queue functionality to Bull.

Priority have been design to process job of higher priority more often than lower priority. It will not process all critical before all hight before all normal... This design choice allow us to save many check and Redis request. But it work like a charm and job with hight priority will be processing significantly faster than other.

It's works like normal queue, except that the ```add``` function allow an ```opts.priority``` options with one of following value: ```[low, normal, medium, hight, critical]```.

Due to current design of Redis connection, this priority queue will increase Redis connection by 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Useful patterns
 Bull can also be used for persistent messsage queues. This is a quite useful
 feature in some usecases. For example, you can have two servers that need to
 communicate with each other. By using a queue the servers do not need to be online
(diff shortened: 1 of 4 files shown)


lchenay:
LGTM for merge.

manast:
Looks great. I think that a simple fix to reduce the number of connections is simply to separate the client from the worker, the worker needs 2 connections instead of 3 as it is now. At least it will go down from 15 to 10 connections.

PR #84, opened 2015-01-25
Add concurrency feature.

lchenay:


manast:
I have merged this, but there is one unit test failing, the one that assumes that after calling pause, it will wait for all the current jobs to be processed before the promise is resolved. This is currently not working, since pause is now a global pause that pauses all the queues in the system. It makes a lot more sense since now is the client that pauses the queue and it pauses all the workers not just his own instance, which is not so useful as I see it. I will try to fix the unit test based on the new pause functionality...

luin:
Really looking forward to this feature.

manast:
Why looking forward when it has been available for weeks? :)

PR #86, opened 2015-01-26
update remove storedprocedure

albertjan:
should remove from "active", "wait", "delayed" and "paused" when it's not on "failed" or "completed". it was always deleting from the ZSET.    And it also used the wrong key. It might be a good idea to add some more unit tests for this function.


manast:
thanks for the fix. It seems as if there is a glitch on the unit test covering the delayed functionality, thats why the test did not pass.

PR #87, opened 2015-01-26
publish to new jobs channel to wake oxen up

albertjan:


albertjan:
The tests seem a bit brittle atm though. I might have some time this week to have a look at that.

PR #94, opened 2015-02-01
Fix lock taking mechanism

### LINKED ISSUE #91: Jobs processed more than once
We're using bull on multiple machines connected to one Redis instance processing long jobs.

Starting up multiple machines (3 in total), occasionally see the same job processed on 2 machines – this only seems to happen if a machine is brought online later then the others, when it comes online it picks up a job that another machine is already working on or completed.

albertjan:
I don't know if this fixes #91. But this does fix a race condition where two queue's were getting notified by redis there is a new job. and both getting a lock on it. Because it's using renewLock to get the lock the first time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+
+*.rdb
--- a/lib/job.js
+++ b/lib/job.js
@@ -77,6 +77,7 @@ Job.prototype.takeLock = function(token, renew){
   if(!renew){
     args.push('NX');
   }
+
   return this.queue.client.setAsync.apply(this.queue.client, args).then(function(result){
     return result === 'OK';
   });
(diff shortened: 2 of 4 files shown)


albertjan:
maybe the test should be changed to check that no more than __one__ queue has processed the job.

albertjan:
Was a bit more intricate then I thought. Because the jobs that are delayed need to be moved there outside of the processjob. And the subscription to the delayed channel didn't happen until after the job was delayed, causing a test to fail because the delayed queue still contained a job. It might be a good idea to tell users to wait for the `ready` event, before adding tests.

albertjan:
I don't know why the test is failing on travis I haven't changed anything in that part. :smile:

albertjan:
Ah so you'll remove the `Promise.resolve()`?

manast:
Yes  can do it. :)

albertjan:
I added a commit to this PR but you had already merged it :smile:

PR #96, opened 2015-02-03
adds stacktrace on error in handler

### LINKED ISSUE #95: Storing the error/stacktrace when a job fails
Hey @manast 

Is there any chance that if an error occurs/is passed to `done()`, that the error could be stored on job so that ui's (such as matador) could display the error and stack trace later to better help find and debug what caused the error?

Thanks

albertjan:
Fixes #95


manast:
cool :)

PR #101, opened 2015-02-13
Emit an event when a job is removed

aslakhellesoy:
I'm building a wrapper for bull that allows consuming status with an [EventSource](http://dev.w3.org/html5/eventsource/). Having this event would make it possible to detect when a job is removed.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -190,6 +190,7 @@ Job.prototype.remove = function(){
       return queue.toKey(name);
   });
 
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
Looks like I need to rebase with master...

aslakhellesoy:
Ok, rebased. Travis failed, but it seems unrelated. If you think this looks good I'll update docs too.

PR #105, opened 2015-03-03
FIX: Using Async version of hset when saving the stacktrace

seriousben:
Fixes using the async/promisified version of hset. Right now, redisClient would emit an error when saving the stacktrace.


seriousben:
@manast Can you have a look at this? the problem was causing a crash for me.

PR #110, opened 2015-03-22
Implementation of tags to jobs

saromanov:
Hi, guys!
I implemented tags for jobs, and its looks something like:

``` javascript
var videoQueue = Queue('video transcoding', 6379, '127.0.0.1');
videoQueue.process(function(job, done){
  videoQueue.getByTag('active', 'bar').then(function(jobs){
        console.log("Active jobs: ", jobs);
    })

  job.progress(42);
  done();
});
videoQueue.add({video: 'http://example.com/video1.mov'}, {tags:['bar']});
videoQueue.add({video: 'http://example.com/video2.mov'});
videoQueue.add({video: 'http://example.com/video3.mov'}, {tags:['foo','bar']});
videoQueue.add({video: 'http://example.com/video4.mov'}, {tags:['foo']});
```

Where, signature of getByTag is `getByTag(type of jobs ['active','delayed',..], tag)`

In general, we can group some of jobs by tag, get it during the work of queue on some of state(progress, paused, failed...) and process it.
What do you think about this?

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -24,6 +24,8 @@ var Job = function(queue, jobId, data, opts){
   this.delay = this.opts.delay;
   this.timestamp = opts.timestamp || Date.now();
   this.stacktrace = null;
+  var tags = opts.tags;
+  this.tags = Array.isArray(tags)?tags:[tags] || [];
 };
 
 Job.create = function(queue, jobId, data, opts){
@@ -49,7 +51,8 @@ Job.prototype.toData = function(){
     opts: JSON.stringify(this.opts || {}),
     progress: this._progress,
     delay: this.delay,
-    timestamp: this.timestamp
+    timestamp: this.timestamp,
+    tags: this.tags
   };
 };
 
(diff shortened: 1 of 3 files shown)


manast:
sorry for the late reply. I would need to think more about it. I guess this is mostly for making UIs on top of bull, or are there any other important cases?

saromanov:
As i said, it seems to me, that the main use case it grouping jobs by some property and it would be better for manage jobs. For example, we have a 3 jobs with tag "foo", if some of this jobs are faild(for example second job), we can stop remaining jobs(third). Yes, this feature is about UI.

grimborg:
I'd love to have this feature. My use case is filtering tasks by customer so that I can display to each customer their pending tasks. Otherwise I have to fetch all the tasks and filter by hand...

bradvogel:
@manast this PR can probably be closed (and replaced with a new github issue?)

manast:
I like this idea, but needs to be implemented in a different way, so that jobs can be retrieved by tags much faster. So, yes maybe a new issue for this would be in place. I will do it when I get some time.

PR #116, opened 2015-04-13
Add timeout to blocking call

albertjan:
In some edge cases we were seeing clients that haven't been cleaned up correctly and leave a remaining blocking call. 

This means the queue won't be reactivated when the next job is added. The job ends up as stalled jobs immediately, because the job is added to the wait list and no handler is called for it. 

It has an added bonus bull checks for stalled jobs every 5 seconds this way. :smile:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,7 +10,7 @@ gulp.task('lint', function (){
   return gulp.src([
     './lib/job.js',
     './lib/queue.js',
-    '!./test/**'
+    './test/**'
     ])
     .pipe(eslint({
       rules: {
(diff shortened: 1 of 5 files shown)


albertjan:
Tests do not fail on my machine. And surely have nothing todo with the changes. I will investigate if I can make the tests a bit less brittle.

albertjan:
@manast I've changed it around so we don't have to check for `null` everywhere. I've done it with a `reject` instead of a `cancel` because using that would have meant a much bigger change. I've also looped the `proccessStalledJobs` into the mix, so stalled jobs get handled without a restart.

seriousben:
this would be useful :)

manast:
well done!

albertjan:
Thanks :smile:

PR #121, opened 2015-04-24
Support for custom client

rossimo:
Adds support for custom clients connections, i.e., using https://github.com/ortoo/node-redis-sentinel.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -61,6 +61,14 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisDB = redisOpts.DB || redisDB;
   }
 
+  function createClient() {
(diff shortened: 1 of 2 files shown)


manast:
Thanks. It seems to be some conflicts with current master. I will have to resolve them before I can merge this pull request.

lib/queue.js
```diff
@@ -61,6 +61,14 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisDB = redisOpts.DB || redisDB;
   }
 
+  function createClient() {
+    if(redisOptions !== undefined && redisOptions.createClient !== undefined){
+      return redisOptions.createClient();
```
richthegeek: This should ensure the resulting client is fully promisified before returning.

PR #122, opened 2015-04-30
Add Redis server required authentication

fginioux:
Hey, 

I change the Queue object for supporting authenticated Redis connection. I need this because I try to used the lib in an Heroku application for task managing. I don't find how could i make an authenticated connection. So check the code and validate it or give me feedback. I want to used the Bull lib. and not my version. Thank's. Fred

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,43 @@ if(cluster.isMaster){
 }
 ```
 
+Authenticated redis client connection
+---------------
+
+Bull can also be used with a Redis server required authentication
+like in Heroku (or other hosting) application hosting. Set-up a Redis
+addons and get configs information. You can deploy bull queue with
+authentication credentials. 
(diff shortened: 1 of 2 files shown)


manast:
This is not needed. You can pass any redis options already to the Bull constructor, just pass 'auth_pass'. check the list of options here: https://github.com/mranney/node_redis

fginioux:
Thanks for the feed back. I read the documentation but seems you too fast ! I used this code for create the queue instance : 
```javascript
var queueOptions = {
  host: '127.0.0.1',
  port: 6379,
  opts: {
    auth_pass: 'my-password'
  }
};

var myQueueAuth = Queue('my-queue-id', queueOptions);
// Works fine !!!!!!
```
Sorry & Thanks again.

PR #123, opened 2015-05-12
Documented the close function

ryan-sandy:
I noticed the close function was not documented in the readme. I documented it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,7 @@ listened by some other service that stores the results in a database.
 * [Queue##resume](#resume)
 * [Queue##count](#count)
 * [Queue##empty](#empty)
+* [Queue##close](#close)
 * [Job](#job)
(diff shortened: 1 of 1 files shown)


seriousben:
Talking about graceful shutdown, dealing with this now. @manast what is the recommended way of doing this.

The pause is not a solution since it pauses the queue processing across all workers.

Is there a way for the close to wait until all processing is done with a timeout?

manast:
@seriousben I think that by just closing the redis connection the queue will stop processing new jobs, so just waiting for the last jobs to finalize would be enough. Of course in the case you have concurrency or delayed jobs this could be a bit more involved. Ideally the code to wait for these jobs could be implemented in bull itself and included in the close method. I do not have so much time for this myself atm.

manast:
@ryan-sandy thanks for the update!

PR #130, opened 2015-06-16
Added a clean function to the queue

ryan-sandy:
This clean function allows you to remove jobs in the queue that are older than a grace period.


manast:
Pretty good. Thanks!

PR #136, opened 2015-06-29
don't call processJobs twice.

albertjan:
it's called in processJobs allready


albertjan:
We'll have to look at the tests. They're still v. brittle.

manast:
Ideally we should delete the bull queue completely before every test run. I will think about something...

PR #137, opened 2015-06-29
Bug fix and improvements

prabhu:
- Fixed bug where test was failing when ran for the second time
- Added support to specify a different redis key prefix
- Added support for automatically cleaning up key names
- Added new cleanAll method to repeatedly call clean with different types


manast:
seems we already have all this fixes so I will close this PR.

PR #139, opened 2015-07-10
Document passing a result to process callback

tobie:
I've struggled for quite a while trying to figure out how to pass a result to the complete event, so I thought I'd document this so others wouldn't waste time as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,12 @@ videoQueue.process(function(job, done){
 
   // call done when finished
   done();
-
+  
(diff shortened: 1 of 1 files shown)


manast:
Thank you, I will merge immediately. I also need to add documentation about the possibility of returning promises instead of calling "done".

tobie:
> I also need to add documentation about the possibility of returning promises instead of calling "done".

Oh! I had tried that but somewhat failed to make it work.

PR #143, opened 2015-07-16
Job promotion allows for delayed job to be started immediately

mikob:
Sometimes you need a job to be started in X time at the _latest_ but if condition arise sooner, it can be started immediately. 

Does this seem like the cleanest and safest way to do so? Is there anything that can be done to make this cleaner, or is this good enough? Would be happy to clean this up, if there's something you have in mind.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -210,6 +210,44 @@ Job.prototype.remove = function(){
     });
 };
 
+/**
+  Forces this job occur immediately if it is delayed, by removing it from "delayed"
+  and placing it in "wait"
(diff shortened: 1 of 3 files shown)


mikob:
I've updated this so that if a promoted job happens to be the next one being waited on by the delayTimer, things will still work. And I've added a test case to check this. The tests get stuck on client.incrAsync, I don't know why. Perhaps @manast can help?

manast:
we already have "promote" functionality so I will close this PR.

PR #151, opened 2015-08-14
Ensure close() doesn't cause the application to hang

mosilla:
Previously, calling `close` could result in Redis connections being closed before internal timers depending on those connections have executed or been cancelled. This would cause the application to hang indefinitely.

This commit renames the old `close` method to `disconnect` and adds a new `close` implementation which wraps `disconnect` but defers the call until all outstanding timer handles have been released.

This ensures that `close` is safe to call from anywhere at any time, even from within a job handler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,11 +454,11 @@ shutdown.
 var Queue = require('bull');
 var queue = Queue('example');
 
-var after100Jobs = _.after(100, function () {
+var after100 = _.after(100, function () {
   queue.close().then(function () { console.log('done') })
 });
 
-queue.on('completed', after100Jobs);
+queue.on('completed', after100);
 ```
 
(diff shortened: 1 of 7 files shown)


manast:
thanks for this well thought and high quality patch.

manast:
For some reason a unit test related with the closing is failing:
```
 1) Queue .close should be callable from within a job handler:
     Error: timeout of 6000ms exceeded. Ensure the done() callback is being called in this test.
  ```
Do you have any idea why?

mosilla:
Thanks for your feedback. I've pushed a new commit which fixes the failing test.

albertjan:
@remkoboscker we should check this out. Might solve some of our shutdown issues.

Issue #156, opened 2015-08-23
.empty(), .clean(), .remove() don't do anything


brandonros:
I'm getting hammered with like, 200k keys because I didn't know Bull doesn't do any automatic cleanup.

Now, when I do .empty() on load, the jobs still process.

When I do .clean(5000), I get

```
Unhandled rejection TypeError: Cannot read property 'timestamp' of null
    at /home/brandon/server/node_modules/bull/lib/queue.js:675:19

```

Please, please tell me what I am doing wrong.

Edit: I'm doing job.remove(), checking redis keys, still there. Restart the app, starts reprocessing all of the jobs. Absolute nightmare.

albertjan:
Are the job id's on the `completed` list after it's done? Are you calling the `done` callback or resolving a promise from within your job handler?

In general it doesn't matter there are jobs in the database, bull will only look at the `wait` and `active` lists for jobs to process.

Could you find the job hash that produced the error you mentioned?

`empty` is not meant to clean the queue it's supposed to resolve when all jobs are done processing.

`clean` how ever should empty the queue but it needs the timestamp to determine the age of the job. 

I do think it should have handled the missing timestamp better though.

brandonros:
I am resolving a promise from within my job handler.

Wouldn't it make sense to add an option to expire completed jobs? I've had 10 users create 200k jobs in under 3 weeks

Even when I call .clean() and it succeeds, nothing is actually cleaned.

manast:
The clean method is quite new, it may have some bugs in it, I have to take a closer look to it:
https://github.com/OptimalBits/bull/pull/130

brandonros:
@manast I think I'm going to have to drop Bull from my project. It doesn't look like jobs that fail are being cleanly removed. New instances that spawn seem to pick up the old jobs, and the next thing I know, I'm rate limited by the external API servers. I am sure I am doing something wrong.

manast:
@brandonros I will look into this as soon as I can, unfortunately I have a lot on my plate right now. If you check the issues labeled with questions you can find similar problems as the one you describe and different ways to resolve it.

albertjan:
@brandonros thats absolutely not the experience we have with bull. We've been running with it production for a while now, I've never seen bull pickup a job that had failed earlier.

Are you sure your jobs are being moved to the `failed` or `completed` set after processing?

I'll see if I can spend some time on the `clean` method this week. We've been able to do without so far, but it will come up at some point.

[...]

PR #159, opened 2015-08-27
clean jobs without timestamp

albertjan:
Allthough I don't really understand how it could be undefined. I'm seeing timestamp is `NaN`. But this should take care of both situations.


albertjan:
@manast do you think this solves: #156?

manast:
I don't know if it resolves that specific issue but this is definitely a bug fix worth merging :+1:

PR #160, opened 2015-08-27
Try to be smart about the handler

albertjan:
This tries to resolve the easy to make mistake with specifying a handler with a callback and returning a promise.

It's a bit intricate but it does the job :smile:.

This will make these work:

``` js
queue.process(function(job, done) {
  return Promise.resolve();
});
```

and

```
queue.process(function(job, done) {
  return 5;
});
```

This one will probably create a race condition (for as far as possible in js):

``` js
queue.process(function(job, done) {
   done(5);
   return Promise.resolve(6);
});
```

And this one will outright not work for obvious reasons:

``` js
queue.process(function(job, done) {
  setTimeout(done, 500);
  return 4;
});
```

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -222,7 +222,31 @@ Queue.prototype.setHandler = function(concurrency, handler){
   handler = handler.bind(this);
 
   if(handler.length > 1){
-    this.handler = Promise.promisify(handler);
+    this.handler = function(job) {
+      // tell jshint to ignore the next line because I really really want to
+      // create a new promise here.
+      /* eslint-disable */
+      return new Promise(function(resolve, reject) {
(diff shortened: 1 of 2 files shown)


albertjan:
@manast have you looked at this? It may help new users.

manast:
It would actually be great to have those examples above as unit tests :)

albertjan:
Yeah I cheated a little on the tests. Will do!

albertjan:
@manast I added tests for this. I'm still not 100% sure this is the right direction. :smile: We might want to educate more and throw an exception when a callback is specified and a promise is returned.

manast:
I think we should deprecate the done callback in version 3 anyway so I will close this PR.

PR #165, opened 2015-09-09
Added methods to find out current job state.

kulbirsaini:
Fixes OptimalBits/bull#162

PR #179, opened 2015-09-21
Issue 162

albertjan:
Same as #165 but with tests and without lint. @kulbirsaini


kulbirsaini:
You are the best :+1:

albertjan:
Thanks :smile:

PR #186, opened 2015-09-25
Correct order delay

### LINKED ISSUE #180: Jobs getting stuck under small load
I have been very interested in replacing `kue` with `bull` and I have been doing some testing, however, I keep running into the problem of jobs getting stuck.

### Scripts

So my simply setup has two scripts, one is `scheduler` and the other one is `worker`

`scheduler`:

```coffeescript
#!/usr/bin/env coffee

fs = require 'fs'
bull = require 'bull'
matador = require 'bull-ui/app'
co = require 'co'
Q = require 'q'

credentials = JSON.parse fs.readFileSync './credentials.json', 'utf8'

redisHost = credentials.redisHost
redisPort = credentials.redisPort
redisPass = credentials.redisPass
redisDBNumber = credentials.redisDBNumber

queue = bull 'test', redisPort, redisHost

app = matador
  redis:
    host: redisHost
    port: redisPort
    password: redisPass

sleep = (duration) ->
  deferred = Q.defer()
  setTimeout ->
    deferred.resolve()
    return
  , duration
  return deferred.promise

co ->

  for i in [0...100]
    job = yield queue.add {}, 
      delay: 10 * 1000
    console.log "Job #{job.jobId} scheduled"

  yield sleep 1000

  for i in [0...10]
    for j in [0...10]
      job = yield queue.add {}, 
        delay: 8 * 1000
      console.log "Job #{job.jobId} scheduled"
    yield sleep 1000

  return

.catch (err) ->

  console.log err

  return
```

Using a bit of ES6 generator logic here, so bare with me.

`worker`:

```coffeescript
#!/usr/bin/env coffee

fs = require 'fs'
bull = require 'bull'
co = require 'co'
Q = require 'q'

credentials = JSON.parse fs.readFileSync './credentials.json', 'utf8'

redisHost = credentials.redisHost
redisPort = credentials.redisPort
redisPass = credentials.redisPass
redisDBNumber = credentials.redisDBNumber

queue = bull 'test', redisPort, redisHost

queue.process (job, done) ->
  console.log "Job #{job.jobId} is processed!"
  done()
  return

queue.on 'error', (err) ->
  console.log err
  return
```

### Test runs

I am using iojs v2.5.0, tested with both redis 2.8.19 and 3.0.3. I flush redis before every run.

So when I run the scheduler first, and wait till all the jobs are scheduled, and then run the worker, then all jobs will go to completion.

If I remove the delay settings from the jobs, then everything seems fine - all jobs are processed.

When I run the worker and the scheduler at the same time - in other words, let the worker process as the scheduler goes, only a fraction of the jobs will be processed (ranging between 5 - 50 jobs out of the 200 jobs scheduled). Once the processing stops, no matter how long I wait, the jobs are stuck in the delayed state.

I suspect there is something going on with how the delayed jobs are setup, any insights?

evanhuang8:
Expanding on the concept of using 21 bits of the timestamp to store jobId to achieve FIFO behavior, I fixed the errors with the lua script.

Also updated the timestamp for the delay timer to avoid it being set to something in the past, fixes #180

I tried to add a test for the FIFO behavior when jobs are scheduled at the exact timestamp, however, the test fails. I am not sure if the test is flawed or that the solution for the FIFO behavior doesn't work, so if you can take a look that will be great. Maybe it has to do with the loss of precision when numbers are converted?

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -26,7 +26,7 @@ var semver = require('semver');
     job -> wait -> active
         |    ^        |   \
         v    |        |    -- > failed
-        delayed 
+        delayed
 */
 
 /**
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -42,6 +42,7 @@ var semver = require('semver');
 var MINIMUM_REDIS_VERSION = '2.8.11';
 var LOCK_RENEW_TIME = 5000; // 5 seconds is the renew time.
 var CLIENT_CLOSE_TIMEOUT_MS = 5000;
+var POLLING_INTERVAL = 5000;
 
 var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   if(!(this instanceof Queue)){
@@ -145,12 +146,26 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   //
   // Init delay timestamp.
   //
-  _this.delayedTimestamp = Number.MAX_VALUE;
+  this.delayedTimestamp = Number.MAX_VALUE;
   updateDelaySet(this, Date.now()).then(function(timestamp){
     if(timestamp){
       _this.updateDelayTimer(timestamp);
     }
   });
+
+  //
+  // Create a guardian timer to revive delayTimer if necessary
+  // This is necessary when redis connection is unstable, which can cause the pub/sub to fail
+  //
+  this.guardianTimer = setInterval(function() {
+    if(_this.delayedTimestamp - Date.now() > POLLING_INTERVAL){
+      updateDelaySet(_this, Date.now()).then(function(timestamp){
+        if(timestamp){
+          _this.updateDelayTimer(timestamp);
+        }
+      });
+    }
+  }, POLLING_INTERVAL);
 
 };
 
@@ -174,6 +189,7 @@ Queue.prototype.disconnect = function(){
 Queue.prototype.close = function(){
   var _this = this;
   clearTimeout(this.delayTimer);
+  clearInterval(this.guardianTimer);
   return new Promise(function(resolve) {
     _this.timers.whenIdle(function(){
       resolve(_this.disconnect());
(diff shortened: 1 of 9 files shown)


lib/queue.js
```diff
@@ -428,8 +433,10 @@ var updateDelaySet = function(queue, delayedTimestamp){
     'local jobId = RESULT[1]',
     'local score = RESULT[2]',
     'if (score ~= nil) then',
+    ' redis.log(redis.LOG_WARNING, tostring(score))',
     ' score = score / 0x100000 ',
-    ' if (score <= ARGV[2]) then',
+    ' if (score <= tonumber(ARGV[2])) then',
+    '  redis.log(redis.LOG_WARNING, jobId)',
```
albertjan: this can probably go :smile:

albertjan:
Great work though

evanhuang8:
Thanks. If you have a chance, take a look at the test regarding FIFO order for jobs at the same timestamp - it still doesn't work

evanhuang8:
Alright I updated the code and everything works now, ran test multiple times on my local machine and all passed, not sure why CI is failing.

Anyway, in my understanding the original solution didn't work because of a loss of precision. Javascript only supports up to 53 bits for integer due to its IEEE754 format, which gives a max of 9007199254740992. However, a timestamp now takes up 41 bits (till year 2030, it's still 41 bits), so it only gives us 53 - 41 = 12 bits to work with.

As a result, I reduce the shifts and everything works now. Unfortunately, the limitation is lower now, only allowing room for 2^12 = 4096 jobs for each unique timestamp before wrapping around. In most cases this should not cause funky behavior, but in case someone is actually hitting this limit, they should be aware.

In addition, in `updateDelaySet` it was originally using a RPUSH to push jobs into the `wait` list, but since the `moveJob` that moves jobs from `wait` to `active` uses `brpoplpush`, it will cause to the order to be LIFO. Not sure if it was done on purpose, but in this commit I am changing it to `LPUSH` to achieve FIFO.

manast:
great job man!!

evanhuang8:
@manast thanks! And regarding the using LUA instead of `multi`, I remember now, the reasoning I had was that by using multi you can't really check if the key actually exists

evanhuang8:
Alright @manast things are now consolidated, let me know if anything else!

manast:
looks great. Just wondering if we could get rid of waiting the 'ready' event on the tests, since I assume the same problem that happens with the tests could happen to normal code, cant it?

evanhuang8:
The guardian timer solves the problem, but it means that a lot of the tests would have to have a timeout of 5000ms+, since they need the guardian timer to revive the delay timer first. For the simplicity of the tests, I think `ready` is better off there

Issue #192, opened 2015-10-06
Job should be able to be promoted from delayed to wait in order to be executed immediately


evanhuang8:
It will be a very useful function to have, especially in debugging. I can submit a pull request in a bit.

evanhuang8:
@manast please take a look! Thanks.

PR #193, opened 2015-10-06
Job Promotion, Attempts Made Fix

evanhuang8:
Added a method to promote a job from delayed immediately to wait, instead of having to wait for the timer. 

Also fix the issue where a successful attempt at processing the job does not increment the `attemptsMade` counter.

Fix job tests level: `get job status` should not be part of `.moveToFailed`


evanhuang8:
Fixes #192

evanhuang8:
@manast all updated, let me know if anything else!

PR #197, opened 2015-10-12
Clarify Queue.add() in README

c0d3runn3r:
README didn't explicitly state that you got a Job when the Queue.add() promise resolves.  This is pretty important if you are using the queue processors pattern and the enqueuing agent wants to track job status.


manast:
thanks!

PR #200, opened 2015-11-06
Improve documentation

zxcabs:
- add `ready` and `error` event in readme


manast:
thanks for the patch!

Issue #90, opened 2015-01-28
SyntaxError: Unexpected token u


lchenay:
```
SyntaxError: Unexpected token u
at Object.parse (native line 0 col 0)
at Function.Job.fromData (/app/node_modules/bull/lib/job.js line 220 col 40)
var job = new Job(queue, jobId, JSON.parse(data.data), JSON.parse(data.opts));
at <unknown> (/app/node_modules/bull/lib/job.js line 36 col 18)
return Job.fromData(queue, +jobId, jobData);
at tryCatch1 (/app/node_modules/bull/node_modules/bluebird/js/main/util.js line 45 col 21)
try { return fn.call(receiver, arg); }
at Promise$_callHandler [as _callHandler] (/app/node_modules/bull/node_modules/bluebird/js/main/promise.js line 660 col 13)
x = tryCatch1(handler, receiver, value);
at Promise$_settlePromiseFromHandler [as _settlePromiseFromHandler] (/app/node_modules/bull/node_modules/bluebird/js/main/promise.js line 675 col 18)
var x = this._callHandler(handler, receiver, promise, value);
at Promise$_settlePromiseAt [as _settlePromiseAt] (/app/node_modules/bull/node_modules/bluebird/js/main/promise.js line 845 col 14)
this._settlePromiseFromHandler(handler, receiver, value, promise);
at Promise$_settlePromises [as _settlePromises] (/app/node_modules/bull/node_modules/bluebird/js/main/promise.js line 988 col 14)
this._settlePromiseAt(i);
at Async$_consumeFunctionBuffer [as _consumeFunctionBuffer] (/app/node_modules/bull/node_modules/bluebird/js/main/async.js line 77 col 12)
fn.call(receiver, arg);
at Async$consumeFunctionBuffer (/app/node_modules/bull/node_modules/bluebird/js/main/async.js line 40 col 14)
self._consumeFunctionBuffer();
at process._tickDomainCallback (node.js line 463 col 13)
```

In redis, the jobs data is empty. No opts, no data, no progress, no timestamps...

```
redis:11042> HGETALL bull:jobName:606
1) "delay"
2) "0"
```

I don't know how this is even possible.

albertjan:
Looks like `JSON.parse(opts)` tries to parse `undefined`

```
> JSON.parse(undefined)
SyntaxError: Unexpected token u
    at Object.parse (native)
    at repl:1:7
    at REPLServer.self.eval (repl.js:110:21)
[...]

PR #202, opened 2015-11-08
Catch and log error when trying to create Job from ID

howardchung:
Currently if the fromId function fails (due to corrupted job?), processStalledJobs will never finish.


howardchung:
What I'm seeing is the same as described in https://github.com/OptimalBits/bull/issues/90

The job appears to only be {delay:0}.

Possibly this might fix it:
https://github.com/lchenay/bull/commit/10f421b95bd50262d495a85bc1b347ccc1066d83

howardchung:
Could we add the catch to Job.fromData instead?

https://github.com/OptimalBits/bull/blob/master/lib/job.js#L45

I don't see any dependencies on Redis commands there, so it should not error due to transient Redis failure.

manast:
That would be better :)

PR #206, opened 2015-11-10
add try/catch to fromData, move completed stalled jobs to completed

howardchung:
This change should stop processStalledJobs from getting an error, which causes it to get stuck due to no catch.


howardchung:
Any problems with this pull request?

howardchung:
@manast ?

manast:
the problem is that there are too many changes not related to the title of the pull request... if you could break it down so that we know why there are there it would be much easier to merge.

howardchung:
rebased and squashed into one commit.  There are only two changes:

* try/catch the fromData.  If the JSON parse fails, remove the job.
* Move stalled completed jobs to completed.

howardchung:
synced to master.  Shouldn't this fix #273 ?

lib/job.js
```diff
@@ -393,8 +393,19 @@ Job.prototype._saveAttempt = function(){
...
 Job.fromData = function(queue, jobId, data){
-  var job = new Job(queue, JSON.parse(data.data), JSON.parse(data.opts));
-  job.jobId = jobId;
+  try {
+    var job = new Job(queue, JSON.parse(data.data), JSON.parse(data.opts));
+    job.jobId = jobId;
+  }
+  catch (err) {
+    //Job unable to be parsed. Remove it.
+    var job = new Job(queue, JSON.parse(data.data), JSON.parse(data.opts));
```
chuym: The above throws because `JSON.parse` attempts to parse undefined. Wouldn't this throw the error again?

howardchung: Yeah. It should be {},{}

lib/queue.js
```diff
@@ -507,6 +507,9 @@ Queue.prototype.processStalledJob = function(job){
             _this.emit('stalled', job);
             return _this.processJob(job, true);
           }
+          else {
+            return job.moveToCompleted();
```
chuym: If this PR is fixing #273 then this line should be on a separate PR.

manast:
I am in the doubt if we should move corrupted jobs to the failed set or sanitize them, I am leaning more towards the failed jobs solution...

Issue #152, opened 2015-08-14
Errors in event listeners are silently ignored 


mosilla:
By default, [EventEmitter](https://nodejs.org/api/events.html#events_class_events_eventemitter) sends errors that occur in event listeners to an `error` handler if one is registered. Otherwise, it throws. Queue is a subclass of EventEmitter, but it doesn't appear to do either of these things:
### With Error Listener

``` javascript
var Queue = require('bull');
var queue = Queue('test');

queue.add({ foo: 'bar' }).then(function () {
    queue.process(function (job, jobDone) {
        console.log('job:', job.data);
        jobDone();
    });
});

queue.on('error', function (err) {
    console.error('error:', err);
});

queue.on('completed', function () {
    queue.close();
    queue.noSuchMethod();
});
```
### Output

```
job: { foo: 'bar' }
```
### Without Error Listener

``` javascript
var Queue = require('bull');
var queue = Queue('test');

queue.add({ foo: 'bar' }).then(function () {
    queue.process(function (job, jobDone) {
        console.log('job:', job.data);
        jobDone();
    });
});

queue.on('completed', function () {
    queue.close();
    queue.noSuchMethod();
});
```
### Output

```
job: { foo: 'bar' }
```

manast:
Good catch, I actually didn't know that event listeners had this behaviour.

evanhuang8:
I would argue that the current behavior is correct, since the error is encountered within the job processing function itself, which means the job fails to complete, but it doesn't necessarily mean the whole queue has encountered an error. 

One way to improve the silent fail behavior is probably have the `handleFailed` function in `processJob` log out some message and the error stack; the other would let the queue emit an `error` event in `handleFailed`, in addition to the current `failed` event. IMO the former is better.

PR #215, opened 2015-12-07
Bubble exceptions in complete and failed listeners

trueter:
After spending two hours to find the typo in my complete handler I think this needs to be added. 
This will help test runners like mocha to fail a test instead of silently timing it out.

Resolves https://github.com/OptimalBits/bull/issues/152


manast:
Actually this should not be needed, just catch the errors from the process function, like this:
```javascript
queue.process(function (job, jobDone) {
+        jobDone();
+      }).catch(function(err){
  // my global error handler
});
```

PR #220, opened 2015-12-22
fixed #218, catch exceptions thrown by queue.processJobs()

### LINKED ISSUE #218: Errors/Exceptions ignored
In following example, I'm returning a data which contains circular reference, so it can't serialized. Bull is not handling this error and running the job in infinite loop. Ideally the following job should fail, by throwing error that , return value can't be serialized.

```
var Queue = require("bull");
var q = Queue("bullbug", 6379, '127.0.0.1');

q.on("error", function(err){ console.log("error", err) });
q.on("failed", function(job,err){ console.log("failed", job.jobId, err)});

q.process(function(job){ 
  console.log("hello", job.data);
  var circular = {};
  circular.x = circular;
  return Promise.resolve(circular);
});

q.add("john");
```

palamccc:


bradvogel:
this can probably be closed

Issue #222, opened 2015-12-29
"low" priority not accepted, keeps defaulting to "normal"


yonivy:
When adding a job with {priority: 'low'} it's added to the 'normal' priority.

PR #223, opened 2015-12-29
fixed 'low' priority changes to 'normal'

yonivy:
jobs can't really be added to the 'low' priority queue because of the way the condition in `getQueue` is formed.
The value is __always `true` under the condition__ `priority == 'low'` (because { 'low': 0, ... })

I opened an issue (#222) on this.


manast:
thanks for the catch and the fix

PR #225, opened 2016-01-06
Fixes Queue.prototype.close with long running jobs

ryan-sandy:
This PR fixes an issue with the lockRenewer in the `Queue.prototype.processJob` function. If a job runs longer than the LOCK_RENEW_TIME, the queue now clears the timer before resetting a new timer.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -548,10 +548,7 @@ Queue.prototype.processJob = function(job){
 
   var lockRenewer = function(){
     job.renewLock(_this.token);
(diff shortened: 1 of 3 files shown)


ryan-sandy:
Fixed. The timer-manager will automatically remove timers if they fire.

nathanmarks:
@manast Would be good to have this released to npm -- I just spent an hour or so figuring this out. Didn't realise it was linked to this issue until the last second.

PR #229, opened 2016-01-11
Fixed #228

### LINKED ISSUE #228: PriorityQueue priorities messed up?
It seems that the jobs do not adhere to the priorities.

I didn't see any tests related to actually checking we get a correct order of execution, so I ran some myself.

This is just the "let's log it out" version, no assertions.

    queue = new PriorityQueue(queueName);
   
    var lowPJobs = [], mediumPJobs = [], highPJobs = [];

    for(var i = 0; i < 4; i++) {
        lowPJobs.push(queue.add({p: 2}, {priority: 'normal'}));
        mediumPJobs.push(queue.add({p: 1}, {priority: 'medium'}));
        highPJobs.push(queue.add({p: 0}, {priority: 'high'}));
    }

    // wait for all the jobs to enter the queue
    Promise
        .all(lowPJobs, mediumPJobs, highPJobs)
        .then(function() {
            queue.process(function(job, done) {
                console.log(job.data);
                done();
            });
        });
    

and the output is always

    { p: 0 }
    { p: 0 }
    { p: 1 }
    { p: 1 }
    { p: 1 }
    { p: 1 }
    { p: 2 }
    { p: 2 }
    { p: 2 }
    { p: 2 }
    { p: 0 }
    { p: 0 }

this 'pattern' is repeating every time the 'highest' priority is present more than 2 times.

Does anybody know anything about why and maybe a clue to what should be fixed? I've skimmed through the code and nothing really caught my eye.

yonivy:
The priorities did not worked correctly.

1. The strategy function received the wrong index so the number of jobs to process did the opposite of what it needs to do (0 for 'critical', 1 for 'high' etc.).
2. Reversing the queues array messed a lot. [Array.prototype.reverse()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) is done in place so we're reversing it on each call, not to mention the other possible side effects that may happen from mutating the original array (didn't test other stuff, but it seems like a bug waiting to happen).

**Note**
The second commit is just to fix some indentation mishaps. I'm working with emacs but I switched to sublime text and installed EditorConfig to prevent any harm.

### DIFF
--- a/lib/priority-queue.js
+++ b/lib/priority-queue.js
@@ -13,57 +13,57 @@ var util = require('util');
    one for every possible priority.
 */
 var PriorityQueue = module.exports = function(name, redisPort, redisHost, redisOptions) {
-    if (!(this instanceof PriorityQueue)) {
-	return new PriorityQueue(name, redisPort, redisHost, redisOptions);
-    }
+  if (!(this instanceof PriorityQueue)) {
+    return new PriorityQueue(name, redisPort, redisHost, redisOptions);
+  }
 
-    var _this = this;
-    this.paused = false;
-    this.queues = [];
+  var _this = this;
+  this.paused = false;
+  this.queues = [];
 
-    for (var key in PriorityQueue.priorities) {
-	var queue = Queue(PriorityQueue.getQueueName(name, key), redisPort, redisHost, redisOptions);
-	this.queues[PriorityQueue.priorities[key]] = queue;
-    }
+  for (var key in PriorityQueue.priorities) {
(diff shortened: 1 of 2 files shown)


manast:
thank you for a great patch

yonivy:
do we have any .eslintrc file or something similar to help with the formatting?

manast:
no, the rules are in the Gulpfile.js

yonivy:
`npm test` is passing on my end but it's failing on `Queue .close` mocha test which I didn't touch... any ideas?

manast:
there is some hazard that sometimes breaks the tests, nobody has been able to figure it out yet.

Issue #234, opened 2016-02-04
Urgent: update delayed set called with a timestamp in the future


evanhuang8:
I recently ran into a situation where the `updateDelaySet` is called with a `delayTimestamp` in the future, which causes future jobs to run. I will try to write a simplified script to replicate the bug (can't really post the application code that I am working on).

I am still unclear about the root cause but one solution that can satisfy the promise that no delayed jobs are prematurely run is to simply set the `delayTimestamp` to `Date.now()` if the former is greater than latter. 

I will submit a pull request shortly, please merge. Thanks!

manast:
please let us know the status of this.

evanhuang8:
Actually I have submitted the PR, please check. Couldn't really write the script since I am out of town at the moment, but essentially I have a task that is scheduled to run immediately, which within it schedules a replicate of itself 30 days later. The second task, supposed to run 30 days later, somehow run immediately

PR #235, opened 2016-02-04
Make sure `updateDelaySet` is not called with a future timestamp

evanhuang8:
Somehow `updateDelaySet` is called with a future timestamp sometimes, and it's very bad. This will guarantees no future jobs are prematurely run before we can hunt down the cause of why the future call is happening.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -58,7 +58,7 @@ Job.prototype.toData = function(){
     timestamp: this.timestamp,
     attempts: this.attempts,
     attemptsMade: this.attemptsMade,
-    stacktrace: JSON.stringify([])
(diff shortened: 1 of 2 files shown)


evanhuang8:
Fixes #234

manast:
I am in the process of cleaning up the unit tests since they have been working unreliably for a while now. It is a larger job than I thought so it will take some time before I will merge this fix (hopefully this week).

manast:
I actually reverted this PR. I did not give it enough thought. A delayed timestamp should always be in the future, since otherwise the job is not delayed to begin with. I am confused about the reason of this PR.

PR #232, opened 2016-02-06
Repeatable jobs

jozsi:
Hello,

I am having a hard time figuring out how to correctly implement a repeatable job. I've tried multiple scenarios:
- once the job has completed, add a new delayed one to the queue
- using retry before/after completion

What would be the correct flow to achieve repeatabiliy?

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -215,7 +215,7 @@ Job.prototype.retry = function() {
 
   return _this.queue.client.evalAsync(script, keys.length, keys[0], keys[1], keys[2], keys[3], _this.jobId)
     .then(function(result) {
-      if (result === -1) {
+      if(result === -1) {
         throw new Error('Job ' + _this.jobId + ' has not yet finished.');
       }
       _this.queue.emit('waiting', _this);


lib/job.js
```diff
@@ -198,22 +198,29 @@ Job.prototype.promote = function(){
...
-  // if queue is LIFO use rpushAsync
-  multi[(this.opts.lifo ? 'r' : 'l') + 'push'](key, this.jobId);
-  multi.publish(channel, this.jobId);
-
-  return multi.execAsync().then(function(){
-    _this.queue.emit('waiting', _this);
-    return _this;
+  var push = (this.opts.lifo ? 'R' : 'L') + 'PUSH';
+  var script = [
+    'if (redis.call("SREM", KEYS[2], ARGV[1]) == 1) or (redis.call("SREM", KEYS[3], ARGV[1]) == 1) then',
```
jozsi: I wonder if we should first check if the job `EXISTS`, like some of the other methods do.

jozsi:
@manast - here's the PR for the LUA rewritten `retry` method that works for `completed` jobs as well.

lib/job.js
```diff
@@ -198,22 +198,29 @@ Job.prototype.promote = function(){
...
+    ' redis.call("' + push + '", KEYS[1], ARGV[1])',
+    ' redis.call("PUBLISH", KEYS[4], ARGV[1])',
+    ' return 0',
+    'end',
+    'return -1'
+  ].join('\n');
+  var keys = ['wait', 'failed', 'completed', 'jobs'].map(function(name) {
```
jozsi: Shouldn't the job be added moved the `delay` queue? By skipping it directly to `wait` it does not honor the delay option (if specified).

ivasilov:
Hey guys, this looks like exactly what I need. I'm running into some really weird bugs where the job is processed twice.
@jozsi do you have any updates on your fork?

jozsi:
@ivasilov - I ended up using this (it uses internal methods, use it at your own risk):
```javascript
const co = require('co');

Queue.prototype.repeat = co.wrap(function*(data, opts) {
  this.on('completed', (job, result) => {
    this.client.sremAsync(this.toKey('completed'), job.jobId)
      .then(() => job._addToDelayed(Date.now() + opts.delay));
  });

  let job;
  const count = yield this.getJobCountByTypes('paused', 'delayed', 'wait', 'active');
  if (count) {
    const id = yield this.client.getAsync(this.toKey('id'));
    job = yield this.getJob(id);
  }

  return job || this.add(data, opts);
});

require('./jobs/bt')(Queue, 'bt')
  .repeat({}, {
    delay: 5 * 60 * 1000
  })
  .catch(console.error);
```

jozsi:
A slightly updated version in order to work with `bull 1.0`:
```js
Queue.prototype.repeat = co.wrap(function*(data, opts) {
  this.on('completed', (job, result) => {
    if (result) {
      this.add(Object.assign({}, data, {
        init: result
      }), opts);
    } else {
      job.getState()
        .then(state => this.client.sremAsync(this.toKey(state), job.jobId))
        .then(() => job.moveToDelayed(Date.now() + opts.backoff));
    }
  });

  let job;
  const count = yield this.getJobCountByTypes('paused', 'delayed', 'wait', 'active');
  if (count) {
    const id = yield this.client.getAsync(this.toKey('id'));
    job = yield this.getJob(id);
  }

  return job || this.add(data, opts);
});

require('./jobs/bt')(Queue, 'bt')
  .repeat({}, {
    attempts: 10,
    backoff: 5 * 60 * 1000
  })
  .catch(console.error);
```

manast:
I  could merge the implementation of repeat and the atomic retry. I can do the PR myself also if you wish.

jozsi:
@manast, my `repeat` implementation is:
1. use-case specific: my providers are observing changes, if there is a change it leaves it as `completed` and creates a new job, otherwise it retries the same job by pushing it back to `delayed` state; this way I can track history of the changes
2. not yet tested (the v1.0 one)

The steps are:
- if there is a job added to the queue in `paused`, `delayed`, `wait` or `active` state, return it
- else, create a new job
- if the job has completed with no result (no observed changes), retry by removing from `completed` and adding to `delayed`
- otherwise, repeat by adding a new job (and setting an `init` value to observe changes)

We need a more general implementation, like:
- job -> repeat(delayed -> wait -> active -> completed)

The following questions arise:
- how does the `failed` state influence it?
- should it care about `attempts(Made)` option?
- how to mark job as `unique`?
- ...?

lib/job.js
```diff
@@ -198,22 +198,29 @@ Job.prototype.promote = function(){
...
-  multi.srem(failed, this.jobId);
-  // if queue is LIFO use rpushAsync
-  multi[(this.opts.lifo ? 'r' : 'l') + 'push'](key, this.jobId);
-  multi.publish(channel, this.jobId);
-
-  return multi.execAsync().then(function(){
-    _this.queue.emit('waiting', _this);
-    return _this;
+  var push = (this.opts.lifo ? 'R' : 'L') + 'PUSH';
+  var script = [
```
chuym: This operation should check if the job is locked.

lib/job.js
```diff
@@ -198,22 +198,29 @@ Job.prototype.promote = function(){
...
+    'return -1'
+  ].join('\n');
+  var keys = ['wait', 'failed', 'completed', 'jobs'].map(function(name) {
+    return _this.queue.toKey(name);
   });
+
+  return _this.queue.client.evalAsync(script, keys.length, keys[0], keys[1], keys[2], keys[3], _this.jobId)
+    .then(function(result) {
+      if(result === -1) {
+        throw new Error('Job ' + _this.jobId + ' has not yet finished.');
```
chuym: This may be misleading. The retry may fail if the job doesn't exist

lib/job.js
```diff
@@ -198,22 +198,29 @@ Job.prototype.promote = function(){
...
-
-  return multi.execAsync().then(function(){
-    _this.queue.emit('waiting', _this);
-    return _this;
+  var push = (this.opts.lifo ? 'R' : 'L') + 'PUSH';
+  var script = [
+    'if (redis.call("SREM", KEYS[2], ARGV[1]) == 1) or (redis.call("SREM", KEYS[3], ARGV[1]) == 1) then',
+    ' redis.call("' + push + '", KEYS[1], ARGV[1])',
+    ' redis.call("PUBLISH", KEYS[4], ARGV[1])',
+    ' return 0',
```
chuym: This is not conventional. Typically, scripts return `1` for successful operations and `0` for failures.

lib/job.js
chuym: I have mixed feelings about it. On the one hand, checking if it the job `EXISTS` is a nice way to gracefully handle leaky job ids. On the other hand, doing so can make errors more subtle and harder to track down, leaking job ids is an issue by itself. Hard fails are often easier to diagnose.

chuym:
The enhancements on `Queue#retry` here are very similar to what I had in mind here https://github.com/OptimalBits/bull/pull/275#issuecomment-229481210 - just needs to check for locks. I don't think it's necessary to actually acquire a lock here since this is atomic.

Issue #243, opened 2016-02-09
Memory leak in `Queue##processJobs`


jbourassa:
Hey,

I have been investigating a memory leak in our workers and I think it's coming from Bull. I came down to a very minimal scenario that leaks, even without adding any job to the queue:

```javascript 
var queue = Queue('some-queue');
queue.process(200, function(job, done) {
  done();
});
```

If you leave this code running long enough, you'll end up with leaking promises. The fact that leaked strings are almost exclusively `bull:scrape-job:wait` and `bull:scrape-job:active` makes me believe that the leak is coming from `Queue##processJobs` — here's a heap dump comparison:

![screen shot 2016-02-09 at 12 10 34 pm](https://cloud.githubusercontent.com/assets/224311/12924073/5d50b286-cf26-11e5-90c7-8bf3d8b1a689.png)

I've dug through the Queue code and I don't really see why this is leaking. Am I missing something?

sam1360:
Are you ever dumping the Queues out of redis using clean? That could be your issue since the jobs will stay in redis till you call clean.

jbourassa:
@sam1360 I am, but this is unrelated; the screenshot is a heap dump of node-inspector showing that memory usage keeps increasing over time, and that there's a bunch of `bull:*` strings that are leaking.

manast:
@jbourassa not sure the strings are leaks. I think V8 caches the strings even after garbage collecting... I tried to find info about it but couldnt find it, but I have similar stuff when debugging memory leak elsewhere.

jbourassa:
@manast Interesting, maybe I'm following a wrong lead. I'm assuming you are not seeing leaks in your bull usage.

I ran the code within the context of my app. I'll try again with a blank app/context, let it run for a while and take a few snapshots. I guess if the memory keeps going up it'll be a sign that the leak is truly in bull. I'll report back.

manast:
@jbourassa do you the three snapshot method for finding leaks? https://www.inqbation.com/website-performance-enhancements-fix-memory-leaks/

I use it when I need to find leaks, I have not investigate them on bull for some time, so it is fully possible some leak has been introduced lately. We should have some regression test for leaks as well... but it is quite time consuming.

jbourassa:
@manast Yes, I'm using multiple snapshots. I dug a little more this morning and could indeed reproduce the promise leak with only bull running on idle (without any jobs being posted).

[...]

Issue #144, opened 2015-07-17
Memory Leak / Error: undefined at Object.ensureErrorObject


Salakar:
Notice memory usage increasing over time quote a lot and ran node debugger to find:

Debugger screens:
https://www.dropbox.com/s/tilf4tc6janjjcq/Screenshot%202015-07-17%2010.56.31.png?dl=0
https://www.dropbox.com/s/smkh1mqg0jjm026/Screenshot%202015-07-17%2011.00.40.png?dl=0

Seems to be something similar to https://github.com/hoodiehq/hoodie.js/issues/418 that issue on another repo. it's Bluebird/promise related.

Overnight with 2 queue items running every 5 secs memory was 2.7gb, and hundreds of thousands of those error strings and Arrays of Function stack traces too.

```
"Error: undefined
    at Object.ensureErrorObject (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/util.js:261:20)
    at Promise.module.exports.Promise._rejectCallback (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/promise.js:451:22)
    at Function.module.exports.Promise.reject.Promise.rejected (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/promise.js:195:9)
    at /Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/lib/queue.js:575:24
    at tryCatcher (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/util.js:26:23)
    at Promise.module.exports.Promise._settlePromiseFromHandler (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/promise.js:489:31)
    at Promise.module.exports.Promise._settlePromiseAt (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/nod"
Error
    at Promise.module.exports.Promise._captureStackTrace (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/debuggability.js:88:23)
    at Promise.module.exports.Promise._then (/Users/Mike/Documents/Homestead/projects/hive/node_modules/bull/node_modules/bluebird/js/main/promise.js:218:13)
[...]

PR #245, opened 2016-02-16
Try to fix Promise related memory leaks

leolannenmaki:
I ran into similar issues as described here https://github.com/OptimalBits/bull/issues/144 and here https://github.com/OptimalBits/bull/issues/243. The issue descriptions and the comment here https://github.com/OptimalBits/bull/issues/243#issuecomment-182936122 helped me to maybe fix the at least some of the leaks. I used node-inspector and snapshotting to debug this.

To test this I had a node program running bull, but doing absolutely nothing. Without these changes the memory consumption would steadily just grow. Now it seems more or less stable.

I also tried to avoid rebinding functions and creating closures in `processJobs` etc.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -41,9 +41,9 @@ Job.create = function(queue, jobId, data, opts){
 };
 
 Job.fromId = function(queue, jobId){
-  // jobId can be null if moveJob returns undefined
+  // jobId can be undefined if moveJob returns undefined
   if(!jobId) {
-    return Promise.resolve(null);
+    return Promise.resolve();
   }
   return queue.client.hgetallAsync(queue.toKey(jobId)).then(function(jobData){
     if(jobData){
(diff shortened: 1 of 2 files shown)


leolannenmaki:
Tests pass locally on Node 4, Travis says no on Node 0.10 https://travis-ci.org/OptimalBits/bull/builds/109556222

jbourassa:
Well done. I was just about to create a PR for this.

For what it's worth, the `finally` recursion leak is "by design", see petkaantonov/bluebird#1009

wearhere:
@manast d58a2bb and ae32056 fix the memory leak @bradvogel and I mentioned, a hearty +1 to this PR! Thanks @leolannenmaki and @jbourassa!

manast:
Than you for this great PR. I will make a release as soon as possible.

PR #248, opened 2016-02-24
Enhance Queue#pause and Queue#resume methods to pause local worker

bradvogel:
Enhances Queue#pause and Queue#resume methods to pause local workers.

This is useful for graceful shutdown. For example:
```
process.on('SIGUSR2', function() {
  // Stop processing new jobs.
  queue.pause(true /* Local */);

  // Wait for all currently processing jobs to complete (not covered in this snippet).
  onAllJobsComplete(function(){
    process.exit(1);
  });
});
```

Let me know if this works! Happy to make changes. I think it's incredibly useful as a step towards supporting graceful shutdown (a la https://github.com/Automattic/kue#graceful-shutdown).

Note that the only part of the API that I'm not completely happy with is that `pause(true)` only actually pauses after the current promise-loop has expired. So it's possible for another job to get processed after `pause(true)` has been called. This is usually fine for the graceful shutdown use case above. But perhaps you have a better idea.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -545,21 +545,6 @@ Queue.prototype.processStalledJob = function(job){
   }
 };
 
-Queue.prototype.pauseLocal = function() {
-  var _this = this;
-  if (!this.paused){
-    this.paused = new Promise(function(resolve) {
-      _this.resumeLocal = function() {
-        resolve();
-        _this.paused = null;
-        return Promise.resolve(); // For API consistency.
-      };
-    });
-  }
-
-  return Promise.resolve(); // For API consistency.
-};
-
 Queue.prototype.processJobs = function(){
   var _this = this;
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -167,6 +167,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     }
   }, POLLING_INTERVAL);
 
+  this.checkWorkerPaused = Promise.resolve();
 };
 
 util.inherits(Queue, events.EventEmitter);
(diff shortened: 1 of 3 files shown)


wearhere:
+1

bradvogel:
@manast let me know if this looks OK and I'll write the unit tests!

manast:
sorry for the late reply, soon I will have more time for development.
Regarding your patch, I think it is better to keep the previous pause/resume methods and just add a boolean argument (true local, false global), to avoid polluting the queue with too many methods.

manast:
Looks great now :).

bradvogel:
Great, let me try to get some unit tests in & squash commits.

lib/queue.js
```diff
@@ -526,10 +545,27 @@ Queue.prototype.processStalledJob = function(job){
   }
 };
 
+Queue.prototype.pauseLocal = function() {
```
wearhere: Should this be public now that using `pause(true)` is recommended?

lib/queue.js
```diff
@@ -358,25 +357,45 @@ Queue.prototype.empty = function(){
...
   will run until finalized).
 
   Adding jobs requires a LUA script to check first if the paused list exist
   and in that case it will add it there instead of the wait list.
 */
-Queue.prototype.pause = function(){
-  return pauseResume(this, true);
+Queue.prototype.pause = function(isLocal /* Optional */){
+  if (isLocal){
+    if (!this.paused){
```
wearhere: Should this be called `pausedLocal` to clarify that it's only this worker that's paused?

Also can you add documentation for this?

bradvogel:
Opened new PR here: https://github.com/OptimalBits/bull/pull/251

PR #249, opened 2016-02-24
Fixes the "Illegal access to a strict mode caller function." bug

### LINKED ISSUE #211: Illegal access to a strict mode caller function
Sometimes when creating a Queue `var testQueue = Queue('test');` I get the above message.  This is on node 0.12.7.

toddbluhm:
- Properly promisify the node redis library based on what the lib
recommends on their site:
https://github.com/NodeRedis/node_redis#promises
- Fixes #211
- Note: this also fixes the current test failure in master


manast:
thanks for the patch!

PR #251, opened 2016-02-26
Adds local variants of pause and resume.

bradvogel:
Enhances Queue#pause and Queue#resume methods to pause local workers.

This is useful for graceful shutdown. For example:
```
process.on('SIGUSR2', function() {
  // Stop processing new jobs.
  queue.pause(true /* Local */);

  // Wait for all currently processing jobs to complete (not covered in this snippet).
  onAllJobsComplete(function(){
    process.exit(1);
  });
});
```

Let me know if this works! Happy to make changes. I think it's incredibly useful as a step towards supporting graceful shutdown (a la https://github.com/Automattic/kue#graceful-shutdown).

Note that the only part of the API that I'm not completely happy with is that `pause(true)` only actually pauses after the current promise-loop has expired. So it's possible for another job to get processed after `pause(true)` has been called. This is usually fine for the graceful shutdown use case above. But perhaps you have a better idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Features:
 - Retries.
 - Priority.
 - Concurrency.
-- Global pause/resume.
+- Pause/resume (globally or locally).
 
 
 Install:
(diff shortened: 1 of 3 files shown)


bradvogel:
Thanks for merging! Mind making an npm release?

manast:
Soon. I am flying to Spain right now...

PR #266, opened 2016-03-08
Queue#pause waits until active jobs are finished

chuym:
This PR makes `Queue#pause` wait until all jobs in `active` state are completed before resolving.

Previously, when pausing a queue, it would immediately resolve the pause promise, making it difficult to perform a clean disconnection.
`Queue#pause` will consume the private method `Queue#whenCurrentJobsFinished`, which checks jobs in `active` state, if there are pending jobs, then it sets up a listener on the `completed` event to recheck for jobs in `active` state. Once active jobs reaches 0, it resolves and in turns resolves the promise returned by `Queue#pause`

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -341,8 +341,9 @@ Queue.prototype.empty = function(){
 */
 Queue.prototype.pause = function(isLocal /* Optional */){
   if(isLocal){
+    var _this = this;
+
     if(!this.paused){
-      var _this = this;
       this.paused = new Promise(function(resolve) {
         _this.resumeLocal = function() {
           resolve();
@@ -350,7 +351,8 @@ Queue.prototype.pause = function(isLocal /* Optional */){
         };
       });
     }
-    return Promise.resolve();
+
+    return this.whenCurrentJobsFinished();
   }else{
     return pauseResumeGlobal(this, true);
   }
@@ -810,6 +812,41 @@ Queue.prototype.clean = function (grace, type) {
       reject(err);
     });
   });
+};
+
+/**
+ * Returns a promise that resolves when active jobs are cleared
+ *
+ * @returns {Promise}
+ */
+Queue.prototype.whenCurrentJobsFinished = function(){
+  var resolve, reject;
+  var promise = new Promise(function(innerResolve, innerReject) {
+    resolve = innerResolve;
+    reject = innerReject;
+  });
+
+  var _this = this;
+
+  var resolver = function(count) {
+    if(count === 0) {
+      resolve();
+    }
+    return count;
+  };
+
+  this.getActiveCount().then(resolver).
+    then(function(count) {
+      if(count === 0) {
+        return null;
+      }
+      _this.on('completed', function() {
+        _this.getActiveCount().then(resolver).catch(reject);
+      });
+      return null;
+    }).catch(reject);
+
+  return promise;
 };
 
 //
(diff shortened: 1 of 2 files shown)


lib/queue.js
```diff
@@ -350,7 +357,9 @@ Queue.prototype.pause = function(isLocal /* Optional */){
         };
       });
     }
-    return Promise.resolve();
+
+    this.whenCurrentJobsFinished().then(pauseResolve).catch(pauseReject);
+    return pausePromise;
```
wearhere: Why not just return `this.whenCurrentJobsFinished()`?

test/test_queue.js
```diff
@@ -835,6 +835,54 @@ describe('Queue', function () {
         return queue.resume(true /* Local */);
       });
     });
+
+    it('should wait until active jobs are finished before resolving pause', function (done) {
```
wearhere: Unit tests, awesome!

lib/queue.js
```diff
@@ -812,6 +821,39 @@ Queue.prototype.clean = function (grace, type) {
...
+  };
+
+  this.getActiveCount().then(resolver).
+    then(function(count) {
+      if(count === 0) return null;
+      _this.on('completed', function() {
+        _this.getActiveCount().then(resolver).catch(reject);
+      });
+      return null;
+    }).catch(reject);
```
wearhere: I think this function would be a little easier to read, and no need for a deferred, if this logic was inside the top promise:

```js
var _this = this;
return new Promise(function(resolve, reject) {
  _this.getActiveCount().then(function(count) {
    if (count === 0) resolve();
    else {
      _this.on('completed', function() {
        _this.getActiveCount().then(function(count) {
          if (count === 0) resolve();
        }).catch(reject);
      });
    }
  }).catch(reject);
});
```

also, if we're only considering "current" jobs / the queue is paused, then I don't believe we need to call `_this.getActiveCount()` multiple times:

```js
var _this = this;
return new Promise(function(resolve, reject) {
  _this.getActiveCount().then(function(count) {
    if (count === 0) resolve();
    else {
      _this.on('completed', _.after(count, resolve));
    }
  }).catch(reject);
});
```

test/test_queue.js
```diff
@@ -835,6 +835,54 @@ describe('Queue', function () {
...
+      var queue = utils.buildQueue();
+      queue.process(function (job, done) {
```
wearhere: Can you not shadow the top-level `done`?

wearhere:
@chuym curious, why are you targeting the `feature/multi-as-lua` branch rather than `master`?

test/test_queue.js
```diff
@@ -835,6 +835,54 @@ describe('Queue', function () {
...
+        setTimeout(done, 200);
+      });
+
+      queue.on('ready', function () {
+        var jobs = [];
+        for(var i = 0; i < 10; i++) {
+          jobs.push(queue.add(i));
+        }
+        Promise.all(jobs).then(function() {
+          queue.pause(true).then(function(count) {
```
wearhere: (part 1) return `queue.pause…` and then… (see part 2 below)

test/test_queue.js
```diff
@@ -835,6 +835,54 @@ describe('Queue', function () {
...
+            var paused = queue.getJobCountByTypes(['wait', 'delayed']).then(function(count) {
+              expect(count).to.be(10);
+              return null;
+            });
+
+            return Promise.all([active, paused]);
+          }).then(function() {
+            done();
+          });
+        });
```
wearhere: (part 2) `}).catch(done)`, no?

manast:
I agree on the comments, but I will merge and fix it myself to gain some time.

wearhere:
Oh, thanks @manast!

Issue #273, opened 2016-03-23
Cleaning a queue with a corrupted job will lead to an error


jlukanta:
I'm getting "SyntaxError: Unexpected token u at Object.parse" error when calling queue.clean() if a job is that to be removed is corrupted.

I think queue.clean() should still proceed even if a job is corrupted. Otherwise, the job will be stuck forever in the queue.

> 2016-03-23T00:21:52.154Z - error: [SyntaxError: Unexpected token u] { stack: 'SyntaxError: Unexpected token 
>     at Object.parse (native)
>     at Function.Job.fromData (/Users/sanitized/node_modules/bull/lib/job.js:546:40)
>     at /Users/sanitized/node_modules/bull/lib/job.js:50:18
> 
> From previous event:
>     at Job.fromId (/Users/sanitized/node_modules/bull/lib/job.js:48:56)
>     at Array.map (native)
>     at /Users/sanitized/node_modules/bull/lib/queue.js:776:29
>     at RedisClient.return_reply (/Users/sanitized/node_modules/redis/index.js:664:25)
>     at HiredisReplyParser.reply_parser.send_reply (/Users/sanitized/node_modules/redis/index.js:332:14)
>     at HiredisReplyParser.execute (/Users/sanitized/node_modules/redis/lib/parsers/hiredis.js:30:18)
>     at Socket.<anonymous> (/Users/sanitized/node_modules/redis/index.js:131:27)
>     at emitOne (events.js:77:13)
>     at Socket.emit (events.js:169:7)
>     at readableAddChunk (_stream_readable.js:146:16)
>     at Socket.Readable.push (_stream_readable.js:110:10)
>     at TCP.onread (net.js:523:20)
> 
> From previous event:
>     at Queue.getJobs (/Users/sanitized/node_modules/bull/lib/queue.js:775:15)
>     at Queue.getCompleted (/Users/sanitized/node_modules/bull/lib/queue.js:748:15)
>     at /Users/sanitized/node_modules/bull/lib/queue.js:822:18
> From previous event:
>     at Queue.clean (/Users/sanitized/node_modules/bull/lib/queue.js:801:10)
>     at Queue.<anonymous> (/Users/sanitized/thumbnailWorker.js:694:11)
>     at emitNone (events.js:67:13)
>     at Queue.emit (events.js:166:7)
>     at /Users/sanitized/node_modules/bull/lib/queue.js:130:11
[...]

PR #274, opened 2016-03-23
Undefined data and undefined opts fix in Job.fromData

jlukanta:
Fix for #273


jlukanta:
Sorry, seems like this is not a complete fix. Will get back to it later.

PR #275, opened 2016-03-24
Undefined data and undefined opts fix in Job.fromData

jlukanta:
Fix for #273

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -391,20 +391,37 @@ Job.prototype._saveAttempt = function(){
 
 /**
 */
-Job.fromData = function(queue, jobId, data){
-  var job = new Job(queue, JSON.parse(data.data), JSON.parse(data.opts));
+Job.fromData = function(queue, jobId, jobData){
+  var data;
+  try{
+    data = JSON.parse(jobData.data);
+  }catch (err) {
+    //swallow exception because data got corrupted somehow.
+    data = {};
+  }
+
+  var opts;
+  try{
+    opts = JSON.parse(jobData.opts);
+  }catch (err) {
+    //swallow exception because opts got corrupted somehow.
+    opts = {};
+  }
+
+  var job = new Job(queue, data, opts);
   job.jobId = jobId;
-  job._progress = parseInt(data.progress);
-  job.delay = parseInt(data.delay);
-  job.timestamp = parseInt(data.timestamp);
-  job.attempts = parseInt(data.attempts);
+  // The following values might be NaN if corrupted
+  job._progress = parseInt(jobData.progress);
+  job.delay = parseInt(jobData.delay);
+  job.timestamp = parseInt(jobData.timestamp);
+  job.attempts = parseInt(jobData.attempts);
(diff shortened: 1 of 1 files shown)


chuym:
Although it's good that `Job#fromId` is sanitizing the input and catching potential failures of `JSON.parse`. I believe this check should be done on `#fromData` as well as that's the method that ultimately constructs the `Job` object and it may be called elsewhere, so it would be nice to sanitize the input there.

chuym:
This also does not remove the invalid key, does it?

manast:
But the question is if we want to sanitize the corrupt/invalid job or actually put it in the failed queue with the attempts variable set to max?

jlukanta:
@chuym `fromData` sanitizes the input not `fromId`.

@chuym This fix should remove the job. `queue.clean` will now get the job, and `Promise.filter` will include the job because the following expression resolves to true: `return job && ((!job.timestamp) || (job.timestamp < Date.now() - grace));`

@manast Sanitizing the corrupt job will result in its removal in the `clean` method. I'm leaning towards sanitizing without putting it in the `failed` set for simplicity. There a few cases to consider if we want to move the job to a `failed` set. For instance, would it make sense to mark a `completed` job as `failed` when calling `clean`? What if we want to clean the `failed` set, and a corrupted job is in the `failed` set? Do we want to emit `failed` event if we discover that a job is corrupted during its removal?

If your concern is that developers wouldn't be able track a corrupted job, perhaps we could have a separate error logging in the library.

bradvogel:
bump on this @chuym

manast:
So what do you think we should do with this?

chuym:
@manast I think this should be merged in for the time being. The reason why this scenario may happen is because a given job key is in a particular set, but job data does not exist, that state might happen when removing a job before: https://github.com/OptimalBits/bull/pull/291/files#diff-6c5bd8371f6ad4b9acf7b619733429e2L231 because that potentially leaves job ids on some sets while removing job data.

I can't think of any particular scenario where this may happen right now as of 1.0rc2. Certainly it's better to not throw a syntax error here, but completely swallowing these errors might hide a more serious underlaying issue.

manast:
I think we could wait a bit before merging this. It does not seem to be a big problem anyway, since there have not been many people reporting this kind of problem.

chuym:
@manast: Just ran into this on 1.0 - I don't have steps to reproduce but this happened after some good jobs failed due a brief network outage. I got a bunch of failed jobs piled up, these had `stacktrace`, `attemptsMade` and `progress` but were missing `data`, `opts`, `timestamp`. Pretty sure these were good to begin with and got corrupted somewhere along the transition to failed.

manast:
we have discussed about improving retry here: https://github.com/OptimalBits/bull/pull/232

chuym:
Thanks!

bradvogel:
I don't think we should should merge this and instead address the root cause: https://github.com/OptimalBits/bull/issues/371

Please close this issue

bradvogel:
@manast please close

PR #284, opened 2016-04-07
[FIX] correctly allows the specification of the db

rook2pawn:
@manast without this change, the only way to specify the DB number is to use the first version of instantiating where the first argument redisPort is an object. This correctly fixes the code to read the db number regardless if it is specified in the first argument as an object property or the last argument as a list of redisOptions.


manast:
thanks for the fix!

Issue #82, opened 2015-01-25
Error when launching Bull many times


lchenay:
This happen when using cluster or when launching Bull in different server.

```
TypeError: Cannot call method 'takeLock' of null
  File "/app/node_modules/bull/lib/queue.js", line 357, in Queue.processStalledJob
      return job.takeLock(_this.token).then(function(lock){
  File "/app/node_modules/bull/node_modules/when/sequence.js", line 34, in <unknown>
                    return when(task.apply(null, args), addResult);
  File "/app/node_modules/bull/node_modules/when/when.js", line 652, in <unknown>
                            return reduceFunc(c, value, i, total);
  File "/app/node_modules/bull/node_modules/when/when.js", line 355, in Object.then
                        ? coerce(onFulfilled(value)) : self;
  File "/app/node_modules/bull/node_modules/when/when.js", line 258, in <unknown>
                        value.then(onFulfilled, onRejected, onProgress)
  File "/app/node_modules/bull/node_modules/when/when.js", line 735, in Object.drainQueue [as _onImmediate]
                task();
  File "timers.js", line 345, in processImmediate [as _immediateCallback]
```

manast:
I can only imagine an scenario when this happens. The queues are restarted, they start processing stalled jobs (jobs in the active queue), and while doing this, some of the queues manage to finalize to process a job and explicitly deletes it. The current code assumes that the job will exists all the time, so it is needed to add a check for undefined job in processStalledJob and in that case do nothing.

lchenay:
Is it possible that 2 instances of Bull started at the same time, try to process same stale jobs ?

You said "queues manage to finalize to process". They are no timeout before considering a active job as a stale jobs ?

manast:
no, two queues cannot process the same job twice, thats why we have the lock mechanism. This is just that one queue "believed" that he had to process a job, but other queue already processed it and deleted it. If the other queue had not deleted the job then the other queue would have not being able to process because of the lock (or at least thats the principle :))

PR #291, opened 2016-05-03
Taking lock on removal.

chuym:
Currently, when removing a job, it is possible that the job is currently being processed, but `Job#remove` will remove it regardless if it's being under process or not, which may result in corrupt state. `Jobs#remove` will throw if the job is under process, so this may break compatibility for users that expect `Job#remove` to always remove the job.

Also, as of this commit: https://github.com/OptimalBits/bull/commit/5d724a1f42bd52cedda5ea8660351b1c210cc3d3 only jobs that are not `completed` or `failed` will be removed from the `active`, `waiting` or `paused`, but since it is possible that the queue will try to process already processed jobs as documented here: https://github.com/OptimalBits/bull/issues/82, I think the queue should either:

1. Move the job out of `active` when #82 happens
2. `Job#remove` should remove from all sets as it was being done before the commit above, to avoid `Queue#processStalledJobs` from picking up jobs in `active` set when there is already no data in redis because `Job#remove` was called before. 

If you take a careful look at the function, it will actually remove the job from all sets when called twice.

Maybe we should do both.

Let me know your thoughts about the above, I'm working on adding tests for `Job#remove` and made sure no tests are broken because of this (I ran the tests manually, some tests fail but they are failing on `master` too, so not a regression)

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -294,9 +294,7 @@ Job.prototype.remove = function(){
         queue.emit('removed', job);
       })
       .finally(function () {
-        return job.releaseLock().then(function() {
-          return job;
-        });
+        return job.releaseLock();
       });
   });
 };
--- a/lib/job.js
+++ b/lib/job.js
@@ -283,8 +283,21 @@ Job.prototype.getState = function() {
 Job.prototype.remove = function(){
   var queue = this.queue;
   var job = this;
-  return scripts.remove(queue, this.jobId).then(function(){
-    queue.emit('removed', job);
+
+  return job.takeLock().then(function(lock) {
+    if (!lock) {
+      throw new Error('Could not get lock for job: ' + job.jobId + '. Cannot remove job.');
+    }
+
+    return scripts.remove(queue, job.jobId)
+      .then(function() {
+        queue.emit('removed', job);
+      })
+      .finally(function () {
+        return job.releaseLock().then(function() {
+          return job;
+        });
+      });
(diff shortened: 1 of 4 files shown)


bradvogel:
ping @manast  :)

lib/job.js
```diff
@@ -283,8 +283,21 @@ Job.prototype.getState = function() {
...
+    }
+
+    return scripts.remove(queue, job.jobId)
+      .then(function() {
+        queue.emit('removed', job);
+      })
+      .finally(function () {
+        return job.releaseLock().then(function() {
+          return job;
+        });
```
manast: I do not think returning job here has any effect, since finally does not "overwrite" the original returned value (undefined in this case).

manast:
So basically the change is deleting already completed or failed jobs... I am not sure why we had that check before, if it was just an arbitrary design choice or if there is a larger issue behind the decision. I cannot figure it out now. I guess that if you add some tests for remove we could merge this.

chuym:
@manast Maybe that was thought as an optimization back then with unforeseen consequences at the time.
I have added tests for this and also an additional change here: https://github.com/OptimalBits/bull/pull/291/commits/f9bf70641e3ec4cf4e76810c516d503a894e9444 - since we take the lock when removing a job, we have to make sure that it is releaesd on completion.

manast:
Looks great, thanks for the PR.

PR #300, opened 2016-05-16
Releasing lock from processStalledJobs

chuym:
Currently, `Queue#processStalledJobs` can run into a state where it locks a `completed` job and doesn't unlock when it does the check to see if the job is `completed` by the time it actually attempts to process it.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -505,7 +505,14 @@ Queue.prototype.processStalledJob = function(job){
           if(!isMember){
             _this.emit('stalled', job);
             return _this.processJob(job, true);
+          } else {
+            return job.releaseLock(_this.token);
           }
+        }).catch(function(err) {
+          // Any uncaught error will come here. We'll ensure that the job lock is released and rethrow
+          // the error so it bubbles normally
+          job.releaseLock(_this.token);
+          throw err;
         });
       }
     });


lib/queue.js
```diff
@@ -505,7 +505,14 @@ Queue.prototype.processStalledJob = function(job){
...
             return _this.processJob(job, true);
+          } else {
+            return job.releaseLock(_this.token);
           }
+        }).catch(function(err) {
+          // Any uncaught error will come here. We'll ensure that the job lock is released and rethrow
+          // the error so it bubbles normally
+          job.releaseLock(_this.token);
+          throw err;
         });
```
manast: Wouldn't be better to put a finally on the promise returned by job.takeLock() in line 501, so that we always release the lock, and just be sure that even if we release an already released lock we do not throw any exception (which I think will be the standard behaviour inside a finally clause anyway:
```javascript
return job.takeLock(_this.token).then(function(lock){
      if(lock){
        var key = _this.toKey('completed');
        return _this.client.sismemberAsync(key, job.jobId).then(function(isMember){
          if(!isMember){
            _this.emit('stalled', job);
            return _this.processJob(job, true);
          }
        });
      }
    }).finally(function(){
      job.releaseLock(_this.token);
    });
```

chuym: I thought about using finally, but then I realized If we release an already released lock, we could  inadvertently release it for a resource that took the lock lawfully. With finally we are in a situation where we release the lock _at least once_, but with catch, we ensure that the lock is released _exactly once_ thus not running the risk of releasing the lock for some other consumer.

manast: Ok, in that case this should do it?
```javascript
 return job.takeLock(_this.token).then(function(lock){
      if(lock){
        var key = _this.toKey('completed');
        return _this.client.sismemberAsync(key, job.jobId).then(function(isMember){
          if(!isMember){
            _this.emit('stalled', job);
            return _this.processJob(job, true);
          }
        }).finally(function(){
           job.releaseLock(_this.token);
        });
      }
  })
```

manast: @chuym ping! :)

manast: After this fix I think we could release rc2.

manast:
@chuym hello, hello?!

lib/queue.js
chuym: Hey @manast, sorry for the delay.

We will be in a similar situation, because this `finally` will remove attempt to remove a lock in the event that `processJob` runs, since `processJob` already releases the lock, we'll be releasing twice.

I think we should still do it in `catch` because `catch` will run on a codepath where `processJob` does not release the lock, thus only releasing the lock once.

chuym:
@manast Apologies for the delay ! Responded above.

PR #310, opened 2016-06-17
Improve Queue#clean performance

chuym:
This PR moves `Queue#clean` to a script for performance purposes.

The test case to measure performance was to create 20,000 jobs, then issue a regular `Queue#clean` call, I measured how long it took in my machine to add the jobs and then the cleaning, these are just one sample of the tests I ran

Current implementation:
adding-jobs: 2657ms
cleaning-jobs: 23515ms

With this PR
adding-jobs: 2658ms
cleaning-jobs: 3708ms

I wanted to keep the script as simple as possible, I'm not 100% well versed on redis lua scripts, so please review for any potential bug on it.

Thanks!

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -791,9 +791,10 @@ Queue.prototype.toKey = function(queueType){
  *
  * @param {int} grace - The grace period
  * @param {string} [type=completed] - The type of job to clean. Possible values
+ * @param {int} The max number of jobs to clean
  * are completed, waiting, active, delayed, failed. Defaults to completed.
  */
-Queue.prototype.clean = function (grace, type) {
+Queue.prototype.clean = function (grace, type, limit) {
   var _this = this;
 
   return new Promise(function (resolve, reject) {
@@ -814,7 +815,7 @@ Queue.prototype.clean = function (grace, type) {
       return reject(new Error('Cannot clean unkown queue type'));
     }
 
-    return scripts.cleanJobsInSet(_this, type, Date.now() - grace).then(function (jobs) {
+    return scripts.cleanJobsInSet(_this, type, Date.now() - grace, limit).then(function (jobs) {
       _this.emit('cleaned', jobs, type);
       resolve(jobs);
       return null;
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -797,8 +797,6 @@ Queue.prototype.clean = function (grace, type) {
   var _this = this;
 
   return new Promise(function (resolve, reject) {
-    var getter;
-
     if(grace === undefined || grace === null) {
       return reject(new Error('You must define a grace period.'));
     }
@@ -809,26 +807,14 @@ Queue.prototype.clean = function (grace, type) {
 
     if(_.indexOf([
       'completed',
-      'waiting',
+      'wait',
       'active',
       'delayed',
       'failed'], type) === -1){
       return reject(new Error('Cannot clean unkown queue type'));
     }
 
-    getter = 'get' + type.charAt(0).toUpperCase() + type.slice(1);
-
-    _this[getter]().then(function (jobs) {
-      // take all jobs outside of the grace period
-      return Promise.filter(jobs, function (job) {
-        return job && ((!job.timestamp) || (job.timestamp < Date.now() - grace));
-      });
-    }).then(function (jobs) {
-      // remove those old jobs
-      return Promise.each(jobs, function (job) {
-        return job.remove(_this.token);
-      });
-    }).then(function (jobs) {
+    return scripts.cleanJobsInSet(_this, type, Date.now() - grace).then(function (jobs) {
       _this.emit('cleaned', jobs, type);
       resolve(jobs);
       return null;
--- a/lib/scripts.js
+++ b/lib/scripts.js
@@ -371,6 +371,68 @@ var scripts = {
 
     return execScript.apply(scripts, args);
   },
+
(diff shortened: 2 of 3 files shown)


lib/scripts.js
```diff
@@ -371,6 +371,68 @@ var scripts = {
...
+        hash = 'cleanOSet';
+        break;
+    }
+
+    var script = [
+      command,
+      'local deleted = {}',
+      'local jobTS',
+      'for _, job in ipairs(JOBS) do',
+      '  if (redis.call("EXISTS", ARGV[1] .. job ..  ":lock") == 0) then',
```
chuym: I'm uncertain if it needs to take the lock here.

manast: hmm, I don't think you need, since you are removing the job atomically after checking if it is locked.

manast:
this is a great PR. Just one important consideration. If deleting the jobs takes 3 seconds, it will mean that redis will be blocked for 3 seconds, since it is performed atomically and no other operations can be performed until it has finished. This can be acceptable or not, depending on the use case.

lib/queue.js
```diff
@@ -804,26 +802,14 @@ Queue.prototype.clean = function (grace, type) {
 
     if(_.indexOf([
       'completed',
-      'waiting',
+      'wait',
       'active',
```
manast: Interesting. This has been a bug since like forever...

lib/scripts.js
```diff
@@ -371,6 +371,68 @@ var scripts = {
...
     return execScript.apply(scripts, args);
   },
+
+  cleanJobsInSet: function(queue, set, ts) {
+    var command;
+    var hash;
+    switch(set) {
+      case 'completed':
+      case 'failed':
+        command = 'local JOBS = redis.call("SMEMBERS", KEYS[1])';
```
manast: why not call the local variable just ```jobs``` to follow the coding standard.

lib/scripts.js
manast: I see you are generating the key in LUA, this is actually not recommended by redis. Also, I don't know how fast is string concatenation in LUA, but it is better to make the concatenation once instead of for every loop iteration.

lib/scripts.js
```diff
@@ -371,6 +371,68 @@ var scripts = {
...
+      '  if (redis.call("EXISTS", ARGV[1] .. job ..  ":lock") == 0) then',
+      '    jobTS = redis.call("HGET", ARGV[1] .. job, "timestamp")',
+      '    if(not jobTS or jobTS < ARGV[2]) then',
+      '      redis.call("LREM", ARGV[1] .. ARGV[3], 0, job)',
+      '      redis.call("LREM", ARGV[1] .. ARGV[4], 0, job)',
+      '      redis.call("ZREM", ARGV[1] .. ARGV[5], job)',
+      '      redis.call("LREM", ARGV[1] .. ARGV[6], 0, job)',
+      '      redis.call("SREM", ARGV[1] .. ARGV[7], job)',
+      '      redis.call("SREM", ARGV[1] .. ARGV[8], job)',
+      '      redis.call("DEL", ARGV[1] .. job)',
```
manast: same comment regarding key generation in LUA as for the lock key. But I am wondering, are all these calls necessary? the job should only exist in one place, is this just as a safeguard for a theoretically impossible situation?

manast: (Im just thinking that the loop could be made 5 or 6 times faster if only removing the job one place)

chuym:
> If deleting the jobs takes 3 seconds, it will mean that redis will be blocked for 3 seconds

Yeah that's a downside of this. Another idea is to batch the deletes in `1000` batches perhaps, to diminish the time this script blocks. Thoughts?

manast:
@chuym I have given you some hints on how to optimize the script, please read my latest comments in the code.
The ultimate solution would be to make this function iterative, as you mentioned, in 1000 item batches, which would be resolved in a millisecond or so.

lib/scripts.js
chuym: I actually attempted to do this, I saw a 10% gain on running time, which is not bad, but I wanted to publish the PR first to get some feedback first.

Originally `Queue#clean` uses `Job#remove` to clean the jobs, which in turns removes from all sets, so I followed that logic here. However, I agree that a job must exist in one place only.

If we have a lock, we can guarantee that a given job is in only one set and will remain so for all different states, so we can make improvements there.

The above reminds me of: https://github.com/OptimalBits/bull/pull/291#issuecomment-217281186 I now understand why there was that check in the first place, `LREM` is a O(n) operation, so it was issued only if needed for performance reasons, however the only way we could guarantee that this is the case is if we have a job lock, otherwise it would leave jobs in a corrupted state in some scenarios (I noticed that there were jobs in `completed` but that had no job data). Since we are taking locks on removal, we could probably bring back that check to improve `Job#remove`.

lib/scripts.js
chuym: Ah yes these can be done only once. For getting the key for a job though, that has to be done on every iteration.

chuym:
@manast Thanks for the comments I'll address them

chuym:
@manast I've ammended the PR, please take a second look when you have the time. Thanks for following up!

After pondering a bit, I decided not to remove in batches since it can be problematic: If we delete in batches, how do we know when we are done? I added support to optionally limit the number of jobs to clean by passing an optional paramter to `Queue#clean`. This way we give clients the tools to handle that themselves.

Issue #309, opened 2016-06-17
API documentation question/suggestion


nathanmarks:
Originally I thought the undocumented methods were to be considered private -- however I see them referenced a lot when searching through other issues as if they are public.

Is the documentation just incomplete and these methods are safe to use? (for eg, `queue.getJobs()` which I am using directly to get `paused` jobs, or any of the other `getXXX` methods for that matter).

There's also undocumented events such as `waiting` which I see is emitted in `Job.create()`.

If these methods are intended to be documented but aren't due to lack of time etc -- have you considered automatically generating markdown API documentation via jsdoc comments?

Eg:
https://github.com/nathanmarks/cbioportal-api-client#api-reference 
https://github.com/nathanmarks/stylishly/blob/master/docs/api/styleManager.md

You can pretty much do whatever you want with the output format via custom templates or even separate files per module. I'm using this library here: https://github.com/jsdoc2md/jsdoc-to-markdown

chuym:
Yeah I'm familiar with jsdoc and I find it pretty a nice way to keep documentation since it lives together with the code. All in for it.

PR #311, opened 2016-06-21
Improvements for Queue#processStalledJobs

chuym:
This PR makes Queue#processStalledJobs more flexible by implementing the following 2 items:
1. `Queue` now takes a `queueOptions` hash that, currently, has only one recognized option: `processStalledJobs`. When set to `false`, the queue won't automatically process stalled jobs.
2. `Queue#processStalledJobs` itself accepts an optional parameter: `limit` which will only process that many jobs at a time.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -431,7 +431,9 @@ Queue.prototype.run = function(concurrency){
   var _this = this;
 
   // In case of connection loss, running `processStalledJobs` will repick jobs here.
-  return this.processStalledJobs().then(function(){
+  var start = this.opts.processStalledJobs ? this.processStalledJobs() : Promise.resolve();
+
+  return start.then(function(){
 
     while(concurrency--){
       promises.push(new Promise(_this.processJobs));
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -489,11 +489,11 @@ Queue.prototype.updateDelayTimer = function(newDelayedTimestamp){
  * Process jobs that have been added to the active list but are not being
  * processed properly.
  *
- * @param {Number?} limit Only process this many number of jobs
+ * @param {Number?} limit Only process this many number of jobs. Greater than 1, otherwise -1
 */
 Queue.prototype.processStalledJobs = function(limit){
   var _this = this;
-  limit = limit >= 0 ? limit : -1;
+  limit = limit > 0 ? limit - 1 : -1;
 
   if(this.closing){
     return this.closing;
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -47,7 +47,7 @@ var LOCK_RENEW_TIME = 5000; // 5 seconds is the renew time.
 var CLIENT_CLOSE_TIMEOUT_MS = 5000;
 var POLLING_INTERVAL = 5000;
 
-var Queue = function Queue(name, redisPort, redisHost, redisOptions){
+var Queue = function Queue(name, redisPort, redisHost, redisOptions, queueOptions){
   if(!(this instanceof Queue)){
     return new Queue(name, redisPort, redisHost, redisOptions);
   }
@@ -57,7 +57,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     var redisOpts = opts.redis || {};
     redisPort = redisOpts.port;
     redisHost = redisOpts.host;
-    redisOptions = redisOpts.opts || {};
+    redisOptions = redisOpts.opts || {};
     redisOptions.db = redisOpts.DB;
   }
 
@@ -76,6 +76,13 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
 
   redisPort = redisPort || 6379;
   redisHost = redisHost || '127.0.0.1';
+
+  queueOptions = _.pick(queueOptions, ['processStalledJobs']);
+  queueOptions = _.defaults(queueOptions, {
+    processStalledJobs: true
+  });
+
+  this.opts = queueOptions;
 
   var _this = this;
 
@@ -423,6 +430,7 @@ Queue.prototype.run = function(concurrency){
   var promises = [];
   var _this = this;
 
+  // In case of connection loss, running `processStalledJobs` will repick jobs here.
   return this.processStalledJobs().then(function(){
 
     while(concurrency--){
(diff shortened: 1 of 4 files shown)


lib/queue.js
```diff
@@ -47,7 +47,7 @@ var LOCK_RENEW_TIME = 5000; // 5 seconds is the renew time.
 var CLIENT_CLOSE_TIMEOUT_MS = 5000;
 var POLLING_INTERVAL = 5000;
 
-var Queue = function Queue(name, redisPort, redisHost, redisOptions){
+var Queue = function Queue(name, redisPort, redisHost, redisOptions, queueOptions){
```
bradvogel: needs readme update?

lib/queue.js
```diff
@@ -423,6 +430,7 @@ Queue.prototype.run = function(concurrency){
   var promises = [];
   var _this = this;
 
+  // In case of connection loss, running `processStalledJobs` will repick jobs here.
   return this.processStalledJobs().then(function(){
```
bradvogel: shouldn't this not be called if `!_this.opts.processStalledJobs` ?

bradvogel: If multiple workers are starting up at various times, all with processStalledJobs turned off, running this means those stuck jobs will still get processed.

lib/queue.js
chuym: Bull's documentation is currently outdated in many regards, I'm addressing this: https://github.com/OptimalBits/bull/issues/309 by documenting everything in jsdoc format, if documentation lives closer to the code, it's much more likely it will be up to date. There are many helpers that generate nice looking documentation, I particularly liked this one: https://camo.githubusercontent.com/724b9224844b6b4f2cd19b3bce8d25015fa54cfa/687474703a2f2f7075752e73682f674f794e652f363663336164636239372e706e67

In any case, yes, this needs a readme update

manast:
thanks for the PR. I am still not convinced that adding this extra complexity is needed. 
Could you elaborate what is the problem we are trying to solve with this?

lib/queue.js
chuym: Good catch:

test cases: https://github.com/OptimalBits/bull/pull/311/commits/1d8e2d2dc9f7e98996d9338ddc93498762e81c9e

fix: https://github.com/OptimalBits/bull/pull/311/commits/6e2fdf56689161a80725840f81eaf559beba4c48

That's the test case you had in mind right?

chuym:
The processStalledJobs functionality is opinionated by assuming that it is completely fine to restart a job in all usage cases, when that's certainly not what you might want on several use cases, specially those where the job handler has side effects that you don't want to accidentally repeat multiple times.

The limit option is useful when you do not want to clog your application by processing stalled jobs - if there are hundreds of thousands of jobs to retry, this means that a single call to `processStalledJobs` will handle all the jobs in a single go

manast:
@chuym ok. but if you disable the process stalled jobs, the active queue will grow overtime and you will not get any notification or explanation why these jobs have not completed.
Regarding the limit parameter, even if processStalledJobs tries to process the all the stalled jobs, it will do so sequentially, so I don't think it will make any noticeable difference...

lib/queue.js
```diff
@@ -423,7 +430,10 @@ Queue.prototype.run = function(concurrency){
   var promises = [];
   var _this = this;
 
-  return this.processStalledJobs().then(function(){
+  // In case of connection loss, running `processStalledJobs` will repick jobs here.
+  var start = this.opts.processStalledJobs ? this.processStalledJobs() : Promise.resolve();
```
bradvogel: is this still needed to be run when the queue starts, instead of just letting it start after `_this.LOCK_RENEW_TIME` ?

chuym:
> the active queue will grow overtime and you will not get any notification or explanation why these jobs have not completed.
Good point, yet, we do not have a periodic `Queue#clean, so completed jobs will also grow overtime without user interaction, although with a straightforward explanation in this case.
Similarly to how users implement `Queue#clean` loops, users can implement their own `Queue#processStalledJobs` as their convenience.

An unbounded `processStalledJobs` is fine for low volume, low parallel queues count, but if you have thousands of jobs per second with many parallel queues, you can imagine how it may grow out of control by pulling thousands of repeated jobs many times where only a subset of these jobs are actually processed.

In fact, I think this can be further improved by both moving this `lrange` to a script, and calling the same function as in `getStalledJobs`, thus avoiding getting jobs that might have been picked up by a parallel `processStalledJobs` call.

lib/queue.js
chuym: The expectations of Bull is to process them immediately on reconnection. Not calling it to begin with result in many failures across tests.

manast:
having an optimized getNextStalledJob should be better than an extra limit parameter that requires finetuning by the user. I can work on that.
regarding the active queue growing overtime is both a more inconsistent behaviour than completed/failed sets, also more difficult to handle. If stalled jobs cannot be restarted, which is the reason for not having auto-process-stalled-jobs, then it will not matter if the user do it manually or not. I think the proper way to solve this is to move the stalled jobs to the failed set...

chuym:
My only reservation is the potential flakiness of `processStalledJobs`, given that Bull considers a job as stalled if it fails to renew the lock on time, and that's not a completely accurate way to tell if the job handler in fact stalled.

I think `processStalledJobs` is currently dangerous, some reasons:

1. If a lock fails to renew, `processJob` won't stop processing and a second worker may work in the same job on parallel. This has major implications when users expects jobs to run exactly once.
2. Moving jobs to completed is not safe, because a job can be moved to completed multiple times - while completed jobs are kept in a set, "returnvalue" is overwritten.
3. If you have many thousands of jobs with high throughput, `processStalledJobs` will clog the queue.

Item 1 above is not addressable, there isn't a safe way to tell if a handler effectively stalled, this is a limitation of Javascript and I don't think `processStalledJobs` will ever be 100% reliable, I would love to be wrong though since it is a nice feature.
Item 2 is an issue that can be fixed as a separate issue
Item 3 is addressed by this PR by giving users control of the behavior of `processStalledJobs`

I understand how useful `processStalledJobs` is, but users should have control of its behavior since it may not fix every usage case.

chuym:
Hey @manast so what is your take on this? Should this be dropped on favor of improving and optimizing `processStalledJobs` ?

manast:
I would like to discuss this feature a bit more so that we are on sync on what is the best approach. There are a few things I would like to decide:
- Have a configurable TTL for jobs. Jobs should fail if they take more time than the defined TTL.
- When a worker that is running a job dies, should the job be restarted automatically by another worker or not? If not, it should be moved to the failed set. If yes, the retry could be implemented using the retry-failed jobs functionality we already have in place.
- New arguments on the Queue constructor bloats the API. I would like instead to have one options argument where options for redis, the queue, etc are well defined and organised.

manast:
btw, moving stalled/stuck jobs back to the wait queue as part of the retry mechanism would be not only an elegant approach but even better performance wise when you have a lot of elements in the active list in a large setup.

bradvogel:
Implemented in https://github.com/OptimalBits/bull/pull/359

PR #318, opened 2016-07-06
Improvements for Job#retry

chuym:
This PR makes some improvements on the current `Job#retry` implementation:

* The retry operation is now implemented through a script, to follow suit with the other operations
* The retry operation does several checks while attempting to retry a job:
  * It checks if the job exists, so that a non-existent job can be accidentally retried
  * It checks if the job is currently locked, to avoid possible corruption on job data
  * It checks if the job is actually failed, so it avoids retrying job that are in any other state that is not failed.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -204,21 +204,26 @@ Job.prototype.promote = function(){
     });
 };
 
+/**
+ * Attempts to retry the job. Only a job that has failed can be retried.
+ *
+ * @return {Promise} If resolved and return code is 1, then the queue emits a waiting event
+ * otherwise the operation was not a success and throw the corresponding error. If the promise
+ * rejects, it indicates that the script failed to execute
+ */
 Job.prototype.retry = function(){
-  var key = this.queue.toKey('wait');
-  var failed = this.queue.toKey('failed');
-  var channel = this.queue.toKey('jobs');
-  var multi = this.queue.multi();
+  var queue = this.queue;
   var _this = this;
-
-  multi.srem(failed, this.jobId);
-  // if queue is LIFO use rpushAsync
-  multi[(this.opts.lifo ? 'r' : 'l') + 'push'](key, this.jobId);
-  multi.publish(channel, this.jobId);
-
-  return multi.execAsync().then(function(){
-    _this.queue.emit('waiting', _this);
-    return _this;
+  return scripts.retryJob(this).then(function(result) {
+    if (result === 1) {
+      queue.emit('waiting', _this);
+    } else if (result === 0) {
+      throw new Error('Couldn\'t retry job: The job doesn\'t exist');
+    } else if (result === -1) {
+      throw new Error('Couldn\'t retry job: The job is locked');
+    } else if (result === -2) {
+      throw new Error('Couldn\'t retry job: The job has been already retried or has not failed');
+    }
   });
 };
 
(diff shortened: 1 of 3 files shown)


chuym:
This is similar to: https://github.com/OptimalBits/bull/pull/232 except that it has more extensive failure handling, more job integrity checks and is in `scripts.js`

test/test_queue.js
```diff
@@ -1297,6 +1297,124 @@ describe('Queue', function () {
...
+    it('should not retry a job that has been removed', function(done) {
+      queue = utils.buildQueue('retry a removed job');
+      queue.on('ready', function() {
+        var attempts = 0;
+        queue.process(function (job, jobDone) {
+          if(attempts === 0) {
+            attempts++;
+            jobDone(new Error('failed'));
+          } else {
+            jobDone();
```
manast: For simplicity here we could just throw and error or not returning anything, and skip the jobDone callback.

lib/job.js
```diff
@@ -204,21 +204,26 @@ Job.prototype.promote = function(){
...
 };
 
+/**
+ * Attempts to retry the job. Only a job that has failed can be retried.
+ *
+ * @return {Promise} If resolved and return code is 1, then the queue emits a waiting event
+ * otherwise the operation was not a success and throw the corresponding error. If the promise
+ * rejects, it indicates that the script failed to execute
+ */
 Job.prototype.retry = function(){
```
manast: Why not allowing retries of completed jobs as well, since this could be reused to implement recurrent jobs in the future.

chuym: One consideration on that though. The current implementation works with failed jobs only, so we'll be breaking some expectations on `Job#retry` by letting it to retry completed job as well.

One common usage scenario I think is having parallel workers that have some kind of retry logic for failed jobs, something like this:

```javascript
setInterval(() => {
  queue.getFailed().then(jobs => {
    return Promise.each(jobs, job => job.retry());
  });
}, 60 * 1000);
```

The above may lead to problems because if another worker retried a job successfully, it might be retried yet again by another worker since that job is also retriable if it's completed.

Perhaps we can allow parameters to `retry` to drive the behavior? or expose `repeat` as a separate Job function instead?

test/test_queue.js
chuym: Ah thanks! will amend

lib/job.js
manast: yes a retry with parameters, or the script part with parameters and then reuse that code in the repeat function.

manast: speaking of which. Wouldn't be nice that retry includes a delay parameter as well?

chuym: Cool, I made the script accept an options parameter to determine on which state the job should be when attempting to reprocess it. This should support retrying a job (from a failed state) or repeating a job (from the completed state).

As for the delay, I took a dive into it, but it was taking me too much time, maybe that can be added on a separate PR?

lib/scripts.js
```diff
@@ -444,6 +444,64 @@ var scripts = {
...
+      '    else',
+      '      return -2',
+      '    end',
+      '  else',
+      '    return -1',
+      '  end',
+      'else',
+      '  return 0',
+      'end'
+    ].join('\n');
```
manast: instead of this nested code I would propose something like this:
```lua
if ((redis.call("EXISTS", KEYS[1]) ~= 1) then
  return 0
end
if ((redis.call("EXISTS", KEYS[2])  ~= 0) then
  return -1
end
.
.
.
return 1;
```

lib/job.js
shmulikmako: why "retry for completed" still not implement?

PR #319, opened 2016-07-08
Pass a connection string instead of separate parameters

### LINKED ISSUE #306: Creating queue from a redis url
Is this an internal API? https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L67

Currently, I do something like this in order to create a queue from a redis url:

```
const { port, hostname, auth } = url.parse(process.env.REDIS_URL);

const queue = bull('id', port, hostname, { auth_pass: String(auth).split(':')[1] });
```

How can I just pass a redis url to Bull? Did I miss something?

afm-sayem:
Fixes #306 

It's very convenient to pass the environment variable for the Redis connection than parsing the URL everytime you want to create a connection. Saves some effort in parsing and passing authentication separately.

I haven't written any test for authentication since I'm unsure how to spawn a Redis server with authentication for running this test. Any help in that regard will be appreciated.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -4,6 +4,8 @@
 var redis = require('redis');
 var events = require('events');
 var util = require('util');
+var assert = require('assert');
+var url = require('url');
 var Job = require('./job');
 var scripts = require('./scripts');
 var TimerManager = require('./timer-manager');
@@ -59,6 +61,19 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisHost = redisOpts.host;
     redisOptions = redisOpts.opts || {};
     redisOptions.db = redisOpts.DB;
+  }
+
+  if (_.isString(redisPort)) {
+    try {
+      var redisUrl = url.parse(redisPort);
+      assert(_.isObject(redisHost) || _.isUndefined(redisHost));
+      redisOptions =  redisHost || {};
+      redisPort = redisUrl.port;
+      redisHost = redisUrl.hostname;
+      redisOptions.password = redisUrl.auth? redisUrl.auth.split(':')[1]: void 0;
+    } catch (e) {
+      throw new Error('Malformed connection string');
+    }
   }
 
   redisOptions = redisOptions || {};
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -61,9 +61,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisHost = redisOpts.host;
     redisOptions = redisOpts.opts || {};
     redisOptions.db = redisOpts.DB;
-  }
-
-  if (!_.isObject(redisPort) && _.isString(redisPort)) {
+  } else if(_.isString(redisPort)) {
     try {
       var redisUrl = url.parse(redisPort);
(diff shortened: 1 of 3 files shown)


lib/queue.js
```diff
@@ -4,6 +4,8 @@
 var redis = require('redis');
 var events = require('events');
 var util = require('util');
+var assert = require('assert');
```
manast: we should not use assert in production code, it is a library aimed for unit testing.

lib/queue.js
```diff
@@ -61,6 +63,19 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisOptions.db = redisOpts.DB;
   }
 
+  if (_.isString(redisPort)) {
```
manast: For more stable code I think it is better to use an else if here (i.e. if redisPort is not an Object, then check if it is a string.

lib/queue.js
```diff
@@ -61,6 +63,19 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
...
+    try {
+      var redisUrl = url.parse(redisPort);
+      assert(_.isObject(redisHost) || _.isUndefined(redisHost));
```
manast: remove this assert.

lib/queue.js
```diff
@@ -61,6 +63,19 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
...
+      redisOptions =  redisHost || {};
+      redisPort = redisUrl.port;
+      redisHost = redisUrl.hostname;
+      redisOptions.password = redisUrl.auth? redisUrl.auth.split(':')[1]: void 0;
+    } catch (e) {
+      throw new Error('Malformed connection string');
```
manast: maybe useful to log the err thrown by the try so that the user can get more information about why the connection string is malformed.

manast:
thanks for the patch. In order to merge this we will also need an update in the README, so that the new feature is made public.

lib/queue.js
afm-sayem: I don't think there's anything wrong with using asserts in try catch block. It'll produce a nice error if the conditions are not fulfilled and propagate them properly. `assert` is just a syntactic sugar of using an if else block where if the conditions fail you throw an error.

You could argue that instead of throwing an error always saying `Malformed connection string` we should send the error message instead.

afm-sayem: `assert` takes a second argument for the error message so I think we should pass a proper error message instead of assertion failure message.

lib/queue.js
```diff
@@ -63,16 +63,19 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisOptions.db = redisOpts.DB;
   }
 
-  if (_.isString(redisPort)) {
+  if (!_.isObject(redisPort) && _.isString(redisPort)) {
```
manast: In my previous comment I mean an else if, the reason is that the first if could actually change redisPort into a string for example, and then you will enter this path which would be wrong.

afm-sayem: Ah I completely misunderstood. Yeah that makes sense since the first block might change redisPort.

PR #335, opened 2016-08-19
Allow custom job ids to be specified in the job options

pricj004:


test/test_job.js
```diff
@@ -56,6 +56,13 @@ describe('Job', function(){
         expect(storedJob.opts.testOpt).to.be('enabled');
       });
     });
+
+    it('should use the custom jobId if one is provided', function() {
```
chuym: Test case suggestion:

Create a job with a customId and test if it can be processed, in the `process` callback, verify that the custom jobId matches the one provided

bradvogel:
Add a bit to the readme too

manast:
I have mixed feelings about this functionality. From one side it seems like something that should be nice to give to certain pro-users, but on the other hand, it may also create a new flora of issues if the user is not able to generate unique ids for every job. I haven't understood yet which consequences it would have to add several jobs using the same jobId... Isn't there a different way to achieve what you want without custom ids?

pricj004:
@manast The general use case is having a concurrency-safe way to add a job if it doesn't already exist, based on some unique value (in our case the user id).

In our case, we're dealing with hundreds of thousands of jobs per hour, and the load on Redis has been pretty intense. We needed a way to avoid adding a job to the queue if one already exists for the user. 

I've deployed our branch running this code to production and the results have been great - we've gone from ~90% CPU utilisation to < 10%.

In terms of other possible ways to achieve the same thing: we could do a transaction script which loops through jobs looking for a match, and inserts it it doesn't find one, but that's slow.

We could also keep a secondary key as a kind of lock to indicate whether a job with that unique value already exists, but that introduces other issues: what happens if some failure occurs and the lock is left orphaned? We'd need some kind of TTL, and it all gets rather complicated.

Being able to override the key seemed the best solution to me. It lets us determine whether a job with that unique value already exists in constant-time, and doesn't have any of the issues around locking.

manast:
I am not sure I understand your use case completely, but what about this: when you call the addJob method you get the unique ID for that given job. If you stored that ID in a redis database using the userId as key, wouldn't you then be able to test in O(1) if the users job exists already? You will of course some code to update the database when jobs are completed/failed, removed and so on.

pricj004:
@manast Unfortunately, that wouldn't be concurrency safe.

We'd need to:

1. Lookup to see if the user's job exists already
2. If not, add it. If so, ignore it.

Another thread could create or remove the job in between 1 and 2. 

We'd have the same problem when removing the lock at the end of the job too, plus the possibility that the job processes but some failure interferes with removing the lock, hence preventing all future jobs from processing for that user, requiring us to build a more complex system of TTLs etc.

manast:
ok. But how do you avoid points 1 and 2 using custom ids in bull?

pricj004:
@manast We wrap 1 and 2 in a transaction. Because we have a known key, we can perform step 1 using an `EXISTS` command (L#84 in `scripts.js`), then conditionally insert depending on the result.

(Note: I've just realised we'll need to execute these within a `multi` call to guarantee atomicity - I'll commit that soon).

manast:
why do you need multi? If the jobId exists already it just returns false or -1 for example, and since it is run in a script it will be atomic.

pricj004:
@manast Oh, are scripts atomic by default? I thought they needed to be wrapped in a `multi` call. If not, then great! No need to make that change then.

lib/scripts.js
```diff
@@ -74,11 +74,11 @@ var scripts = {
...
 
     var argvs = _.map(jobArgs, function(arg, index){
-      return ', ARGV['+(index+2)+']';
+      return ', ARGV['+(index+3)+']';
     })
 
     var script = [
-      'local jobId = redis.call("INCR", KEYS[5])',
+      'local jobId = ARGV[2] == "" and redis.call("INCR", KEYS[5]) or ARGV[2]',
       'redis.call("HMSET", ARGV[1] .. jobId' + argvs.join('') + ')',
```
manast: Why not use if/else here for readability ?

lib/scripts.js
```diff
@@ -65,7 +65,7 @@ var scripts = {
       return result === 1;
     });
   },
-  addJob: function(client, toKey, lifo, job){
+  addJob: function(client, toKey, lifo, customJobId, job){
     var jobArgs = _.flatten(_.toPairs(job));
```
manast: since lifo is optional and customId also, I think it is more convenient to put it in an opts argument, and also move it to the last argument in the list:```js function(client, toKey, job, opts) ```

lib/scripts.js
```diff
@@ -78,7 +78,8 @@ var scripts = {
     })
 
     var script = [
-      'local jobId = ARGV[2] == "" and redis.call("INCR", KEYS[5]) or ARGV[2]',
+      'local numericJobId = redis.call("INCR", KEYS[5])',
+      'local jobId = ARGV[2] == "" and numericJobId or ARGV[2]',
```
manast: I think it is now better to call numericJobId to something like jobCounter which is what it is. If no custom job id is provided, then this jobCounter will be assigned as the jobId giving it the proper semantics.

lib/scripts.js
```diff
@@ -87,7 +88,7 @@ var scripts = {
     var delayTimestamp = job.timestamp + job.delay;
     if(job.delay && delayTimestamp > Date.now()){
       script.push.apply(script, [
-        ' local timestamp = tonumber(ARGV[' + (argvs.length + 3) + ']) * 0x1000 + bit.band(jobId, 0xfff)',
+        ' local timestamp = tonumber(ARGV[' + (argvs.length + 3) + ']) * 0x1000 + bit.band(numericJobId, 0xfff)',
         ' redis.call("ZADD", KEYS[6], timestamp, jobId)',
```
manast: good catch since we use the jobCounter for the delay logic.

lib/scripts.js
```diff
@@ -80,7 +80,9 @@ var scripts = {
     var script = [
       'local numericJobId = redis.call("INCR", KEYS[5])',
       'local jobId = ARGV[2] == "" and numericJobId or ARGV[2]',
-      'redis.call("HMSET", ARGV[1] .. jobId' + argvs.join('') + ')',
+      'local fullJobId = ARGV[1] .. jobId',
+      'if redis.call("EXISTS", fullJobId) == 1 then return jobId end',
```
manast: you could rename fullJobId to jobIdKey since it represents more accurately why we are doing the concatenation with ARGV[1] i

pricj004:
@manast Cool, good suggestions. I have implemented them.

TomKaltz:
@pricj004 @manast I will be using this feature to make sure duplicate jobs don't get added to my queue.  When submitting a new job with the same ID as a previously FAILED job, the new job is not added to the queue.  Is this desired behavior for most.  I would really like it to retry/rerun the failed job.  Can someone explain why the current behavior was chosen?

manast:
@TomKaltz couldn't you just use the retry functionality? or listen to the fail event and delete the job in that case?

TomKaltz:
@manast your suggestion would work I guess.  I just think it would be a good feature for duplicate submitted jobId to automatically retry if it had previously failed.  Currently job is ignored if ID exists and has previously failed.  I see more use-case for my feature request than current functionality of ignoring the job in that state.  Is there any reason why bull can't adopt this functionality?

manast:
@TomKaltz mostly that it requires some extra work and I guess nobody has time to do it right now...

PR #343, opened 2016-09-12
delete the pdfqueue

dengyaolong:
pdfqueue have not declared


manast:
It is just an example in pseudocode.

PR #349, opened 2016-09-14
declare pdfQueue to fix 'ReferenceError: pdfQueue is not defined'

dengyaolong:
When I try to run the quick guide, it throws an error: ReferenceError: pdfQueue is not defined.
Declare it and solve.


manast:
thanks

PR #355, opened 2016-10-10
Update README.md - Queue##clean parameters

PyroSA:
Better match for Queue.prototype.clean


manast:
I dont think we can remove that documentation. The user needs to know which are the valid types as well as the default one if non provided, and an explanation of what the limit argument does.

PyroSA:
Sorry - thought I duplicated 'limit' from the arguments specified below as well :sad: 
Initially only intended to correct the options segment.
I'll add limit in.

Issue #258, opened 2016-03-01
Make brpoplpush and lock an atomic operation


manast:
We have an hazard where while a job is being moved from wait to active, another worker may pick the job as being stalled since the worker that moved the job did not have time to create a lock. This hazard does not seem to be dangerous but it is a malfunction that should be avoided.

(related to #257 )

manast:
this cannot be done properly due to redis limitations (we cannot use blocking commands in multi or lua scripts)

wearhere:
What if the worker locked the job before changing its status?

manast:
if we keep using brpoplpush it is not possible... we would need a polling mechanism, and we want to avoid that.

bradvogel:
Addressed in https://github.com/OptimalBits/bull/pull/358/files

bradvogel:
just fyi: I asked if this is supported in Redis 3.2. Unfortunately it is not: https://twitter.com/BradVogel/status/791189530548195328

PR #358, opened 2016-10-15
Rewrite the handling of process stalled jobs to be atomic so it doesn…

### LINKED ISSUE #356: Jobs get double processed (if removed immediately upon completion)
There exists a situation where a job can get processed a second time by the `processStalledJobs` periodic cleanup job. This only happens when jobs are removed upon completion by the worker that processed them (a common pattern - see #354).

We only discovered this because we saw `Error: Missing Job xx when trying to move from active to completed` show up in our server logs that is running a high-volume (100/events/sec) Bull queue.

This is how it happens:

| Time | Process A | Process B |
| --- | --- | --- |
| 1 | In the regular Bull run loop of Process A, getNextJob moves a job from `wait` to `active` |  |
| 2 |  | In Process B, processStalledJob happens to run, pulls all active jobs (`this.client.lrangeAsync(this.toKey('active')`), and beings to iterate over them |
| 3 | processes the job normally | pulls job data via `Job.fromId` (the job data still exists at this point) |
| 4 | completes processing the job and the application code calls `job.remove()` which removes the job data and the lock |  |
| 5 |  | calls `scripts.getStalledJob` which only checks if the job is in `completed` (which it isn't anymore so it continues to grab the lock) |
| 6 |  | job is processed again |
| 7 |  | upon job completion, `Error: Missing Job 59694333 when trying to move from active to completed` is thrown because the job data doesn't exist anymore |

Perhaps a solution here is to have `scripts.getStalledJob` ensure the job is in the `active` state, not merely checking if it's not in `completed` - since now we know that it could have been removed prior to this check. So if a job is in `active` AND doesn't have an existing lock, then processStalledJobs knows that another worker isn't processing it. However, since the `active` queue is a LIST, checking the existence of the element in the list is expensive (requires list iteration in the lua script).

Our temporary workaround is to delay calling `job.remove()` at the completion of the job to leave around a 'tombstone' so `processStalledJobs` will see it in the completed queue and skip over it.

bradvogel:
…'t double-process jobs. See https://github.com/OptimalBits/bull/issues/356 for more on how this can happen.

The change here is to use a LUA script to atomically iterate the active queue AND get the lock. This ensures that the job is actually in 'active' when it's being processed, which fixes #356. If a stalled job was found, it will continue to call itself recursively until all stalled jobs are processed. This strategy is also slightly more efficient than before because it iterates the jobs all within a script, reducing the back-and-forth with the Redis server.

Additionally, this addresses long-standing issue where a job can be considered "stalled" even though it was just moved to active and before a lock could be obtained by the worker that moved it (https://github.com/OptimalBits/bull/issues/258), by waiting a grace period (with a default of LOCK_RENEW_TIME) before considering a job as possibly stalled. This gives the (real) worker time to acquire its lock after moving it to active.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -225,7 +225,6 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   // Bind these methods to avoid constant rebinding and/or creating closures
   // in processJobs etc.
   this.processStalledJobs = this.processStalledJobs.bind(this);
-  this.processStalledJob = this.processStalledJob.bind(this);
   this.getNextJob = this.getNextJob.bind(this);
   this.processJobs = this.processJobs.bind(this);
   this.processJob = this.processJob.bind(this);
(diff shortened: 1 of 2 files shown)


bradvogel:
Also, per https://github.com/OptimalBits/bull/pull/311#issuecomment-227990524, should a stalled job be moved back to 'wait' to be picked up again, versus what I'm doing here and processing it immediately? That would speed up processing these jobs immensely.

bradvogel:
Closing this in favor of https://github.com/OptimalBits/bull/pull/359 which puts jobs back into the 'wait' queue. It's much faster.

Issue #354, opened 2016-10-10
Removing completed jobs automatically & atomically


bradvogel:
It's our desire to remove jobs automatically upon completion. My understanding of the Bull API is that we should do this by listening to the `completed` event on the queue and call `job.remove()`, like this:

```
this.on('completed', job => job.remove());
```

However, we're finding that some completed jobs aren't getting removed. The likely reason for this is that the server is occasionally being killed right after the job is moved to "completed", but before `job.remove()` can be called. Since events are broadcast locally, there is no second server to pick up the message, the job stays around forever. Our current workaround is to run a cron job every hour that calls `clean(10 * 1000, 'completed')` on the queue.

I wanted to get your thoughts on the following possible approaches to clean up completed jobs:
1. Add a `removeOnCompleted` option to the job and have the `moveToCompleted` script atomically remove the job upon completion.
2. Use the new [distributed events architecture](https://github.com/OptimalBits/bull/commit/7b3859a49a435b33c6d3c55d683878afcf58b232) to broadcast the 'completed' event to all workers, so one can remove the job.

manast:
do you have any chance to test with master to see if it works better with the distributed events?
An atomic remove may be an interesting feature and not difficult to implement either.

bradvogel:
Not yet. Will test though. I think we should internalize removal through a new `removeOnComplete` job option, pending the outcome of bug https://github.com/OptimalBits/bull/issues/356 which might affect this.

bradvogel:
Fixed in https://github.com/OptimalBits/bull/pull/361/files

PR #361, opened 2016-10-15
Add new 'removeOnComplete' option to atomically remove a job when it …

bradvogel:
…is completed (useful for high-volume queues). Fixes https://github.com/OptimalBits/bull/issues/354

### DIFF
--- a/lib/scripts.js
+++ b/lib/scripts.js
@@ -134,7 +134,7 @@ var scripts = {
 
     return execScript.apply(scripts, args);
   },
-  moveToCompleted: function(job, token){
+  moveToCompleted: function(job, token, removeOnComplete){
     var keys = _.map([
       'active',
       'completed',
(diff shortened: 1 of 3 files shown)


lib/scripts.js
```diff
@@ -143,20 +143,23 @@ var scripts = {
...
-      '  return redis.call("del", KEYS[4])',
+      'if redis.call("EXISTS", KEYS[3]) == 1 then', // Make sure job exists
+      ' if redis.call("GET", KEYS[4]) == ARGV[3] then', // Makes sure we own the lock
+      '  redis.call("LREM", KEYS[1], 0, ARGV[1])',
+      removeOnComplete ? deleteJob : moveJob,
+      '  redis.call("DEL", KEYS[4])',
+      '  return 0',
+      ' else',
+      '  return -1',
       ' end',
```
manast: A slight issue here, you need to name differently the script when you specify removeOnComplete, otherwise, since scripts are cached, if you do not use removeOnComplete in a future instance, it will use the old cached one. I know this is probably not going to happen, but for the sake of formality :)

bradvogel: Oh yes, great catch. Pushed up a fix. Also pushed a small microoptimization for queues with a large number of active jobs (like ours)

bradvogel:
Let me know if there's anything else before mergning

manast:
A future improvement of this feature could be a setting to specify how many jobs to keep. By doing that you could keep a history of the latest x processed jobs, which can be useful info, while not keeping all of them which in same cases is not practical.

bradvogel:
Yes, good idea on that.

PR #188, opened 2015-09-27
Migrating from node_redis to ioredis.

kulbirsaini:
Related Issue: https://github.com/OptimalBits/bull/issues/185

I have started this migration from node_redis to ioredis. Please review and suggest changes.

**Pending tasks**
- [ ] Improve/fix tests
- [ ] Update Readme
- [ ] Bump version since breaking change

**Status**
All tests passing with a few timing out when run with `npm test`.

*Borrows changes from PR #158

PR #375, opened 2016-11-11
[fix] don't try to remove listeners if already removed

doublerebel:
Fixes unhandled rejection SISMEMBER errors as seen in build https://travis-ci.org/OptimalBits/bull/builds/174710354

#### Details
While working on #188 I noticed some errors in the test suite.  The `Job::finished` watchdog was running even when the listeners were already removed.  This PR clears the watchdog interval so that the watchdog is not run unnecessarily.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -353,11 +353,13 @@ Job.prototype.finished = function(){
   return new Promise(function(resolve, reject){
     status(resolve, reject).then(function(finished){
       if(!finished){
+        var interval;
         function onCompleted(job){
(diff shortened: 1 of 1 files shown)


manast:
good catch!

lib/job.js
```diff
@@ -378,7 +381,7 @@ Job.prototype.finished = function(){
         //
```
manast: after merging I noticed that clearInterval could be called inside removeListeners to avoid one line of code :D

Issue #370, opened 2016-11-09
Jobs can be processed and left in the `wait` state


bradvogel:
@manast Following the discussion in https://github.com/OptimalBits/bull/pull/359, I found a hazard that I think we should address before releasing a new version: a job can be processed an still left in wait. Here's how:

| Time | Process A | Process B |
| --- | --- | --- |
| 1 | In the regular Bull run loop of Process A, getNextJob moves a job from `wait` to `active` |  |
| 2 |  | In Process B, `moveUnlockedJobsToWait` happens to run and pick up the job in step 1, and moves it back to wait |
| 3 | Process A gets lock on the job (that is now in 'wait') and begins to process it | |

If there is no other process around to move the job back to `active` then Process A will complete the job and leave it in `wait`. This can lead to a data inconsistency.

bradvogel:
To fix this, what if `takeLock` takes a parameter (perhaps `ensureActive`) to ensure that the job is in the active state? Then `processJob` would use that parameter - so it can only ever keep a lock while the job is in `active` - while other uses of the lock (eg `job.remove` is called) can acquire the lock in any state.

manast:
I would need to check the code but is there any reason to allow locking a job that it is not in the active set? If now, then we can just have that constraint in takeLock, it could fail the same way it does when trying to get a lock on a already locked job...

bradvogel:
`job.remove()` tries to get the lock before removing, and I assume we still want to be able to call `remove()` on jobs that aren't in the active queue

bradvogel:
#371 describes a more dangerous case: a job can get double processed because a worker is allowed to take the lock while the job is in `completed`, cc @doublerebel

doublerebel:
@bradvogel I think I understand how this could happen, do you have a test case to reproduce this issue?  It doesn't seem to be covered by the test suite.

Also is there a reason not to lock a job from leaving wait until failed/complete?  Rather than releasing the lock between the move to active and when processing starts.

bradvogel:
I don't have a test case yet. It's a bit difficult to write since this is a subtle race condition. Your test case in https://github.com/OptimalBits/bull/issues/371#issuecomment-260158407 should cover it though.

We can't lock the job atomically while it's being moved from wait to active. We use the Redis operation `brpoplpush` to move the job, but due to redis limitations (described in https://github.com/OptimalBits/bull/issues/258), `brpoplpush` can't be used inside a lua script that we'd need to atomically lock it while moving.

[...]

PR #377, opened 2016-11-13
Fixes double-processing issue described in https://github.com/Optimal…

bradvogel:
…Bits/bull/issues/371#issuecomment-260158407.

Double-processing happens when two workers find out about the same job at the same time via `getNextJob`. One worker is taking the lock, processing the job, and moving it to completed before the second worker can even try to get the lock. When the second worker finally gets around to trying to get the lock, the job is already in the completed state. But it processes it anyways since it got the lock.

So the fix here is for the takeLock script to ensure the job is in the active queue prior to taking the lock. That will make sure jobs that are in wait, completed, or even removed from the queue altogether don't get double processed. Per the discussion in #370 though, takeLock is parameterized to only require the job be in active when taking the lock while processing the job. There are other cases such as job.remove() that the job might be in a different state, but we still want to be able to lock it.

This fixes existing existing broken unit test "should process each job once".

This also prevents hazard https://github.com/OptimalBits/bull/issues/370.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -619,6 +619,8 @@ Queue.prototype.processJob = function(job, renew){
         renew = true;
         lockRenewId = _this.timers.set('lockRenewer', _this.LOCK_RENEW_TIME / 2, lockRenewer);
       }
+      // TODO: if we failed to re-acquire the lock while trying to renew, should we let the job
+      // handler know and cancel the timer?
       return locked;
     }, function(err){
       console.error('Error renewing lock ' + err);
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -614,7 +614,10 @@ Queue.prototype.processJob = function(job, renew){
   // by another worker. See #308
   //
   var lockRenewer = function(){
-    return job.takeLock(_this.token, renew, true /* Ensure in active */).then(function(locked){
+    // The first call to lock the job should ensure that the job is in the 'active' state,
+    // because it might have gotten picked up already by another processor. We don't need
+    // to do this on subsequent calls.
+    return job.takeLock(_this.token, renew, !renew).then(function(locked){
       if(locked){
         renew = true;
         lockRenewId = _this.timers.set('lockRenewer', _this.LOCK_RENEW_TIME / 2, lockRenewer);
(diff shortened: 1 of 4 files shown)


doublerebel:
Excellent, I'll give this a shot with my ioredis branch.

bradvogel:
This isn't quite ready yet. Still investigating a flaky test failure.

bradvogel:
Ok @manast this is now ready for review.

lib/queue.js
```diff
@@ -619,6 +619,8 @@ Queue.prototype.processJob = function(job, renew){
         renew = true;
         lockRenewId = _this.timers.set('lockRenewer', _this.LOCK_RENEW_TIME / 2, lockRenewer);
       }
+      // TODO: if we failed to re-acquire the lock while trying to renew, should we let the job
+      // handler know and cancel the timer?
       return locked;
```
manast: When this happens, it must be that the job has been locking the event loop for too long. Could we fail here? but some other worker is already processing the job, should we abort (but we cant do that either)? if we do nothing there is a risk a malfunctioning job keeps being stollen by workers and never complete (or fail)

lib/queue.js
```diff
@@ -614,7 +614,10 @@ Queue.prototype.processJob = function(job, renew){
   // by another worker. See #308
   //
   var lockRenewer = function(){
-    return job.takeLock(_this.token, renew, true /* Ensure in active */).then(function(locked){
+    // The first call to lock the job should ensure that the job is in the 'active' state,
+    // because it might have gotten picked up already by another processor. We don't need
+    // to do this on subsequent calls.
+    return job.takeLock(_this.token, renew, !renew).then(function(locked){
       if(locked){
```
manast: nice optimization!

manast:
This looks great and also green 💃

lib/queue.js
bradvogel: Ideally we tell the job processor somehow that the job is now aborted. It's not really a "failure" in the bull sense because the job is still be worked on by the second worker. Perhaps we introduce a new "aborted" event and ask job processors to listen to it and abort their processing if one of their jobs gets stolen. Though if a job gets stolen, then it means that processor is behind in its event loop, so it's probably not going to get that event anyways.

Issue #333, opened 2016-08-12
Cleaning Unhandled rejection Error


Bersam:
I have multiple job queues, got big as 10K items in each. Also I have a cron job that at first pause a queue then it tries to clean completed jobs and resume queue again, but I get this error randomly, here's part of code: [https://gist.github.com/Bersam/c46148decf0acf4b8544a66f13f11bcd](clean.js)

```
 Unhandled rejection Error: Could not get lock for job: 3. Cannot remove job.
     at /var/www/instafollow/node_modules/bull/lib/job.js:292:13
     at tryCatcher (/var/www/instafollow/node_modules/bluebird/js/release/util.js:16:23)
     at Promise._settlePromiseFromHandler (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:504:31)
     at Promise._settlePromise (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:561:18)
     at Promise._settlePromise0 (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:606:10)
     at Promise._settlePromises (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:685:18)
     at Promise._fulfill (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:630:18)
     at Promise._resolveCallback (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:425:57)
     at Promise._settlePromiseFromHandler (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:516:17)
     at Promise._settlePromise (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:561:18)
     at Promise._settlePromise0 (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:606:10)
     at Promise._settlePromises (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:685:18)
     at Promise._fulfill (/var/www/instafollow/node_modules/bluebird/js/release/promise.js:630:18)
     at Command.callback (/var/www/instafollow/node_modules/bluebird/js/release/nodeback.js:42:21)
     at normal_reply (/var/www/instafollow/node_modules/redis/index.js:714:21)
     at RedisClient.return_reply (/var/www/instafollow/node_modules/redis/index.js:816:9)
     at JavascriptRedisParser.Parser.returnReply (/var/www/instafollow/node_modules/redis/index.js:188:18)
[...]

PR #379, opened 2016-11-14
ioredis; atomic locking with redlock; support Node 6.x [Fixes #185, Fixes #370]

### LINKED ISSUE #185: Change from node_redis to ioredis
So it can support Redis 3 cluster

### LINKED ISSUE #370: Jobs can be processed and left in the `wait` state

doublerebel:
Hello,

This PR changes the redis dependency from **node_redis** to **ioredis**, fixing #185 (and completing PR #188).  **ioredis** uses promises "all-the-way-down", exposing errors in the test suite.  These tests are fixed and passing, including the race condition described in #370.

There are several commits here, to demonstrate the bugs and fixes made to the test suite.  I am happy to squash this down, but right now each commit contains a description of the issues solved with the updated tests.

As part of solving #370, [**node-redlock**](https://github.com/mike-marcacci/node-redlock) is now used to implement atomic locking with a guaranteed owner.  **node-redlock** is a promise-based lib which keeps track of its own secret token/value for each job/resource, which makes node-redlock valuable whether or not it's used with redis clustering.

Thanks for your work, looking forward to your feedback.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1,7 +1,7 @@
 /*eslint-env node */
 'use strict';
 
-var redis = require('redis');
+var redis = require('ioredis');
 var Disturbed = require('disturbed');
 var util = require('util');
 var assert = require('assert');
@@ -15,8 +15,6 @@ var uuid = require('node-uuid');
 var semver = require('semver');
 var debuglog = require('debuglog')('bull');
 
-Promise.promisifyAll(redis.RedisClient.prototype);
-Promise.promisifyAll(redis.Multi.prototype);
 
 /**
   Gets or creates a new Queue with the given name.
@@ -71,7 +69,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
     redisPort = redisOpts.port;
     redisHost = redisOpts.host;
     redisOptions = redisOpts.opts || {};
-    redisOptions.db = redisOpts.DB;
+    redisOptions.db = redisOpts.db || redisOpts.DB;
   } else if(_.isString(redisPort)) {
     try {
       var redisUrl = url.parse(redisPort);
@@ -386,8 +384,8 @@ Queue.prototype.count = function(){
   multi.llen(this.toKey('paused'));
   multi.zcard(this.toKey('delayed'));
 
-  return multi.execAsync().then(function(res){
-    return Math.max(res[0], res[1]) + res[2];
+  return multi.exec().then(function(res){
+    return Math.max(res[0][1], res[1][1]) + res[2][1];
   });
 };
 
@@ -415,7 +413,9 @@ Queue.prototype.empty = function(){
   multi.del(this.toKey('meta-paused'));
   multi.del(this.toKey('delayed'));
 
-  return multi.execAsync().spread(function(waiting, paused){
+  return multi.exec().spread(function(waiting, paused){
+    waiting = waiting[1];
+    paused = paused[1];
     var jobKeys = (paused.concat(waiting)).map(_this.toKey, _this);
 
     if(jobKeys.length){
@@ -696,9 +696,7 @@ Queue.prototype.getNextJob = function(opts){
 };
 
 Queue.prototype.multi = function(){
-  var multi = this.client.multi();
-  multi.execAsync = Promise.promisify(multi.exec);
-  return multi;
+  return this.client.multi();
 };
 
 /**
@@ -765,8 +763,8 @@ Queue.prototype.getJobCountByTypes = function() {
     }
   });
 
-  return multi.execAsync().then(function(res){
-    return _.reduce(res, function(total, n) {
+  return multi.exec().then(function(res){
+    return _.reduce(res[0], function(total, n) {
       return total + n;
     }) || 0;
   });
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -256,7 +256,9 @@ Queue.prototype.distEmit = function(){
 
 Queue.prototype.on = function(){
   var args = Array.prototype.slice.call(arguments);
-  Disturbed.prototype.on.apply(this, args);
+  var promise = Disturbed.prototype.on.apply(this, args);
+  var _this = this;
+  promise.catch(function(err){ _this.emit('error', err); });
   return this;
 };
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -524,11 +524,7 @@ Queue.prototype.run = function(concurrency){
       promises.push(new Promise(_this.processJobs));
     }
 
-    if (_this.STALLED_JOB_CHECK_INTERVAL > 0){
-      clearInterval(_this.moveUnlockedJobsToWaitInterval);
-      _this.moveUnlockedJobsToWaitInterval =
-        setInterval(_this.moveUnlockedJobsToWait, _this.STALLED_JOB_CHECK_INTERVAL);
-    }
+    _this.startMoveUnlockedJobsToWait();
 
     return Promise.all(promises);
   });
@@ -576,26 +572,30 @@ Queue.prototype.updateDelayTimer = function(newDelayedTimestamp){
 Queue.prototype.moveUnlockedJobsToWait = function(){
   var _this = this;
 
-  if(this.closing){
-    return this.closing;
-  } else{
-    return scripts.moveUnlockedJobsToWait(this).then(function(responses){
-      var handleFailedJobs = responses[0].map(function(jobId){
-        return _this.getJobFromId(jobId).then(function(job){
-          _this.distEmit('failed', job.toJSON(), new Error('job stalled more than allowable limit'));
-          return null;
-        });
+  return scripts.moveUnlockedJobsToWait(this).then(function(responses){
+    var handleFailedJobs = responses[0].map(function(jobId){
+      return _this.getJobFromId(jobId).then(function(job){
+        _this.distEmit('failed', job.toJSON(), new Error('job stalled more than allowable limit'));
+        return null;
       });
-      var handleStalledJobs = responses[1].map(function(jobId){
-        return _this.getJobFromId(jobId).then(function(job){
-          _this.distEmit('stalled', job.toJSON());
-          return null;
-        });
-      });
-      return Promise.all(handleFailedJobs.concat(handleStalledJobs));
-    }).catch(function(err){
-      console.error('Failed to handle unlocked job in active:', err);
     });
+    var handleStalledJobs = responses[1].map(function(jobId){
+      return _this.getJobFromId(jobId).then(function(job){
+        _this.distEmit('stalled', job.toJSON());
+        return null;
+      });
+    });
+    return Promise.all(handleFailedJobs.concat(handleStalledJobs));
+  }).catch(function(err){
+    console.error('Failed to handle unlocked job in active:', err);
+  });
+};
+
+Queue.prototype.startMoveUnlockedJobsToWait = function() {
+  if (this.STALLED_JOB_CHECK_INTERVAL > 0){
+    clearInterval(this.moveUnlockedJobsToWaitInterval);
+    this.moveUnlockedJobsToWaitInterval =
+      setInterval(this.moveUnlockedJobsToWait, this.STALLED_JOB_CHECK_INTERVAL);
   }
 };
 
@@ -707,7 +707,11 @@ Queue.prototype.processJob = function(job){
   Returns a promise that resolves to the next job in queue.
 */
 Queue.prototype.getNextJob = function(opts){
-  return this.moveJob('wait', 'active', opts).then(this.getJobFromId);
+  if(!this.closing){
+    return this.moveJob('wait', 'active', opts).then(this.getJobFromId);
+  }else{
+    return Promise.reject();
+  }
 };
 
 Queue.prototype.multi = function(){
@@ -927,13 +931,18 @@ Queue.prototype.whenCurrentJobsFinished = function(){
         resolve();
       }else{
         resolver = _.after(count, function(){
+          _this.removeListener('stalled', resolver);
           _this.removeListener('completed', resolver);
           _this.removeListener('failed', resolver);
+          clearInterval(_this.moveUnlockedJobsToWaitInterval);
           resolve();
         });
 
+        _this.on('stalled', resolver);
         _this.on('completed', resolver);
         _this.on('failed', resolver);
+
+        _this.startMoveUnlockedJobsToWait();
       }
     }, reject);
   });
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -87,7 +87,6 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   }
 
   redisOptions = redisOptions || {};
-  var redisDB = redisOptions.db || 0;
 
   function createClient() {
     var client;
@@ -149,14 +148,17 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   this.timers = new TimerManager();
 
   // emit ready when redis connections ready
-  this._initializing = Promise.join(
-    this.client.selectAsync(redisDB),
-    this.bclient.selectAsync(redisDB),
-    this.eclient.selectAsync(redisDB)
-  ).then(function(){
+  var initializers = [this.client, this.bclient, this.eclient].map(function (client) {
+    return new Promise(function(resolve) {
+      client.once('ready', resolve);
+    });
+  });
+
+  this._initializing = Promise.all(initializers)
+  .then(function(){
     return Promise.join(
-      _this.eclient.subscribeAsync(_this.toKey('delayed')),
-      _this.eclient.subscribeAsync(_this.toKey('paused'))
+      _this.eclient.subscribe(_this.toKey('delayed')),
+      _this.eclient.subscribe(_this.toKey('paused'))
     );
   }).then(function(){
     debuglog(name + ' queue ready');
@@ -282,8 +284,8 @@ Queue.prototype.disconnect = function(){
   }
 
   return Promise.join(
-    _this.client.quitAsync(),
-    _this.eclient.quitAsync()
+    _this.client.quit(),
+    _this.eclient.quit()
   ).then(endClients, endClients);
 };
 
@@ -422,7 +424,7 @@ Queue.prototype.empty = function(){
       multi = _this.multi();
 
       multi.del.apply(multi, jobKeys);
-      return multi.execAsync();
+      return multi.exec();
     }
   });
 };
@@ -491,7 +493,7 @@ function pauseResumeGlobal(queue, pause){
     return queue.toKey(name);
   });
 
-  return queue.client.evalAsync(script, keys.length, keys[0], keys[1], keys[2], keys[3], pause ? 'paused' : 'resumed');
+  return queue.client.eval(script, keys.length, keys[0], keys[1], keys[2], keys[3], pause ? 'paused' : 'resumed');
 }
 
 Queue.prototype.run = function(concurrency){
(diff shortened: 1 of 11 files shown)


lib/queue.js
```diff
@@ -147,11 +147,16 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
...
+  var initializers = [];
+  [this.client, this.bclient, this.eclient].forEach(function (client) {
+    var promise = new Promise(function(resolve) {
+      client.once('ready', resolve);
+    });
+    initializers.push(promise);
+  });
+
+  this._initializing = Promise.join.apply(null, initializers)
+  .then(function(){
```
manast: here we could also use ```Promise.all(initializers)``` for more succinct code.

lib/queue.js
```diff
@@ -147,11 +147,16 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   this.timers = new TimerManager();
 
   // emit ready when redis connections ready
-  this._initializing = Promise.join(
-    this.client.select(redisDB),
-    this.bclient.select(redisDB),
-    this.eclient.select(redisDB)
-  ).then(function(){
+  var initializers = [];
+  [this.client, this.bclient, this.eclient].forEach(function (client) {
```
manast: use .map instead of forEach for a shorter and more semantically correct code.

test/test_cluster.js
```diff
@@ -44,18 +44,21 @@ function workerMessageHandlerWrapper(message) {
...
+    var _i = 0;
+    for(_i; _i < os.cpus().length - 1; _i++) {
+      worker = cluster.fork();
+      worker.on('message', workerMessageHandlerWrapper);
+      workers.push(worker);
+      console.log('Worker spawned: #', worker.id);
+    }
+  });
+
   var queue;
```
manast: great improvement!

lib/queue.js
```diff
@@ -257,7 +257,9 @@ Queue.prototype.distEmit = function(){
 
 Queue.prototype.on = function(){
   var args = Array.prototype.slice.call(arguments);
-  Disturbed.prototype.on.apply(this, args);
+  var promise = Disturbed.prototype.on.apply(this, args);
+  var _this = this;
+  promise.catch(function(err){ _this.emit('error', err); });
   return this;
```
manast: I changed the API in disturbed so that it will handle local events as well which will break this. I'll fix it asap.

lib/scripts.js
```diff
@@ -304,6 +304,7 @@ var scripts = {
    * Takes a lock
    */
   takeLock: function(queue, job, token, renew){
+    console.log('attempting lock for job id #', job.jobId)
     var lockCall;
```
manast: we may need a debug flag to switch on/off this debug logs.

lib/scripts.js
```diff
@@ -185,7 +180,9 @@ var scripts = {
       token
     ];
 
-    return execScript.apply(scripts, args).then(function(result){
+    return job.takeLock(null, !!job.lock)
+    .then(function(lock) { return lock ? execScript.apply(scripts, args) : -2; }, function(err) { if (!(err instanceof Redlock.LockError)) { throw err; } })
+    .then(function(result){
       switch (result){
```
manast: this could be formatted into several lines or refactored into a separate function

lib/scripts.js
```diff
@@ -303,59 +302,28 @@ var scripts = {
...
-      queue.LOCK_RENEW_TIME
-    ];
 
-    return execScript.apply(scripts, args);
+    var lock = job.lock;
+    if (renew && !lock) {
+      throw new Error('Unable to renew nonexisting lock');
+    } else if (renew) {
+      return lock.extend(queue.LOCK_RENEW_TIME);
+    } else if (lock) {
```
manast: I think the if here is not needed, lock must be defined if we reached this point.

lib/scripts.js
```diff
@@ -303,59 +302,28 @@ var scripts = {
...
+      return new Promise(function(resolve) {
+        resolve(null);
```
manast: just use ```return Promise.resolve(null);```

lib/scripts.js
```diff
@@ -315,7 +315,8 @@ var scripts = {
       });
     }
     var redlock = new Redlock(queue.clients, queue.redlock);
-    return redlock.lock(job.lockKey(), queue.LOCK_RENEW_TIME);
+    return redlock.lock(job.lockKey(), queue.LOCK_RENEW_TIME)
+    .then(job.queue.client.hset(job.queue.toKey(job.jobId), "lockAcquired", 1));
   },
```
manast: the problem here is that getting the lock and setting the lockAcquired property is not an atomic operation anymore...

lib/scripts.js
```diff
@@ -314,9 +314,31 @@ var scripts = {
...
+
+    var redlock;
+    if (ensureActive) {
+      var isJobInList = this._isJobInList.replace(/KEYS\[1\]/, '"' + job.queue.toKey('active') + '"')
+        .replace('return item', 'if item')
+        .replace(/ARGV\[1\]\)/, '"' + job.jobId + '")');
+      var lockAcquired = 'and redis.call("HSET", "' + queue.toKey(job.jobId) + '", "lockAcquired", "1")';
+      var success = 'then return 1 else return 0 end';
+      var opts = {
+        lockScript: function(lockScript) {
```
manast: ok, so here we can actually "monkey patch" the redlock code. I was actually wondering how we could solve the issue with ensuring that the job is in the active list while getting a redlock atomically. But how robust is this when new versions of redlock are released? Isn't there a risk that our code would stop working?

manast:
great PR. A Lot of good stuff all over the place. Please check my comments. I am sure @bradvogel will also give you some valuable feedback.

lib/scripts.js
doublerebel: This log is removed in a later commit, I'll make sure I have no such console.* statements in the final PR.

lib/scripts.js
doublerebel: You're right, I'll clean this up.

lib/scripts.js
doublerebel: In this case, if the lock is defined we can return null early and not need to call redlock.

lib/queue.js
```diff
@@ -581,26 +577,30 @@ Queue.prototype.moveUnlockedJobsToWait = function(){
...
-    return this.closing;
-  } else{
-    return scripts.moveUnlockedJobsToWait(this).then(function(responses){
-      var handleFailedJobs = responses[0].map(function(jobId){
-        return _this.getJobFromId(jobId).then(function(job){
-          _this.distEmit('failed', job.toJSON(), new Error('job stalled more than allowable limit'));
-          return null;
-        });
+  return scripts.moveUnlockedJobsToWait(this).then(function(responses){
+    var handleFailedJobs = responses[0].map(function(jobId){
```
doublerebel: Ooh, good catch.

doublerebel:
Thanks @manast.  I really appreciate all the Promise API improvements, I'm not as familiar with it.  I'll rebase this PR, clean it up and repush.

lib/queue.js
doublerebel: I tried this, and it seems that spread must have an array, but in bull sometimes scripts.moveUnlockedJobsToWait returns undefined:
```js
 Cluster
Worker spawned: # 1
Worker spawned: # 2
Worker spawned: # 3
Worker spawned: # 4
Worker spawned: # 5
Worker spawned: # 6
Worker spawned: # 7
Failed to handle unlocked job in active: TypeError: Cannot read property 'map' of undefined
    at /home/charles/source/bull/lib/queue.js:576:40
    at tryCatcher (/home/charles/source/bull/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/home/charles/source/bull/node_modules/bluebird/js/release/promise.js:507:35)
    at Promise._settlePromise (/home/charles/source/bull/node_modules/bluebird/js/release/promise.js:567:18)
    at Promise._settlePromise0 (/home/charles/source/bull/node_modules/bluebird/js/release/promise.js:612:10)
    at Promise._settlePromises (/home/charles/source/bull/node_modules/bluebird/js/release/promise.js:691:18)
```

doublerebel:
I rebased, and cleaned up all the commit based on your feedback.  The previous set of commits is saved at [nextorigin/bull/commits/issue-185-dirty](https://github.com/nextorigin/bull/commits/issue-185-dirty).

Now there are two major commits: the ioredis refactor, and the redlock refactor, almost all the rest are test improvements.  I get all tests passing on every run here, so I hope that is reflected shortly in travis 😃

doublerebel:
With the lock and promise improvements, this PR might fix #333 as well.  Leaving this note for future reference.

package.json
```diff
@@ -20,10 +20,11 @@
     "bluebird": "^3.4.6",
     "debuglog": "^1.0.0",
     "disturbed": "^1.0.3",
+    "ioredis": "^2.4.0",
     "lodash": "^4.16.6",
     "mocha": "^3.1.2",
     "node-uuid": "^1.4.7",
-    "redis": "^2.6.3",
+    "redlock": "nextorigin/node-redlock",
```
bradvogel: I don't think Bull should have this dependency. Please try to upstream your fix into node-redlock.

lib/scripts.js
```diff
@@ -305,93 +311,51 @@ var scripts = {
...
+    } else if (lock) {
+      return Promise.resolve(null);
+    }
 
-    var lockCall;
-    if (renew){
-      lockCall = 'redis.call("SET", KEYS[1], ARGV[1], "PX", ARGV[2])';
+    var redlock;
+    if (ensureActive) {
+      var isJobInList = this._isJobInList.replace(/KEYS\[1\]/, '"' + job.queue.toKey('active') + '"')
```
bradvogel: This is too fragile to be running string replacement. Please make `_isJobInList` a function that generates the appropriate string.

bradvogel:
What about all the scripts that look for `:lock` in the key such as `moveUnlockedJobsToWait`? Won't these break?

lib/scripts.js
```diff
@@ -305,93 +311,51 @@ var scripts = {
...
-      'if (found == false) then',
-      '  return 0',
-      'end'
-    ].join('\n');
+  takeLock: function(queue, job, renew, ensureActive){
+    var lock = job.lock;
+    if (renew && !lock) {
+      throw new Error('Unable to renew nonexisting lock');
+    } else if (renew) {
+      return lock.extend(queue.LOCK_RENEW_TIME);
```
bradvogel: The lock renewer runs on `_this.LOCK_RENEW_TIME / 2` interval, so this will keep extending it out farther and farther on each renewal. So if it gets stalled, it'll potentially be locked for many minutes. Is there a way to extend it to the absolute time `(currentTime + queue.LOCK_RENEW_TIME)`?

doublerebel:
@bradvogel they actually work exactly the same, redlock's method is very similar to bull's method and it works with `moveUnlockedJobsToWait` as a drop-in replacement.

We can see `moveUnlockedJobsToWait` working in some of the tests, the long process time is the stalled script running twice to move the jobs here:
https://travis-ci.org/OptimalBits/bull/jobs/175763931#L823

Also in my previous runs where I was logging heavily, I logged that process in particular so I could ensure it was working, the `moveUnlockedJobsToWait` calls are logged here:
https://travis-ci.org/nextorigin/bull/jobs/175597819#L1652

lib/scripts.js
doublerebel: You're right, I'll clean that up.

lib/scripts.js
doublerebel: The [redlock extend method uses PEXPIRE](https://github.com/nextorigin/node-redlock/blob/master/redlock.js#L16), and should be functionally equivalent to the SET ... PX method bull uses now.  It only sets the expire from the current time, it does not add to the existing TTL, so it's typical redis expire behavior.

bradvogel:
Ok, I see. I verified moveUnlockedJobsToWait still works locally. Thanks!

lib/scripts.js
bradvogel: Oh yes, you're right. Filed https://github.com/mike-marcacci/node-redlock/issues/26 to clarify their docs.

doublerebel:
@bradvogel For sure, I appreciate the push for code quality.  I've rewritten `._isJobInList` in 3a41230 and re-pushed.

bradvogel:
btw, I'm not completely familiar with Redis cluster, but a potential hazard that might cause issues with cluster is how we concatenate job keys in some scripts: https://github.com/nextorigin/bull/blob/issue-185/lib/scripts.js#L439. It's my understanding that if keys aren't passed in to scripts via the `KEYS` argument, then there's no guarantee that they will run on the right cluster (http://redis.io/commands/EVAL). So we might need to rethink how we run these scripts. Anyways, outside the scope of this PR.

doublerebel:
@bradvogel good call, I'll make a note of that.  Perhaps we can ask for some user stories for bull+redis cluster usage from those issues.  I run a lot of jobs, and have been using it with Sentinel+Redishappy+Consul for failover, but not a redis cluster per se.

lib/queue.js
```diff
@@ -615,17 +637,14 @@ Queue.prototype.processJob = function(job){
   //
   var renew = false;
   var lockRenewer = function(){
-    // The first call to lock the job should ensure that the job is in the 'active' state,
-    // because it might have gotten picked up already by another processor. We don't need
-    // to do this on subsequent calls.
-    return job.takeLock(_this.token, renew, !renew).then(function(locked){
-      if(locked){
+    return job.takeLock(renew, true).then(function(lock){
```
bradvogel: Mismerge. The second parameter should be `!renew`. Also include comment

doublerebel: Here [`takeLock` returns early](https://github.com/nextorigin/bull/blob/5c1e9d9b5e8e480050b558c197e37a48210b924c/lib/scripts.js#L327) and relies on redlock to extend the lock, `ensureActive` is only considered for new locks.  So I don't know if insinuating the second parameter matters on a renew is necessary?

lib/scripts.js
manast: yes, I understand that, what I mean is that lock will always defined in line 311, so you don't need the if, the else is enough.

manast:
please let me know if everybody is happy so that I can manually merge this in the morning swedish time.

albertjan:
Looks brilliant, does this introduce a new locking mechanism? If so is there some docs on that because I'll need to update fsprojects/oxen to stay compatible with bull. :)

bradvogel:
I'll give it one more thorough review in a few hours.

@manast do you think this should be released as 2.0? While technically none of the APIs documented in the readme have changed with this PR, anyone using undocumented `createClient` (https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L96) will now need to create an ioredis client. What do you think?

bradvogel:
Ran a few more tests and this looks good to me. There's a merge conflict though.

doublerebel:
Must be from today's new commits on master.  I can rebase again first thing tomorrow.

manast:
I guess it is safer to release it as 2.0, to avoid the risk of breaking users code. I can write a notice explaining that the changes are mostly internal but large enough to motivate a major release.

lib/scripts.js
```diff
@@ -305,93 +316,51 @@ var scripts = {
...
+            lockScript.replace('return', 'and'),
+            lockAcquired,
+            success
+          ].join('\n');
+        }
+      };
+      redlock = new Redlock(queue.clients, _.extend(opts, queue.redlock));
     } else {
-      lockCall = 'redis.call("SET", KEYS[1], ARGV[1], "PX", ARGV[2], "NX")';
+      redlock = new Redlock(queue.clients, queue.redlock);
```
bradvogel: I personally think it's cleaner to monkey patch Redlock (eg `redlock.lockScript = ...`) for now until PR https://github.com/mike-marcacci/node-redlock/pull/25 is resolved. To mitigate against Redlock changing, just lock the version in the package.json (so `"redlock": "2.0.2"`).

manast:
@albertjan it uses a different locking mechanism (redlock) that is supposed to work in a distributed environment (needed when running redis cluster). I wondering though if we are going to announce that bull works with redis cluster, I have not made any tests myself, and as mentioned before, our LUA scripts may not work well in such an environment.

doublerebel:
@albertjan 
Redlock algo is almost exactly like bull's current lock logic, it's the technique described in [redis distlock](http://redis.io/topics/distlock).  node-redlock keeps track of its own per-job secret, so the javascript process owns the lock and checks on renew just the same.  To extend the lock, node-redlock uses `PEXPIRE` where bull uses `SET ... PX`, but that is just syntax.  It's convenient to have the lock packaged in a promise-based, tested codebase.

Part of the reason to implement ioredis is to start exploring redis sentinel and cluster support, however as described in [redis distlock](http://redis.io/topics/distlock#why-failover-based-implementations-are-not-enough), bull will lose lock guarantees in a failover scenario unless redlock is implemented.  node-redlock manages those cluster guarantees as well.

There are some .NET libraries for redlock [listed in redis distlock](http://redis.io/topics/distlock#implementations).  Bull is now doing up to 3 operations in one block to get a lock:
* check to see if job is in active list
* set the lock
* set the lockAcquired counter

Unfortunately it looks like none of the .NET libs have the lock script as a publicly accessible or modify-able variable, I have a [PR open for node-redlock](https://github.com/mike-marcacci/node-redlock/issues/25) for that functionality.

lib/scripts.js
doublerebel: Unfortunately it's a private variable without my patch.  I've petitioned for an answer from the dev, I can fork it if necessary, I maintain forks of other packages too.

lib/scripts.js
doublerebel: I definitely don't need else after return or throw.  Updated in this latest push.

doublerebel:
Rebased on master and re-pushed.  Fingers crossed for node-redlock#25.

lib/scripts.js
bradvogel: Yeah, that's what I mean - overwrite the private variable. Lock the version down in package.json (eg `"redlock": "2.0.2"`) to make sure that variable won't change on us. Then in code:

```
var redlock = new Redlock(..)
redlock.lockScript = redlock.lockScript.replace(...)
```

This is much safer because it still uses the real `redlock` module, and a locked version that can't update. It's clearer in package.json so other devs can see what version we're using. You can even add a comment in package.json (http://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install) to indicate why it's locked to a specific version.

Pointing this to `nextorigin/node-redlock` is dangerous because if you or anyone on your team accidentally pushed to that master branch, it'll break Bull for **all** users. I trust that you wouldn't do that, but it's unnecessary risk.

cc @manast

doublerebel:
@bradvogel I totally agree about pulling packages from Github, I would have released a new package like I did with stream-combiner2-withopts.  The lockscript was private like trapped in a closure private, but thankfully the PR was merged.  I swear 90% of the time I have to maintain a forked package, it's just because some option is trapped in a closure.

The only change in this latest push is to update redlock to `^2.1.0` in package.json.  That should complete it unless you guys caught anything else!

PR #380, opened 2016-11-14
373 fix creating queue with port as string

liquorvicar:


manast:
thanks for the fix, I actually pushed a fix myself: https://github.com/OptimalBits/bull/commit/69f91497199c0c382ab0ab4fb17fe473cea8902f
what do you think, which one should be use?

liquorvicar:
Personally I prefer the fix in this PR as it's clearer and displays the intention of the code better (inspired by @shane-walker 's comment). Plus it has the added bonus of a test for this scenario. But it's your call as maintainer of the project. I won't get upset if you choose not to use this one ;-)

PR #395, opened 2016-11-28
Add support for rate limiting

### LINKED ISSUE #329: rate limiting?
Is it possible to limit a job (keyed on an argument) to only run x times a period? does bull have a plugin framework so commonly used functions can be shared across jobs?

### LINKED ISSUE #362: Add a rate limiter so that the queue does not process more than N jobs per second.

leahciMic:
Fix #362
Fix #329

This commit adds support for rate limiting of a queue. The rate limit applies to each instance of the queue. If you have 3 threads processing the queue, and you want to limit your jobs to 1 per second, it will be 1 job per second amongst the 3 threads.

`queue#process` first argument has been changed to an options object, and concurrency now resides within this object, as well as the new rate limiting options. (If a number is specified as the first argument to `queue#process`, it will be treated as the old version of the API which I
have deprecated in favor of passing in the options object).

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -385,7 +385,7 @@ Queue.prototype.createRateLimiter = function(rateLimit){
   var limiter = Promise.promisify(Limiter({
     interval: rateLimit.duration,
     maxInInterval: rateLimit.max,
-    redis: this.client,
+    redis: createClient(),
     namespace: this.keyPrefix + ':ratelimit:'
   }));
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -13,6 +13,7 @@ var _ = require('lodash');
 var Promise = require('bluebird');
 var semver = require('semver');
 var debuglog = require('debuglog')('bull');
+var Limiter = require('redis-rolling-rate-limiter');
 
 
 /**
@@ -114,6 +115,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   // Create queue client (used to add jobs, pause queues, etc);
   //
   this.client = createClient();
+  this.rateLimitClient = createClient();
 
   getRedisVersion(this.client).then(function(version){
     if(semver.lt(version, MINIMUM_REDIS_VERSION)){
@@ -337,18 +339,29 @@ Queue.prototype.close = function( doNotWaitJobs ){
 
   @method process
 */
-Queue.prototype.process = function(concurrency, handler){
+Queue.prototype.process = function(options, handler){
   var _this = this;
-  if(typeof concurrency === 'function'){
-    handler = concurrency;
-    concurrency = 1;
+  if(typeof options === 'function'){
+    handler = options;
+    options = {};
+  }
+
+  _.defaults(options, {
+    concurrency: 1,
+    rateLimit: undefined
+  });
+
+  // support deprecated passing of concurrency as first argument
+  if(typeof options === 'number') {
+    console.warn('[DEPRECATED] you should pass an object with { concurrency: Number } to queue#process');
+    options = { concurrency: options };
   }
 
   this.setHandler(handler);
 
   var runQueueWhenReady = function(){
     _this.bclient.once('ready', function(){
-      _this.run(concurrency).catch(function(err){
+      _this.run(options).catch(function(err){
         console.error(err);
       });
     });
@@ -359,7 +372,7 @@ Queue.prototype.process = function(concurrency, handler){
   this.bclient.on('error', runQueueWhenReady);
   this.bclient.on('end', runQueueWhenReady);
 
-  return this.run(concurrency).catch(function(err){
+  return this.run(options).catch(function(err){
     console.error(err);
     throw err;
   });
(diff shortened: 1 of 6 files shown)


leahciMic:
@manast I'm not quite sure if replacing concurrency with options object like I have is the best way to include this new option. I'm not sure what would be better either. Let me know what you think, I'm open to feedback and improvements.

lib/queue.js
```diff
@@ -114,6 +115,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   // Create queue client (used to add jobs, pause queues, etc);
   //
   this.client = createClient();
+  this.rateLimitClient = createClient();
```
manast: Could we reuse a redis client? people were already complaining that 3 redis clients per queue instance is too much (since many uses bull with redis as a service and pay by client).

lib/queue.js
```diff
@@ -337,18 +339,29 @@ Queue.prototype.close = function( doNotWaitJobs ){
 
   @method process
 */
-Queue.prototype.process = function(concurrency, handler){
+Queue.prototype.process = function(options, handler){
   var _this = this;
```
manast: The major problem with breaking the API is that it will require a new major release of bull (3.0.0), so it would be better if we could keep backwards compatibility. Maybe it could be possible to put it in the redisOpts in the constructor as a transition solution, and later move the redis opts inside a redis object (inside the more generic opts).

lib/queue.js
```diff
@@ -518,14 +531,40 @@ function pauseResumeGlobal(queue, pause){
...
+Queue.prototype.run = function(options){
   var promises = [];
   var _this = this;
+  var concurrency = options.concurrency;
+  var rateLimiter;
+
+  if (!options.rateLimit) {
+    rateLimiter = undefined;
+  } else {
+    var limiter = Limiter({
```
manast: in order to keep the run function small it would be great to refactor the rateLimiter outside of it.

lib/queue.js
```diff
@@ -603,14 +642,34 @@ Queue.prototype.startMoveUnlockedJobsToWait = function() {
...
-Queue.prototype.processJobs = function(resolve, reject){
-  var _this = this;
-  var processJobs = this.processJobs.bind(this, resolve, reject);
 
+Queue.prototype.processJobs = function(resolve, reject, options){
+  var _this = this;
+  var processJobs = this.processJobs.bind(this, resolve, reject, options);
   if(!this.closing){
+    var idleTime = 0;
+    var obeyRateLimit = function() {};
```
manast: same here as in run, refactor outside processJobs. These two methods are quite critical so better to keep them as small and clean as possible.

lib/queue.js
```diff
@@ -603,14 +642,34 @@ Queue.prototype.startMoveUnlockedJobsToWait = function() {
...
+
+    if (options.rateLimiter) {
+      obeyRateLimit = function() {
+        return options.rateLimiter()
+          .then(function(timeLeft) {
+            if (!timeLeft) { return; }
+            idleTime += timeLeft;
+            return Promise.delay(timeLeft)
+              .then(obeyRateLimit);
+          });
```
manast: I am a bit unsure, but I think this kind of asynchronous recursive structure will leak memory. We had to use a process.nextTick in order to "break" the promise chain and remove a leak we had in processJobs

manast:
Great PR. Please check my comments. There will be probably more people interested in commenting.

lib/queue.js
leahciMic: yeah sure, is it reasonable to reuse the queue client(`this.client`)?

lib/queue.js
leahciMic: No problem, will do

lib/queue.js
leahciMic: Not a problem, will do!

lib/queue.js
manast: I think so.

lib/queue.js
```diff
@@ -642,32 +642,19 @@ Queue.prototype.startMoveUnlockedJobsToWait = function() {
...
-    }
+    var startTime = new Date().getTime();
 
     process.nextTick(function(){
       (_this.paused || Promise.resolve())
-        .then(obeyRateLimit)
         .then(_this.getNextJob)
         .then(function(job) {
-          job.idleTime = idleTime;
+          if (job !== undefined) {
```
manast: another thing I am wondering, what does idleTime represent here? Because when startTime is defined, there may not even jobs in the queue, so it could happen that long time passes before a job arrives and gets processed, which will set idleTime to a large value although the job was processed immediately.

lib/queue.js
```diff
@@ -771,7 +758,13 @@ Queue.prototype.processJob = function(job){
 */
 Queue.prototype.getNextJob = function(opts){
   if(!this.closing){
-    return this.moveJob('wait', 'active', opts).then(this.getJobFromId);
+    var _this = this;
+
+    return ((this.rateLimiter && this.rateLimiter()) || Promise.resolve())
```
manast: we could also assign Promise.resolve to this.rateLimiter when no rateLimiter defined,  for a simpler code at this point.

manast:
@leahciMic it would be great if you could answer some of the questions in the review so that we can merge this quickly :)

lib/queue.js
leahciMic: Great suggestion, I will make this change

leahciMic:
@manast thanks for the feedback. I will incorporate it shortly.

There is one thing not mentioned here that I have discovered by using this branch in production. The library I'm using for rate limiting does not handle Redis reconnections gracefully like bull itself does. It also does not handle `close` very well.

I will find some time to address these issues this week.

lib/queue.js
manast: @leahciMic ok, so what you are talking about is a mechanism that reports how long time a job is waiting in the queue, and if that time is too long it can spin new worker instances. That is a  pretty cool concept actually. Used together with AWS for example, the queue could just grow or shrink as needed to guarantee a given wait target. I think thought that this should be handled in a separate issue, and it will also allow us to better comment on different ways to accomplish it.

manast: https://github.com/OptimalBits/bull/issues/414

leahciMic: @manast I agree that this should be handled as a separate issue and discussed separately. I will add more details about my use case to that issue.

sammarks:
@leahciMic had a chance to look at this yet? I'd love to use this feature in one of my projects.

manast:
I will revisit this PR soon. Rate limiting is a very important feature that we need to get right.

etiennea:
Any news on this?

manast:
It is planned for the 3.0.0 release

manast:
Finally I opened a new PR very much inspired by this one, but instead of using an external dependency I just implemented the rate limiter proposed in redis documentation. It is not a rolling limiter but I think for most cases it will work just fine and it was quite simple to implement:
https://github.com/OptimalBits/bull/pull/682

Issue #397, opened 2016-11-29
Error: Unable to renew nonexisting lock


igorissen:
Hi,

It is the 2nd time I'm having this issue in production after upgrading to the latest version `"bull": "^2.0.0"`. I'm still using node `v4.6.0`.

```
Error: Unable to renew nonexisting lock
1 at Object.scripts.takeLock (/home/ubuntu/project/node_modules/bull/lib/scripts.js line 314 col 13)
         throw new Error('Unable to renew nonexisting lock');
2 at Job.takeLock (/home/ubuntu/project/node_modules/bull/lib/job.js line 127 col 18)
         return scripts.takeLock(this.queue, this, renew, ensureActive)
3 at lockRenewer (/home/ubuntu/project/node_modules/bull/lib/queue.js line 644 col 16)
         return job.takeLock(renew, true).then(function(lock){
4 at null.<anonymous> (/home/ubuntu/project/node_modules/bull/lib/timer-manager.js line 80 col 5)
         fn();
5 at ontimeout (timers.js line 198 col 34)
6 at Timer.listOnTimeout (timers.js line 92 col 15)
```

cklmercer:
Having the same issue here. Haven't had any luck trying to figure it out.

emhagman:
Having a similar issue as well!

bradvogel:
ping @doublerebel

doublerebel:
@bradvogel on it.

@igorissen @cklmercer @emhagman Can you give me some more details about your setup?  That will help me pinpoint the best solution.

* Is Bull being used with NodeJS cluster?
* How many concurrent workers are processing jobs?
* Does the error crash Bull, or do you just see it in the logs?
* Do any of these errors coincide with failed jobs?

### Explanation:
In 2.0.0 we catch more errors where they were being dropped before, which is an improvement.  However, in certain situations Bull just logs the error to console (related: #58, #196), so now these caught errors show in logs but _should_ be harmless.

I most often see @cklmercer 's error:
```
Error renewing lock LockError: Exceeded 0 attempts to lock the resource "bull:aob-order-synchronizer:3555:lock".
```
This happens when multiple workers attempt to process the same job and one of them doesn't win the lock.  This can be safely ignored.  Let me know any details you can, and we'll take it from there.

cklmercer:
@doublerebel Thank you for your reply. It's reassuring to know the cause behind that particular error message. 

 Regarding the primary error, here's some of those details you asked for...

[...]

PR #398, opened 2016-12-02
Fixed listening for global events

floolean:
- Fixed listening for global events
- Temporary fix from https://github.com/OptimalBits/bull/issues/397
- Emit global event 'active'


bradvogel:
Remove that temporary fix. That's not safe to have in the source because it's just hiding a critical bug

manast:
global events works now, but you need to listen to 'global:eventname'.

PR #399, opened 2016-12-03
[fix test] don't renew a lock that is already released [Fixes #397]

### LINKED ISSUE #397: Error: Unable to renew nonexisting lock

doublerebel:
https://github.com/OptimalBits/bull/issues/397

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -656,6 +656,11 @@ Queue.prototype.processJob = function(job){
 
   var timeoutMs = job.opts.timeout;
 
+  function clearTimer(){
+    _this.timers.clear(lockRenewId);
+    return Promise.resolve();
+  }
+
   function handleCompleted(data){
     try{
       JSON.stringify(data);
@@ -670,6 +675,7 @@ Queue.prototype.processJob = function(job){
       return;
     }
 
+    clearTimer();
     return job.moveToCompleted(data)
       .then(function(){
         return _this.distEmit('completed', job.toJSON(), data);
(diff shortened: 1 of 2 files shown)


lib/queue.js
```diff
@@ -680,6 +686,7 @@ Queue.prototype.processJob = function(job){
     _this.processing--;
     var error = err.cause || err; //Handle explicit rejection
     return job.moveToFailed(err)
+      .then(clearTimer)
```
manast: don't we have an hazard here where the timer is cleaned *after* we moved to failed?

bradvogel: How so? The worst case is that it gets renewed after moving to failed, but then immediately released.

lib/queue.js
```diff
@@ -670,6 +675,7 @@ Queue.prototype.processJob = function(job){
       return;
     }
 
+    clearTimer();
```
bradvogel: I think it'd be slightly more readable to make this a part of the promise chain, eg

```
return clearTimer().
    then(function() {
       return job.moveToCompleted(data);
    })
```

PR #415, opened 2016-12-16
#371 fix, check for lock before marking complete/failed

### LINKED ISSUE #371: "Unexpected token u in JSON at position 0" while processing job
I'm noticing that some of our jobs are "corrupted" and the queue is throwing the following error when trying to process them:

```
Error processing job: SyntaxError: Unexpected token u in JSON at position 0
    at Object.parse (native)
    at Function.Job.fromData (/Users/brad/dev/bull-queue/node_modules/bull/lib/job.js:418:33)
    at /Users/brad/dev/bull-queue/node_modules/bull/lib/job.js:68:18
    at tryCatcher (/Users/brad/dev/bull-queue/node_modules/bluebird/js/release/util.js:16:23)
```

In inspecting the job in redis using `hgetall bull:test:1` I can see that only `attemptsMade`, `stacktrace`, and `failedReason` keys are present in the job data.

This leads me to believe that a job failed (and `_saveAttempt` was called) _after_ the job data had been already removed - which can only happen if the job was completed by another worker. And the only way another worker could have completed it is if the first worker failed to renew the lock.

To fix this, what do you think about making `Job.prototype.moveToFailed` an atomic script that guarantees that we own the lock before moving the job. So if another worker picked up the job and it got double processed, we won't do farther damage by trying to move that completed job back to the delayed queue.

This is related to #273.

rissem:
Fixes #371 

I was able to reproduce this error with this gist https://gist.github.com/rissem/8565caf2290973c616c1de5376088c61
This essentially follows the reproduction case @bradvogel outlined in #371 

Please let me know if this approach looks right to you. Happy to make changes to get this pull request accepted. Thanks!

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -662,7 +662,7 @@ Queue.prototype.processJob = function(job){
   }
 
   function handleCompleted(data){
-    job.takeLock(true, false).then(function(lock) {
+    job.takeLock(true /* renwew */, false /* ensureActive */).then(function(lock) {
       // This substraction is duplicate in handleCompleted and handleFailed because it have to be made before throwing any
       // event completed or failed in order to allow pause() to work correctly without getting stuck.
       _this.processing--;
@@ -688,7 +688,7 @@ Queue.prototype.processJob = function(job){
   }
 
   function handleFailed(err){
-    job.takeLock(true, false).then( function(lock) {
+    job.takeLock(true /* renwew */, false /* ensureActive */).then( function(lock) {
       return clearTimer()
         .then(job.moveToFailed(err))
         .then(job.releaseLock.bind(job))
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -662,35 +662,46 @@ Queue.prototype.processJob = function(job){
   }
 
   function handleCompleted(data){
+    job.takeLock(true, false).then(function(lock) {
+      // This substraction is duplicate in handleCompleted and handleFailed because it have to be made before throwing any
+      // event completed or failed in order to allow pause() to work correctly without getting stuck.
+      _this.processing--;
+
+      if(_this.closed){
+        return;
+      }
+
+      clearTimer();
+      return job.moveToCompleted(data)
+        .then(function(){
+          return _this.distEmit('completed', job.toJSON(), data);
+        });
+    }, function(err){
+      console.error('failed ot obtain lock before marking completed, bailing', err)
+    } )
     try{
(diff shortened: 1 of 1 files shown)


lib/queue.js
```diff
@@ -662,35 +662,46 @@ Queue.prototype.processJob = function(job){
   }
 
   function handleCompleted(data){
+    job.takeLock(true, false).then(function(lock) {
```
bradvogel: Small nitpick: comment boolean params: eg `job.takeLock(true /* renew */, false /* don't ensure active */)`

bradvogel:
This looks good to me. @manast and @doublerebel does this make sense to you too? It seems safe and low overhead to try to renew the lock before moving the job. It still leaves open the very small chance that the lock will have expired (in a stalled event loop condition) between the time it's renewed and when the job is actually moved. The only way to fix that would be to ensure the lock (atomically) when moving it to completed or failed. But it'd be nice to get away from checking lock existence in Lua scripts, and instead use the redlock library like we currently do. Thoughts?

bradvogel:
ping @manast @doublerebel . Does this look ok to merge?

manast:
If we merge this, it should be a temporary fix until moveToFailed is atomic, right? the code for moveToComplete is already taking a lock inside the move function, so it seems redundant to take it once more as in this patch:
https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L196

bradvogel:
You're right - the code in this PR to take the lock in moveToComplete can be removed since it's redundant with the code you linked to.

What were you thinking for making `moveToFailed` atomic? Having it use the new atomic `move` function isn't enough because we need to check the lock _prior_ to calling `saveAttempt` here: https://github.com/OptimalBits/bull/blob/master/lib/job.js#L179. Otherwise it'll write job data even though it didn't have the lock, and run into https://github.com/OptimalBits/bull/issues/371.

So that leaves two solutions to make `moveToFailed` atomic:
1) Make `_saveAttempt` a lua script that internally checks for the lock prior to writing the job data
2) Use the code in this PR that calls `takeLock` before calling `saveAttempt`

Since we're moving away from checking the lock in scripts (otherwise it defeats the purpose using redlock), I think #2 is the best option.

What do you think? I'll be happy to make a new PR with the changes.

bradvogel:
ping @manast on ^^ :)

bradvogel:
ping @manast on ^^ :)

PR #418, opened 2016-12-19
fixes #363 method to permanently fail a job

### LINKED ISSUE #363: method to permanently fail a job
Feature request: a method on the job, such as `job.fail()` that can be called inside the job handler to move a job to failed (bypassing any more job attempts).

Use case:
We're creating jobs using the `attempts` property, like this:

```
queue.add({}, {
    attempts: 10
});
```

So jobs are automatically retried if an error is returned from the handler. However, there are circumstances in which the job handler might encounter a permanent error (such as something being removed the database that's needed by the handler), and it would be nice if it had the ability to stop the job from being retried.

So ideally it could call something like `job.fail()` that would just fail the job:

```
queue.process(function (job, jobDone) {
  if (permanentError){
     job.fail(); // Will move the job to 'failed' after it's finished
  }
 jobDone();
});
```

How does this sound? I can file a PR.

rissem:
add markUnrecoverable method to Job

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -174,15 +174,15 @@ Job.prototype.move = function(src, target, returnValue){
   return scripts.move(this || 0, src, target);
 }
 
-Job.prototype.markUnrecoverable = function(){
-  this._unrecoverable = true;
+Job.prototype.discard = function(){
+  this._discarded = true;
 }
 
 Job.prototype.moveToFailed = function(err){
   var _this = this;
   return this._saveAttempt(err).then(function() {
     // Check if an automatic retry should be performed
-    if(_this.attemptsMade < _this.attempts && !_this._unrecoverable){
+    if(_this.attemptsMade < _this.attempts && !_this._discarded){
       // Check if backoff is needed
       var backoff = _this._getBackOff();
       if(backoff){
--- a/lib/job.js
+++ b/lib/job.js
@@ -174,11 +174,15 @@ Job.prototype.move = function(src, target, returnValue){
   return scripts.move(this || 0, src, target);
 }
 
+Job.prototype.markUnrecoverable = function(){
+  this._unrecoverable = true;
+}
+
 Job.prototype.moveToFailed = function(err){
   var _this = this;
   return this._saveAttempt(err).then(function() {
     // Check if an automatic retry should be performed
-    if(_this.attemptsMade < _this.attempts){
+    if(_this.attemptsMade < _this.attempts && !_this._unrecoverable){
       // Check if backoff is needed
       var backoff = _this._getBackOff();
       if(backoff){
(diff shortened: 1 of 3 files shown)


lib/job.js
```diff
@@ -174,11 +174,15 @@ Job.prototype.move = function(src, target, returnValue){
   return scripts.move(this || 0, src, target);
 }
 
+Job.prototype.markUnrecoverable = function(){
```
bradvogel: Please document in readme

test/test_queue.js
```diff
@@ -1288,6 +1288,27 @@ describe('Queue', function () {
...
+          job.markUnrecoverable();
+          jobDone(new Error('unrecoverable error'))
+        });
+
+        queue.add({ foo: 'bar' }, {
+          attempts: 5
+        });
+      });
+      queue.on('failed', function () {
+        //console.log('failed?', arguments);
```
bradvogel: remove commented out code

bradvogel: shouldn't the `expect(tries).to.equal(1)` go in here?

test/test_queue.js
```diff
@@ -1288,6 +1288,27 @@ describe('Queue', function () {
...
 
+    it('should not retry a job if it has been marked as unrecoverable', function (done) {
+      var tries = 0;
+      queue = utils.buildQueue('test retries and backoffs');
+      queue.on('ready', function () {
+        queue.process(function (job, jobDone) {
+          tries++;
+          expect(tries).to.equal(1)
+          job.markUnrecoverable();
+          jobDone(new Error('unrecoverable error'))
```
bradvogel: nit: use semi-colons consistently

test/test_queue.js
rissem: I intended it to catch a second attempt at the job (tries = 2 but should be 1)

manast:
If a job is marked as unrecoverable, and then the process working on it dies, the mark is not persisted. Is this something that we should care about?

rissem:
I originally wrote it as `.fail()` but it seemed unnatural to me. Jobs already fail when an exception is thrown or when an error is returned via the callback. What's different here is not that it's failing but that it won't be retried. I think w/ a `fail()` method, a new user may try to use it instead of returning an error / throwing an exception, and be surprised when the job wasn't retried. Just my two cents though. Happy to change to `.fail()` if you think that makes more sense.

bradvogel:
@manast could this be merged?

manast:
the only thing I think should be fixed is the name of the method: ```markUnrecoverable``` is too techie, we should have a high level name for this function. If you think ```fail``` can lead to confusion then it is maybe better something like ```failPermanently``` or ```discard```,  se should something short and nice, open for suggestions.

rissem:
I like `discard`, will update pull request with that.

Issue #401, opened 2016-12-04
mixing getJobCountByTypes including paused yields 0


jamesjjk:
We are in the process of replacing Kue with Bull and so far so good. However we noticed whilst implementing repeatable jobs that when we yield getJobCountsByType and include 'paused' the result yields 0 whilst there are delayed and active jobs present.

```
const count = yield _queue.getJobCountByTypes('paused', 'delayed', 'wait', 'active');
//RETURNS 0 (expected positive int)
```

```
const count = yield _queue.getJobCountByTypes('delayed', 'wait', 'active');
//RETURNS positive int
```

manast:
hmm, when you are getting the count for paused jobs, is the queue actually paused?
Because when the queue is paused, it renames the wait queue so that the redis client that is processing jobs will pause the processing (globally), until the queue is resumed. So what I suspect it that getJobCountByType is not taking this fact into consideration, and returning 0 for wait when the queue is paused-

jamesjjk:
@manast I didn't have the queue actually paused, and there were delayed, waiting, and active items available. However the count always returned 0 once I included paused literal.

manast:
ok, I will write a unit test to see if I can reproduce it.

GoNode5:
```javascript
QueueSend.getJobCountByTypes(['paused', 'delayed','failed']).then(function(result) {
  console.log('paused/delayed /failed jobcount:',result)	
})

QueueSend.getJobCountByTypes(['failed']).then(function(result) {
  console.log('jobcount:',result)	
})

QueueSend.getFailedCount().then(function (count) {
	console.log('failed count',count)
})
```
results in:
failed count 28
paused/delayed /failed jobcount: 0
jobcount: 28

PR #420, opened 2016-12-20
correct getJobCountByTypes

### LINKED ISSUE #419: getJobCountByTypes returns only the number of the first type
https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L794

GoNode5:
closes #419 #401 my first PR ...


GoNode5:
I never done any PR so pls bear with me :-) thnx

manast:
I agree, we should have a test case that reproduces the error and that this PR fixes.

PR #428, opened 2017-01-09
replace manual caching with ioredis.defineCommand

sloops77:
Really is very simple. 
This has two benefits: 
1. uses ioredis's internal caching - less code, less for bull to do
2. its makes bull compatible with newrelic's transaction monitoring (tbh im not sure what was incompatible - im just happy it's working)

what say you?


manast:
Yes it is simpler. Regarding 2, I do not think it makes any difference.

PR #434, opened 2017-01-17
Fix Lock Redis script to not be a different script for every Job ID #426

### LINKED ISSUE #426: redis script cache gets bloated after update to bull 2.0
Hello,

I'm experiencing a growing Lua script cache after updating to bull-2.0
After 20 days of uptime my redis Lua cache grown up to 15GB

Anyone experiencing same issue?

Thanks

joeandrews:
NOTES: this will still violate the redis cluster requirements that each script should explicitly define the keys it writes to in the KEYS array, however the current implementation also violates this. This pull request should fix the growing lua_memory_cache issues. #426

### DIFF
--- a/lib/scripts.js
+++ b/lib/scripts.js
@@ -24,20 +24,26 @@ function execScript(client, hash, lua, numberOfKeys){
 
 var scripts = {
   _isJobInList: function(keyVar, argVar, operator) {
-    keyVar = keyVar || 'KEYS[1]';
-    argVar = argVar || 'ARGV[1]';
-    operator = operator || 'return';
-    return [
-      'local function item_in_list (list, item)',
-      '  for _, v in pairs(list) do',
-      '    if v == item then',
-      '      return 1',
-      '    end',
-      '  end',
-      '  return nil',
-      'end',
-      ['local items = redis.call("LRANGE",', keyVar, ' , 0, -1)'].join(''),
-      [operator, ' item_in_list(items, ', argVar, ')'].join('')
+   	keyVar = keyVar ? 'splitKey[1]..":"..splitKey[2]..":active"' : 'KEYS[1]';
+		argVar = argVar || 'ARGV[1]';
+		operator = operator || 'return';
+		return [
+			'local function item_in_list (list, item)',
+			'  for _, v in pairs(list) do',
+			'    if v == item then',
+			'      return 1',
+			'    end',
+			'  end',
+			'  return nil',
+			'end',
+			'local splitKey={}',
+			'local i=1',
+			'for str in string.gmatch(KEYS[1], "([^:]+)") do',
+					'splitKey[i] = str',
+					'i = i + 1',
+			'end',
+			[ 'local items = redis.call("LRANGE", ', keyVar, ' , 0, -1)' ].join(''),
+			[ operator, ' item_in_list(items, ', argVar, ')' ].join('')
     ].join('\n');
   },
   isJobInList: function(client, listKey, jobId){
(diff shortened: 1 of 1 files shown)


manast:
thanks for the fix. It will take some time for me to review the code so it may take a few days before I can merge it.

albertjan:
For everyone's sanity this view of the diff is a lot easier to digest: https://github.com/OptimalBits/bull/pull/434/files?w=1

manast:
@joeandrews it would be great if you could check the review feedback :)

joeandrews:
@manast, thanks for the review. My understanding of Redis scripts are that each script should be a pure function and as such should be passed in any data required to run. The script cache simply caches the function and the idea is that you call that function with different KEYS and ARGV arrays. If the keyVar and argVar is causing a script to be different or unique because of a hard coded key, redis will cache that in the lua cache. My change simply changes the different script for every job to two a maximum of 4 different scripts depending on the keyVar and argVar requirements. If there are other points in the code that do this they will also have to be changed. If I have a moment I can dive in further, but I would have to get up to speed on how Bull is using scripts and the purpose of each script.

manast:
What you say is correct. I am not sure how the old implementation worked all actually, but if you check the execScript implementation you will see that we use a hash (a name) to define a new command in ioredis (this uses internally EVAL SHA). Could be that ```definecommand``` is overwriting the old "command", in such case it is also wrong since we are not reusing the scripts properly and instead sending them everytime to the redis server instead of just sending a small hash value: https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L15-L23

gblock0:
@manast could the `client` be getting reinitialized and forgetting the commands that have been defined? When l added in  `execScript` before `client.defineCommand`: 
```
if (!client || !client[hash]) {
	console.log('hash', hash);
	console.log('numKeys', numberOfKeys);
}
```
`moveactive`, `addJobLPUSH`, `addJob:delayed`, `updateDelaySet`, and `moveUnlockedJobsToWait` would get printed very often.

prdn:
This issue is really becoming important to solve to allow a proper use of bull-2 in any production environment. I'll try to have a look and help in next days @manast @joeandrews

Thanks for trying to fix this

holm:
Our alarms just went off due to this. Nothing really in Redis in terms of keys, but very high memory usage. Any fix for this would be greatly appreciated.

manast:
I could merge this, but as you see in the comments there is a small issue with it, that potentially can cause problems. I hope to have some time over to implement a better solution.

manast:
ok. I reviewed this once more time. There does not seem to be any issues at all with it, I was wrong before. So ready to merge.

tinybug:
This PR will cause job stuck in active but not running actually. When I checkout to #456 , it run normal. My job has priority option. @manast

manast:
@tinybug could you share a small code snippet that I can use to reproduce it?

tinybug:
@manast  Thanks reply! The bull's test code work well, I should dig into my code to find out why I got this happen.

tinybug:
@manast I find out how to reproduce this bug, if the queue name contain colon, like 'test:queue', this problem will happen, please have a check.

manast:
Ok, so I realised this fix is still generating a new script per jobId, this is due to using 'argVar' inside the script, which is different for every jobId.

manast:
@tinybug you are right. This is a new bug introduced in this PR-

holm:
First of, @manast, thanks for all the hard work. 

Since we suffer from this issue in production, I decided to take a look at the issue also. As you also found, it's the use of `argVar` that now causes unique scripts for each job. Fixing it the right way seems complicated, since we would ideally have to pass in the jobId through Redlock, but there is no obvious way to do that. I guess it could be extended, but it all seems to stretch the library to something that it probably shouldn't support.

I think that leaves the option of vendoring Redlock or applying a "better hack". Continuing on the idea of splitting the lock key to get the `jobId`, I think you could resolve it by splitting the key a bit differently. Since they key will be `part1:part2:...:partn:jobId:lock`, you could split it to extract the parts as one and the `jobId` as another. You can then reassemble as needed to get just the jobId or the name of the active key.

manast:
@holm I am actually patching redlock so that we can pass custom keys and args to the lock script. I am also optimizing the call using ioredis defineCommand, which should save a bit of bandwidth. I do not expect to get this merged into redlock repo as is, but maybe we can figure out a way that is acceptable for its maintainer in the future.

manast:
I have pushed my fixes. I could reproduce the memory leaks and with the new code they are not there anymore. I will need to verify all the tests (some where failing locally) and after that we could do a new release.

manast:
I did one more fix in the redlock fork and now almost all tests passes. The modifications in redlock are not so pretty, so I will need to change that to something more generic that could have a chance to be accepted by redlocks maintainer.

PR #448, opened 2017-02-16
Some fixes regarding pausing locally

### LINKED ISSUE #446: locally pausing a queue does not work when there are multiple active workers
```javascript
queue.pause(true /* local */);
```

will never resolve if there is more than one simultaneously active node process working on jobs.

This means that if we try to do
```javascript
queue.pause(true /* local */).then(function() {
  queue.close();
});
```
our queue will stop processing jobs, but will never close.

This is happening because:

Suppose queueA and queueB are two separate node processes working on the same queue. Suppose they are each processing one job.

If we call `queueA.pause(true)`, queueA will [count how many jobs](https://github.com/OptimalBits/bull/blob/v2.1.2/lib/queue.js#L952) are active [globally](https://github.com/OptimalBits/bull/blob/v2.1.2/lib/queue.js#L824), getting 2.

queueA will then [wait for 2 (completed|failed|stalled) events](https://github.com/OptimalBits/bull/blob/v2.1.2/lib/queue.js#L964-L966) before pausing.

Since queueA is only working on one job, and queueA [won't start any more jobs](https://github.com/OptimalBits/bull/blob/v2.1.2/lib/queue.js#L612) because it is trying to pause, there will only ever be 1 (completed|failed|stalled) event, so queueA will never pause.

I don't think that listening for a (completed|failed|stalled) event globally is a good solution here. Suppose queueB processes two jobs before queueA can finish it's one job. Then queueA would receive two events, even though queueA is still working on it's job, and would pause incorrectly.

### LINKED ISSUE #447: Jobs can sometimes be processed after a queue is paused locally.
Jobs can sometimes be processed after a queue is paused locally.

This happens because we wait up to five seconds for a job to become available when we try to get the next job.

This is the sequence of events when this happens:

The queue [does](https://github.com/OptimalBits/bull/blob/v2.1.2/lib/queue.js#L754) `queue.bclient.brpoplpush` to get the next job, but none are available yet.

We locally pause the queue; `queue.bclient.brpoplpush` is still pending. Since some jobs are still actively processing, we wait for them to finish before resolving the pause.

A new job is pushed to the queue.

`queue.bclient.blpoprpush` returns with the new job, and our queue starts processing it.

The original set of active jobs all complete and `queue.pause(true)` resolves. The new job is still processing.

So we think our queue is paused, but we're actually still processing a job.

If we kill the queue, the active job may be terminated before it can complete.

josephwarrick:
Fixes #446 
Fixes #447

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -315,6 +315,7 @@ Queue.prototype.close = function( doNotWaitJobs ){
     return this.closing;
   }
 
+
   return this.closing = this._initializing.then(function(){
     clearTimeout(_this.delayTimer);
     clearInterval(_this.guardianTimer);
@@ -949,25 +950,23 @@ Queue.prototype.whenCurrentJobsFinished = function(){
   var _this = this;
   var resolver;
   return new Promise(function(resolve, reject) {
-    _this.getActiveCount().then(function(count) {
-      if(count === 0){
+    if(_this.processing === 0){
+      resolve();
+    }else{
+      resolver = _.after(_this.processing, function(){
+        _this.removeListener('stalled', resolver);
+        _this.removeListener('completed', resolver);
+        _this.removeListener('failed', resolver);
+        clearInterval(_this.moveUnlockedJobsToWaitInterval);
         resolve();
-      }else{
-        resolver = _.after(count, function(){
-          _this.removeListener('stalled', resolver);
-          _this.removeListener('completed', resolver);
-          _this.removeListener('failed', resolver);
-          clearInterval(_this.moveUnlockedJobsToWaitInterval);
-          resolve();
-        });
+      });
 
-        _this.on('stalled', resolver);
-        _this.on('completed', resolver);
-        _this.on('failed', resolver);
+      _this.on('stalled', resolver);
+      _this.on('completed', resolver);
+      _this.on('failed', resolver);
 
-        _this.startMoveUnlockedJobsToWait();
-      }
-    }, reject);
+      _this.startMoveUnlockedJobsToWait();
+    }
   });
 };
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -979,7 +979,7 @@ Queue.prototype.whenCurrentJobsFinished = function(){
   var _this = this;
   var resolver;
   var count = this.processing + this.retrieving;
-  return new Promise(function(resolve, reject) {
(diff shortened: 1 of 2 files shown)


lib/queue.js
```diff
@@ -141,6 +141,13 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   // Create blocking client (used to wait for jobs)
   //
   this.bclient = createClient();
+  Object.defineProperty(this, 'jobWaitCount', { get: function() {
+    return this.bclient.commandQueue.length;
+  } });
+  Object.defineProperty(this, 'whenCurrentWaitFinished', { get: function() {
+    var currentMove = this.bclient.commandQueue.peekFront()
+    return currentMove && currentMove.command.promise || Promise.resolve();
+  } });
```
manast: why are these methods defined like this instead of in Queue.prototype?

manast:
vyr good job, please check my comments.

test/test_queue.js
```diff
@@ -977,6 +977,48 @@ describe('Queue', function () {
         }).catch(done);
       });
     });
+
+    it('should pause the queue locally when more than one worker is active', function () {
+      var queue1 = utils.buildQueue('pause-queue');
+      var queue1IsProcessing = new Promise(function(resolve, reject) { //eslint-disable-line no-unused-vars
```
josephwarrick: Just habit. `new Promise(function(resolve){})` always looks odd to me. I'll remove them.

lib/queue.js
josephwarrick: Mostly so that they can be dynamic properties, rather than functions:
```javascript
console.log('we are waiting for', queue.jobWaitCount, 'jobs')
// we are waiting for 4 jobs
```
rather than
```javascript
console.log('we are waiting for', queue.getJobWaitCount(), 'jobs');
// we are waiting for 4 jobs
```

I also like that they're defined close to `queue.bclient` since they're basically just attributes of `queue.bclient` that I'm giving easier names to.

Would you prefer that I change them to methods on `Queue.prototype`?

manast: I think it is better to put them on the prototype, mostly so that we can follow the same standard as in the rest of the code. It becomes easier to contribute for others.

josephwarrick: Fair enough.

I've moved them to methods on `Queue.prototype`.

Issue #244, opened 2016-02-11
Can't set delay to greater than 24 days


laoshanlung:
It seems that when you set the delay to more than 24 days, the job gets executed immediately. 

I take a look at the test https://github.com/OptimalBits/bull/blob/master/test/test_queue.js#L869, and change delay to `24*30*3600*1000` (30 days), the test still passes immediately which is weird

Btw, the expect statement seems to be wrong (I am not sure, but I think it should be `expect().to.be(true)`)

manast:
@laoshanlung will need to take a look into it.

PR #452, opened 2017-02-17
Fix for max 24 day delay issue #244

mwielbut:
The implementation of setTimeout in Nodejs relies on a 32 bit integer for timeout values. The current default timeout in the delay timer is the max 64 bit integer which causes any possible delay values to start the delay timer. Delays longer than 2^31-1 ms (roughly 24 days) cause an integer overflow and the timeout triggers immediately.

This pull request replaces the default timeout with the max 32bit signed integer so that only valid ms values can start the timer. Anything longer than 24 days will be caught by the guardian timer.

To reproduce this issue clear anything in the delay set and create a new job with a delay > 2^31 ms. It will execute immediately.

https://nodejs.org/docs/latest/api/timers.html#timers_settimeout_callback_delay_args

#244

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -230,7 +230,7 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   //
   // Init delay timestamp.
   //
-  this.delayedTimestamp = MAX_TIMEOUT_MS;
+  this.delayedTimestamp = Number.MAX_VALUE;
   scripts.updateDelaySet(this, Date.now()).then(function(timestamp){
     if(timestamp){
       _this.updateDelayTimer(timestamp);
@@ -557,7 +557,7 @@ Queue.prototype.run = function(concurrency){
 Queue.prototype.updateDelayTimer = function(newDelayedTimestamp){
   var _this = this;
 
-  if(newDelayedTimestamp < _this.delayedTimestamp){
+  if(newDelayedTimestamp < _this.delayedTimestamp && newDelayedTimestamp < (MAX_TIMEOUT_MS + Date.now())){
     clearTimeout(this.delayTimer);
     this.delayedTimestamp = newDelayedTimestamp;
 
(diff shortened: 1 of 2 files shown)


manast:
thanks for the fix. It seems reasonable, but the delay related unit tests are failing, any idea why?

mwielbut:
Simplified the fix a bit after I realized I was setting the delay and not the timestamp value but one of the tests is still failing intermittently.  If I run the test runner with a max of 7000ms per test it runs ok. There seems to be an issue with the queue.close() completing in time. Not sure.

mwielbut:
It looks like the one failing test fails even if I roll back the changes in the PR. It's a timing issue with the way the `close` function works. If `getNextJob` happens to kick off before the `closing` variable is set, it will attempt a `brpoplpush` waiting for the next retrieval and will need to wait the 5 seconds in `LOCK_RENEW_TIME`. This will cause the test to fail as it will exceed the 5000ms max. I guess either the test parameters need to be adjusted to allow for 6000ms test timeout or the test needs to be changed to pass `close(true)` to set `doNotWaitActive` to `true` and bypass the wait.

manast:
If you fix that I will merge at once :)

mwielbut:
All set. Thanks!

oneEyedSunday:
So @mwielbut is it right to assume that one can set longer delays now? Delays like 5weeks, 3 months, 1 year, etc?

manast:
ok, so I am going to revisit this issue more carefully, because it does not make sense to limit to max 24 days. This is just a "checker" for delayed jobs, of course you should be able to delay jobs more than 24 days, but in that case we just will poll the delayed set with an interval of 24 days.

oneEyedSunday:
Interesting @manast cos in my current application I need to repeat jobs in the order of weeks and months. Would like to see how this develops. Is there an issue or task I can watch concerning this to get updates?

manast:
This will be high prio for me, but I am very busy working for bull 4.x atm.

oneEyedSunday:
Hello @manast has this been fixed in v4 beta? Just circled back to this today, im still on v3

PR #454, opened 2017-02-23
Updated old redis.createClient constructor calls to the ioredis way

brantstuns:
since createClient is deprecated in ioredis


brantstuns:
Noticed there was some node_redis remnants still hanging out so I cleaned them up. It's mentioned in the migration notes in ioredis that createClient() instantiation calls are less preferable than the new redis() syntax. Also updated the api docs in the readme to point to the ioredis options object argument in the client creation as it's a little different than node_redis

PR #456, opened 2017-02-27
implementation of improved priority queues

manast:
This is a new implementation for priority queues. Not only is this implementation much more simpler than the old one, it is also far more stable and performant. It works without polling and does not require extra connections to redis.

### DIFF
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -318,6 +318,49 @@ describe('Queue', function () {
       });
     });
 
+    it('should processes jobs by priority', function(done){
+      var normalPriority = [],
+        mediumPriority = [],
+        highPriority = [];
+
+      // for the current strategy this number should not exceed 8 (2^2*2)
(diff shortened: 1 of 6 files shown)


manast:
@bradvogel You may be interested in review this. The overhead for standard, non prioritized jobs is minimal, just an extra call to ZREM per job is necessary.

test/test_queue.js
```diff
@@ -318,6 +318,49 @@ describe('Queue', function () {
       });
     });
 
+    it('should processes jobs by priority', function(done){
+      var normalPriority = [],
+        mediumPriority = [],
+        highPriority = [];
+
+      // for the current strategy this number should not exceed 8 (2^2*2)
+      // this is done to maitain a deterministic output.
```
manast: This comment does not hold anymore since we do not support any strategies (yet).

bradvogel:
this lgtm!

Issue #322, opened 2016-07-20
Possible deadlock on Failed jobs


chuym:
I'm noticing that several failed jobs remain locked.
Monitoring the TTL of the lock, I notice that it renews exactly at `LOCK_RENEW_TIME / 2`

The only place where such lock happens is in: https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L570

By inspecting the logic, I think there is a race condition that is causing a deadlock to take place:
1. A job starts being processed and a lock renew loop kicks off
2. A job fails and gets moved to `failed`
3. Just there, the lock renew function calls itself again and attempts to renew the lock
4. Before step 3 is completed, the timer is cleared by the `finally` clause in https://github.com/OptimalBits/bull/blob/27321806a95d07439412ffc666088cdd9ebadaa4/lib/queue.js#L627
5. The lock renew function completes and sets the timer again in https://github.com/OptimalBits/bull/blob/27321806a95d07439412ffc666088cdd9ebadaa4/lib/queue.js#L570

manast:
Could this be fixed in 1.1.2 after improvements in stalled jobs or otherwise in master after @doublerebel improvements?

manast:
As far as I can see the race condition is still there. I will try to find a solution for this.

manast:
Since we are checking if the timer is stopped, this hazard should not exist anymore:
https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L684
I will re-open if not the case.

PR #472, opened 2017-03-28
Defining process names

manast:
This PR implements a mechanism for reusing the same queue to different job types. It allows to specify several named process functions, one for every type. It can be compared to having a large process function with a switch case, but the implemented solution allows for cleaner and more modular uses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ __Arguments__
 #### Queue##Process
 
 ```ts
-process(concurrency?: number, processor: (job, done?) => Promise<any>)
+process(name?: string, concurrency?: number, processor: (job, done?) => Promise<any>)
 ```
 
 Defines a processing function for the jobs placed into a given Queue.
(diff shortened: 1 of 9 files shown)


manast:
@bradvogel you are welcome to take a look to this PR. Despite the simple feature of being able to define multiple process functions per queue, I also made a lot of cleanups and small fixes. Also the unit tests are now running much more quiet than before.

manast:
As far as I can see, this PR also fixes https://github.com/OptimalBits/bull/issues/322

NikosEfthias:
@manast are the multiple queue process functions called concurrently or one after the other in order ?

manast:
they are called in order, but if you define a concurrency argument or if you have several workers, some of them will be processed in parallel...

NikosEfthias:
well in my case I have only one worker and multiple requests so the entire point is to serialize the action and queue requests. Seems this will do the trick. I'm npm installing the new version then and ill give you a brief review

NikosEfthias:
Ah I see the pr has not been merged yet

PR #478, opened 2017-03-29
fix broken example in readme

mattkrick:


manast:
thanks for the patch, appreciated.

Issue #190, opened 2015-10-02
`completed` and `failed` should be zsets, not sets


evanhuang8:
What's the reasoning for using sets instead of lists for `completed` and `failed` jobs? By using sets for the two categories, it will not be safe or optimal to fetch jobs from the two, especially for management purposes. As of now, `getCompleted` and `getFailed` will result in the whole set being fetched, which can potentially result in slow performance and even crash the queue if the sets are too big. 

By using `clean`, I suppose the two sets can be kept in reasonable size, but there are use cases where jobs should be kept in the queue indefinitely, for records or other purposes. As far as I know, we are not doing any set operations on the two, so why don't we just use lists instead?

manast:
The rationale was that it was much faster to check if a job had failed or completed by having them on sets. I do not remember anymore why I did it like this. This is an old (and probably wrong) design decision that maybe does not hold anymore. We can just save the state on the job itself and that would be even faster.

evanhuang8:
I see, I think it's still necessary to keep them in a set then. Thinking more about searching/monitoring, it might be good to maintain a list along side with the set?

manast:
I think the best solution is to keep the elements in a ZSET instead with the jobId or the timestamp as score. Then we get the best of two worlds. We can check in O(1) if a job is in a given set, and we can also return the elements in ranges to support pagination, etc.

PR #489, opened 2017-04-06
Implemented #190

manast:
This PR changes the completed and failed sets to be zsets. This allows us to efficiently get a subset of the completed or failed jobs, which is very useful for pagination on UIs, etc.
Along with this PR there are also some refactoring, for instance I started to move the LUA scripts to separate files, which is much more cleaner and convenient than what we have today.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -726,7 +726,6 @@ Queue.prototype.processJob = function(job){
 
     stopTimer();
 
-    // TODO: Should moveToFailed ensure the lock atomically in one of its Lua scripts?
     // See https://github.com/OptimalBits/bull/pull/415#issuecomment-269744735
     job.takeLock(true /* renew */, false /* ensureActive */).then( function(/*lock*/) {
       return job.moveToFailed(err).finally(function(){
@@ -858,8 +857,6 @@ Queue.prototype.getJobCountByTypes = function() {
     switch(type) {
       case 'completed':
       case 'failed':
-        multi.scard(key);
-        break;
       case 'delayed':
         multi.zcard(key);
         break;
@@ -890,8 +887,8 @@ Queue.prototype.getJobCounts = function(){
   return this.client.multi()
     .llen(this.toKey('wait'))
     .llen(this.toKey('active'))
-    .scard(this.toKey('completed'))
-    .scard(this.toKey('failed'))
+    .zcard(this.toKey('completed'))
+    .zcard(this.toKey('failed'))
     .zcard(this.toKey('delayed'))
     .exec().then(function(result){
       result.forEach(function(res, index){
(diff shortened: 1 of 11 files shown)


manast:
@bradvogel ping :)

bradvogel:
Awesome! Will review shortly

lib/scripts/updateDelaySet.lua
```diff
@@ -0,0 +1,35 @@
+--[[
```
bradvogel: thanks for moving these into separate files - much cleaner!

lib/queue.js
```diff
@@ -961,17 +958,6 @@ Queue.prototype.getJobs = function(queueType, type, start, end){
     case 'LIST':
       jobs = this.client.lrange(key, start, end);
       break;
-    case 'SET':
```
bradvogel: much cleaner!

lib/scripts/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+var fs = require('fs');
+var path = require('path');
+
+//
+// for some very strange reason, defining scripts with this code results in this error
+// when executing the scripts: ERR value is not an integer or out of range
+//
+module.exports = function(client){
+  loadScripts(client, __dirname);
```
bradvogel: why is this needed in addition to `loadScriptIfNeeded`?

lib/scripts.js
```diff
@@ -230,11 +238,57 @@ var scripts = {
         }
       });
   },
-  moveToCompleted: function(job, removeOnComplete){
+  _moveToCompleted: function(job, removeOnComplete){
```
bradvogel: I don't think this is called anymore

lib/scripts.js
```diff
@@ -28,6 +27,14 @@ function isCommandDefined(client, hash){
   return !!client[hash];
 }
 
+var path = require('path');
+var fs = require('fs');
+function loadScriptIfNeeded(queue, name){
```
bradvogel: It'd be nice if `scripts/index.js` exposed this API

lib/scripts.js
bradvogel: I think `Job.prototype.move()` should be removed to, which would allow us to remove the `move` script (which appears to be redundant).

lib/scripts/index.js
manast: This is what it really should be. I want to load all the scripts at start define them as ioredis commands. However this code does not work, the scripts give ERR when executing them. I have spent several hours trying to figure it out without success. On the other hand, ```loadScriptIfNeeded``` worked well. Ultimately I would like to revisit the code in *index.js* to make it work.

lib/scripts.js
manast: Yes. I have understood that there is no point in trying to have a generic move function, so I will have specific methods instead.

lib/scripts.js
manast: yes, you are right. I will move it.

manast:
I closed this PR in favour of this one  https://github.com/OptimalBits/bull/pull/496 which would be merged into 3.0.0 branch instead of master.

Issue #480, opened 2017-03-30
[Feature] re-architecting the non-polling mechanism


manast:
today we use the redis command *BRPOPLPUSH*, which allows us to move a job atomically from the wait list to the active list. The advantages are:
- very low latency to start processing a job (with empty queue).
- atomically move a job from one list to another.
- low CPU usage since no polling is required.

However, there are a number of disadvantages which, in my opinion are too much of a burden:
- cannot perform the move operation and other operations (such as job lock) atomically.
- requires a dedicated extra connection to redis.

So I have been thinking if we can have a new mechanism that includes the advantages without the named disadvantages. This is what I propose:
- poll only once after processing a job.
- use pubsub to notify workers that new jobs have been added.
- listen to close/connect events checking if there are elements to process in the wait list.

Basically when a worker starts, the first think he does is set the subscription to the "add job" event. After that it checks if there are jobs to process in the wait list, if not, he keeps idling until an "add job" event has been received. If the connection to redis is lost, start the process again by polling once the wait list.

When the worker finds a job in the queue, he can move it atomically from wait to active as well as to set the lock on the job.

If there is a guarantee that redis delivers all the published messages as long as there is a valid connection, then this mechanism should work. According to redis documentation: 
"*Because Redis Pub/Sub is fire and forget currently there is no way to use this feature if you application demands reliable notification of events, that is, if your Pub/Sub client disconnects, and reconnects later, all the events delivered during the time the client was disconnected are lost.*"
So the question is if we can "reliably" decide that a client has been disconnected, if so we can poll on the next reconnect and everything should be fine.

bradvogel:
Big +1 on this.

If we fix this and are finally able to move jobs from to `active` and lock them in a single atomic call (which we can't currently due to https://github.com/OptimalBits/bull/issues/258), we could:
1) greatly simplify checking for stalled jobs. This would allow us to rename `moveUnlockedJobsToWait` back to `checkForStalledJobs` since we won't have to check for Case B here: https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L442. We would no longer need to store `lockAcquired` in Redis. So `checkForStalledJobs` can simply just scan for all unlocked jobs in the active queue, and assume they're actually stalled.
[...]

PR #496, opened 2017-04-13
Atomic move

manast:
Implements #480 and #190


lib/job.js
```diff
@@ -71,80 +70,65 @@ Job.fromId = function(queue, jobId){
...
   Takes a lock for this job so that no other queue worker can process it at the
   same time.
 */
-Job.prototype.takeLock = function(renew, ensureActive){
-  var _this = this;
-  return scripts.takeLock(this.queue, this, renew, ensureActive).then(function(lock) {
-    if (lock){
-      _this.lock = lock;
-    } 
+Job.prototype.takeLock = function(){
```
bradvogel: This looks like it can be removed - right?

bradvogel:
Looks like there's a lot of dead code that can be removed now:
`Job.prototype.move`
`Job.prototype.takeLock`
`Job.prototype.lockKey`
`Job.prototype.releaseLock`
`Job.prototype.delayIfNeeded`
`Job.prototype.moveToCompleted`
`scripts.move`

lots of others... feels good to clean it up!

lib/scripts/moveToActive.lua
```diff
@@ -0,0 +1,42 @@
...
+  -- get a the lock
+  redis.call("SET", lockKey, ARGV[2], "PX", ARGV[3])
+  redis.call("LREM", KEYS[1], 1, jobId) -- remove from wait
+  redis.call("ZREM", KEYS[3], jobId) -- remove from priority
+  redis.call("LPUSH", KEYS[2], jobId) -- push in active
+  return redis.call("HGETALL", jobKey) -- get job data
+end
+
+--[[
+  Release lock:
```
bradvogel: What was this for?

lib/queue.js
```diff
@@ -726,65 +725,27 @@ Queue.prototype.processJob = function(job){
...
-    job.takeLock(true /* renew */, false /* ensureActive */).then( function(/*lock*/) {
-      return job.moveToFailed(err).finally(function(){
-        _this.processing--;
-      }).then(function(){
-        return _this.distEmit('failed', job, error);
-      });
-    }, function(err){
-      _this.emit('error', err, 'failed to re-obtain lock before moving to failed, bailing');
+    //  job.takeLock(true /* renew */, false /* ensureActive */).then( function(/*lock*/) {
+    return job.moveToFailed(err).then(function(){
```
bradvogel: I think there's a hazard here: the lock can expire immediately after `stopTimer` is called and before the job is moved to failed. If `moveUnlockedJobsToWait` happens to run then, it'll reprocess the job. That's why we previously called `job.takeLock` right before calling job.moveToFailed.

To fix this, I think `stopTimer` should be moved to be called in a `.finally()` block after `moveToFailed`. That way `moveToFailed` can assume it has a lock.

lib/job.js
```diff
@@ -57,7 +56,7 @@ Job.create = function(queue, name, data, opts){
   var job = new Job(queue, name, data, opts);
 
   return addJob(queue, job).then(function(jobId){
-    job.jobId = jobId;
+    job.id = jobId;
```
bradvogel: This API change should be noted in the changelog

lib/queue.js
```diff
@@ -148,18 +151,13 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
   _.extend(this.redlock, redisOptions.redlock || {});
 
   //
-  // Create blocking client (used to wait for jobs)
-  //
-  this.bclient = createClient('block');
```
bradvogel: One less connection - nice!

lib/queue.js
```diff
@@ -268,14 +277,14 @@ var Queue = function Queue(name, redisPort, redisHost, redisOptions){
...
   this.moveUnlockedJobsToWait = this.moveUnlockedJobsToWait.bind(this);
-  this.getNextJob = this.getNextJob.bind(this);
-  this.processJobs = this.processJobs.bind(this);
   this.processJob = this.processJob.bind(this);
   this.getJobFromId = Job.fromId.bind(null, this);
 };
 
 util.inherits(Queue, Disturbed);
 
+Queue.ErrorMessages = errors.Messages;
```
bradvogel: Note this in the changelog?

lib/queue.js
```diff
@@ -284,11 +293,11 @@ Queue.prototype.isReady = function(){
 }
 
 Queue.prototype.getJobMoveCount = function(){
-  return this.bclient.commandQueue.length;
+  return this.client.commandQueue.length;
 };
 
 Queue.prototype.whenCurrentMoveFinished = function(){
```
bradvogel: What was this for originally? Seems like it can be removed to reduce complexity.

lib/queue.js
```diff
@@ -284,11 +293,11 @@ Queue.prototype.isReady = function(){
 }
 
 Queue.prototype.getJobMoveCount = function(){
```
bradvogel: What was this for originally? Seems like it can be removed to reduce complexity.

lib/scripts.js
```diff
@@ -301,200 +213,66 @@ var scripts = {
...
-   * would be in 'active' but NOT have a job lock:
+   * Looks for unlocked jobs in the active queue. 
    *
-   *  Case A) The job was being worked on, but the worker process died and it failed to renew the lock.
+   *    The job was being worked on, but the worker process died and it failed to renew the lock.
    *    We call these jobs 'stalled'. This is the most common case. We resolve these by moving them
    *    back to wait to be re-processed. To prevent jobs from cycling endlessly between active and wait,
    *    (e.g. if the job handler keeps crashing), we limit the number stalled job recoveries to MAX_STALLED_JOB_COUNT.
-
-   *  Case B) The job was just moved to 'active' from 'wait' and the worker that moved it hasn't gotten
```
bradvogel: You might want to remove `This is the most common case` above since it's now the only case.

lib/scripts.js
```diff
@@ -584,131 +377,104 @@ var scripts = {
...
-  var _this = this;
+    return runScript(queue.client, 'reprocessJob', lua, keys, args);
+  },
 
-  // Get all jobids and empty all lists atomically.
-  var multi = this.multi();
+  //
+  // TODO: DEPRECATE.
+  //
+  move: function(job, src, target){
```
bradvogel: I don't think this is used anymore - remove?

lib/queue.js
manast: I do not think this could happen in practice since the lock extender timer executes twice as often as the lock expiration time. However, if I move it to finally it also passes the unit tests, so I can move it there. Ultimately I will also make moveToFailed atomic, to further minimize any possible hazards.

lib/scripts/moveToActive.lua
manast: old stuff. removed now.

lib/scripts/moveToFinished.lua
```diff
@@ -0,0 +1,56 @@
...
+      ARGV[3]  msg property
+      ARGV[4]  return value / failed reason
+      ARGV[5]  token
+      ARGV[6]  shouldRemove
+      ARGV[7]  event channel
+      ARGV[8]  event data (? maybe just send jobid).
+
+     Output:
+      0 OK
+      1 Missing key.
```
manast: fixed.

lib/job.js
manast: done.

lib/job.js
manast: It is only used by a unit test indirectly, will try to get rid of it.

lib/queue.js
manast: I have removed it.

lib/scripts.js
manast: yes, this has been removed.

lib/scripts.js
manast: Only used in a unit test. will remove it.

manast:
I pushed some updated addressing the code review. Regarding performance, I do not think it will affect in any measurable manner. If you have a busy queue, the publishing of 'added' event will not be used since the worker will just pick new jobs directly after processing the previous one. If not busy, the extra overhead of the publish will be negligible. Besides, in the previous blocking version we also published similar events, so the overhead was already there, and the blocking call was timed out after 2.5 seconds, effectively making it a polling queue. Overall the queue is much faster now, I will perform some benchmarks soon. The last thing remaining is a guard timmer and logic for handling the queue properly after a reconnection, we need better unit tests for that.

lib/job.js
bradvogel: Yeah, would be nice if the unit test implemented it itself. Just to keep the core queue implementation pure and simple.

lib/queue.js
```diff
@@ -678,113 +682,65 @@ Queue.prototype.processJob = function(job){
...
-    job.takeLock(true /* renew */, false /* ensureActive */).then( function(/*lock*/) {
-      return job.moveToFailed(err).finally(function(){
-        _this.processing--;
-      }).then(function(){
-        return _this.distEmit('failed', job, error);
-      });
-    }, function(err){
-      _this.emit('error', err, 'failed to re-obtain lock before moving to failed, bailing');
-    });
+    //  job.takeLock(true /* renew */, false /* ensureActive */).then( function(/*lock*/) {
```
bradvogel: remove commented out line

bradvogel:
Yes- makes sense. Nice work! Excited for this.

manast:
with these latest commits I feel ready for merging in 3.0.0 branch and I would like to release the first alpha version for the 3.x series.

bradvogel:
Looks good to me to merge

chuym:
FWIW with respect to event listeners approach - it is actually not as CPU intensive as a polling alternative, if anything, it can be a source of potential memory leaks when having an unbound number of event listeners, nodejs avoids this by having a default max limit of event listeners of 10 to prevent a potential memory leak. The `EventEmitter` class is actually pretty straightforward, here: https://github.com/nodejs/node/blob/master/lib/events.js

lib/commands/addJob-7.lua
```diff
@@ -0,0 +1,99 @@
...
+    local len = redis.call("LLEN", target)
+    local id = redis.call("LINDEX", target, len - (count-1))
+    if id then
+      redis.call("LINSERT", target, "BEFORE", id, jobId)
+    else
+      redis.call("RPUSH", target, jobId)
+    end
+  end
+
+  redis.call("PUBLISH", KEYS[4], jobId)
```
manast: good catch. I will fix it.

lib/commands/moveUnlockedJobsToWait-4.lua
```diff
@@ -0,0 +1,54 @@
+--[[ 
+  Looks for unlocked jobs in the active queue. There are two circumstances in which a job
+   would be in 'active' but NOT have a job lock:
+   
+     Case A) The job was being worked on, but the worker process died and it failed to renew the lock.
+       We call these jobs 'stalled'. This is the most common case. We resolve these by moving them
+       back to wait to be re-processed. To prevent jobs from cycling endlessly between active and wait,
+       (e.g. if the job handler keeps crashing), we limit the number stalled job recoveries to MAX_STALLED_JOB_COUNT.
+
+    DEPRECATED CASE:
```
manast: I will remove the comment. With single instance redis, this case could not happen anymore since the lock is obtained atomically with the move. In the future, if we want to support redis with replication, we need to re-introduce redlock, but we will be able to take the lock *before* moving the job, so in practice the case should not happen either.

lib/job.js
```diff
@@ -22,20 +26,15 @@ var Job = function(queue, name, data, opts){
     name = '__default__';
   }
 
-  opts = opts || {};
+  this.opts = _.extend({}, opts);
```
manast: I will update using _.defaults since it is also semantically better. I feel options should always be kept in the opts field only, so delay should be in opts instead of a property of job, however, delay can also be defined when retrying a job, and in that case it would not be an option but a property used by the queue mechanics. Timestamp is also a bit special since normally it should be an internal property, but  sometimes you want to override it, just as job.id.

lib/job.js
```diff
@@ -71,80 +71,64 @@ Job.fromId = function(queue, jobId){
...
     timestamp: this.timestamp,
-    attempts: this.attempts,
     attemptsMade: this.attemptsMade,
     failedReason: this.failedReason,
     stacktrace: this.stacktrace || null,
     returnvalue: this.returnvalue || null
   };
 };
 
+Job.prototype.toData = function(){
```
manast: yes, a nice simplification.

lib/job.js
manast: yes, it is not tested, but used in some unit test where we need to have a lock: https://github.com/OptimalBits/bull/blob/v3.0.0/test/test_job.js#L100

lib/commands/moveToFinished-3.lua
```diff
@@ -0,0 +1,56 @@
...
+      ARGV[3]  msg property
+      ARGV[4]  return value / failed reason
+      ARGV[5]  token
+      ARGV[6]  shouldRemove
+      ARGV[7]  event channel
+      ARGV[8]  event data (? maybe just send jobid).
+
+     Output:
+      0 OK
+      -1 Missing key.
```
manast: you are right. I added a -2 error code when a lock is missing. I noticed however that this error codes are not being used, I need to fix it as well, it may be hidding errors right now.

lib/commands/moveToFinished-3.lua
```diff
@@ -0,0 +1,56 @@
...
+
+     Events:
+      'completed'
+]]
+
+if redis.call("EXISTS", KEYS[3]) == 1 then -- // Make sure job exists  
+  if ARGV[5] ~= "0" then
+    local lockKey = KEYS[3] .. ':lock'
```
manast: I will add issue for such an improvement.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+v.3.0.0-alpha
+=============
+
+- job.jobId changed to job.id.
+- refactored error messages into separate error module.
+- completed and failed job states are now represented in ZSETs.
```
manast: - job.jobId -> job.id  to reduce redundancy.
- error messages refactored so that we can unit test without needing to copy/paste the messages.
- ZSETs allows us to efficiently get ranges of completed and failed jobs, very useful for building UIs and scripts.

manast:
@chuym thanks for the review I will commit the fixes directly to the 3.0.0 branch.

Issue #488, opened 2017-04-06
[Feature] Support for running jobs in child processes


manast:
Many times, processes are long and synchronous. Currently design does not allow to run these kind of processes properly due to the locking mechanism. However, we can spawn a child process where the job is run instead. If the same child process is reused for every job (and we will need as many child processes as concurrency levels), the overhead will be minimum. Also as an added benefit, if the child process dies for any reason (memory leaks, or other cause), the worker will still be able to spawn a child replacement and continue working normally.

PR #504, opened 2017-04-21
Sanitized options

manast:
This PR aims to cleanup the constructor of the Queue and make the options more sane. Also support for configuring a lot of settings that were constants before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,14 @@ Roadmap:
 Install:
 --------
 
-    npm install bull
+    npm install bull@2.x --save
 
 Note that you need a redis version higher or equal than 2.8.11 for bull to work properly.
+
+**IMPORTANT**
+
+We are in the progress of developing ```bull 3.0.0```, which means that the latest *unstable* version would be something like
+bull-3.0.0-alpha.1. It is recommended that you stick to version 2.x until 3.0.0 stable is released. Some things to expect in  3.x: https://github.com/OptimalBits/bull/milestone/4
 
 Quick Guide
 -----------
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Bull Job Manager
 
 <img src="https://image.freepik.com/free-icon/strong-bull-side-view_318-52710.jpg" width="200" />
 
-The fastest, most reliable redis based queue for nodejs.
+The fastest, most reliable Redis based queue for nodejs.
 Carefully written for rock solid stability and atomicity.
 
 
@@ -50,8 +50,6 @@ There are a few third party UIs that can be used for easier administration of th
 * [react-bull](https://github.com/kfatehi/react-bull)
 * [toureiro](https://github.com/Epharmix/Toureiro)
 
-We also have an official UI which is at the moment bare bones project: [bull-ui](https://github.com/OptimalBits/bull-ui)
-
 Roadmap:
 --------
 
@@ -66,7 +64,7 @@ Install:
 
     npm install bull@2.x --save
 
-Note that you need a redis version higher or equal than 2.8.11 for bull to work properly.
+Note that you need a Redis version higher or equal than 2.8.11 for bull to work properly.
 
 **IMPORTANT**
 
@@ -78,9 +76,9 @@ Quick Guide
 ```javascript
 var Queue = require('bull');
 
-var videoQueue = Queue('video transcoding', {redis: {port: 6379, host: '127.0.0.1'}});
-var audioQueue = Queue('audio transcoding', 'redis://127.0.0.1:6379');
-var imageQueue = Queue('image transcoding');
+var videoQueue = new Queue('video transcoding', 'redis://127.0.0.1:6379');
+var audioQueue = new Queue('audio transcoding', {redis: {port: 6379, host: '127.0.0.1'}}); // Specify Redis connection using object
+var imageQueue = new Queue('image transcoding');
 var pdfQueue = new Queue('pdf transcoding');
 
 videoQueue.process(function(job, done){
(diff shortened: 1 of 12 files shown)


manast:
@bradvogel  @chuym new PR you may be interested of. This is a small one, will break backwards compatibility.

bradvogel:
Awesome! Will review shortly

README.md
```diff
@@ -78,10 +78,10 @@ Quick Guide
 ```javascript
 var Queue = require('bull');
 
-var videoQueue = Queue('video transcoding', 6379, '127.0.0.1');
-var audioQueue = Queue('audio transcoding', 6379, '127.0.0.1');
-var imageQueue = Queue('image transcoding', 6379, '127.0.0.1');
-var pdfQueue = Queue('pdf transcoding', 6379, '127.0.0.1');
+var videoQueue = Queue('video transcoding', {redis: {port: 6379, host: '127.0.0.1'}});
```
bradvogel: Might be clearer to use the redis string version here, eg `redis://127.0.0.1:6379`, since the documentation lists the `redisConnectionString` as the second option (albeit optional - but some people might not notice that)

README.md
```diff
@@ -78,10 +78,10 @@ Quick Guide
...
+var audioQueue = Queue('audio transcoding');
+var imageQueue = Queue('image transcoding');
+var pdfQueue = Queue('pdf transcoding');
```
bradvogel: I think it's clearer to use use `new` in the examples, e.g. `new Queue('pdf transcoding');`, so it's clearer that an object instance is returned.

README.md
```diff
@@ -79,9 +79,9 @@ Quick Guide
 var Queue = require('bull');
 
 var videoQueue = Queue('video transcoding', {redis: {port: 6379, host: '127.0.0.1'}});
-var audioQueue = Queue('audio transcoding');
+var audioQueue = Queue('audio transcoding', 'redis://127.0.0.1:6379');
 var imageQueue = Queue('image transcoding');
-var pdfQueue = Queue('pdf transcoding');
+var pdfQueue = new Queue('pdf transcoding');
```
bradvogel: for the examples above too, for consistency.

README.md
```diff
@@ -390,34 +386,40 @@ listened by some other service that stores the results in a database.
...
-    redisConnectionString {String} A connection string containing the redis server host, port and (optional) authentication.
-    redisOptions {Object} Options to pass to the redis client. https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
+```typescript
+  interface QueueOptions {
+    prefix?: string = 'bull',
+    redis : RedisOpts, // ioredis defaults
+    createClient?: (type: enum('client', 'subscriber'), redisOpts?: RedisOpts) => redisClient,
+
+    // Advanced settings
+    settings?: QueueSettings {
```
bradvogel: I'm working on a PR to document these

README.md
```diff
@@ -390,34 +386,40 @@ listened by some other service that stores the results in a database.
...
+      lockDuration?: number = 5000,
+      stalledInterval?: number = 5000,
+      maxStalledCount?: number = 1, // The maximum number of times a job can be recovered from the 'stalled' state
+      guardInterval?: number = 5000,
+      retryProcessDelay?: number = 5000
```
bradvogel: Should include `lockRenewTime`?

bradvogel:
Created PR https://github.com/OptimalBits/bull/pull/507/files into this PR. Let me know how that looks.

The readme is coming along nicely!

README.md
```diff
@@ -390,34 +386,40 @@ listened by some other service that stores the results in a database.
...
-    redisOptions {Object} Options to pass to the redis client. https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options
+```typescript
+  interface QueueOptions {
+    prefix?: string = 'bull',
+    redis : RedisOpts, // ioredis defaults
+    createClient?: (type: enum('client', 'subscriber'), redisOpts?: RedisOpts) => redisClient,
+
+    // Advanced settings
+    settings?: QueueSettings {
+      lockDuration?: number = 5000,
```
bradvogel: What do you think about defaulting the lock to 30 seconds? My thoughts:
-A lot of people complain about jobs being double processed currently. So there must be a lot of poorly written job processor code out there that stalls the event loop :). Or other folks, like me, forget that we're running our code on tiny instances in the cloud where we forget that that it will max out the limited CPU. A 30sec timeout would give a bit more buffer. At least until we figure out a generic solution like #488.
-The tradeoff is that it will take longer for jobs to be considered 'stalled'. So instead waiting max 5sec to find out if a job was stalled, we'd wait 30sec. I think this is generally OK and that most people aren't running jobs that are that time-sensitive. Actual stalled jobs [due to process crashes] should be extremely rare anyways.

What do you think?

README.md
bradvogel: Added this in #507

README.md
bradvogel: #507

README.md
```diff
@@ -390,34 +386,40 @@ listened by some other service that stores the results in a database.
...
+      stalledInterval?: number = 5000,
```
bradvogel: Per above comment, I think this should be moved to 30sec also. It doesn't do much good to make it a lot less than the lock time. Also note that since this is run once per process, if it was set to 30sec here, across 6+ worker processes it'd get called about every 5sec anyways.

README.md
manast: Ok. My idea was to show different ways to instantiate in the examples. I could clarify with a comment instead. I like to be able to instantiate without new, but maybe this is not a best practice anymore.

README.md
manast: I guess it is ok to have a larger value as default, specially now when it can be configured. I also look forward #488, which for long duration jobs should work much better.

PR #509, opened 2017-04-23
fix undefined ref

bradvogel:


manast:
good catch. I wonder why eslint does not find this kind of issues automatically :/.

PR #510, opened 2017-04-23
Fix regression where lock where lock renewer wasn't stopped immediate…

bradvogel:
…ly when the job completed. This wasn't a functional regression though (hence why it didn't fail any tests) because `timerStopped` prevented the timer from running again.


manast:
thanks.

PR #512, opened 2017-04-23
longer default lock timeout

bradvogel:
Per https://github.com/OptimalBits/bull/pull/504#discussion_r112824888


bradvogel:
Looking into the failed test in Travis...

bradvogel:
Ok, fixed test. Should be good to go

PR #514, opened 2017-04-23
Remove unused files. If these are needed for manually testing locally…

bradvogel:
…, we could put them in a new /private/sandbox dir?


bradvogel:
Merged this into https://github.com/OptimalBits/bull/pull/515 since it's needed for Travis to pass

PR #339, opened 2016-08-26
Reorganization - sanitzing lib folder

fearphage:
This does a lot. I would normally submit multiple PRs for this, but I thought I'd try to one-shot it. This does several things:
- move sample files from `lib` to `examples` folder
- replaced `gulp-eslint` with `eslint`
- moved eslint config to `package.json` (and test eslint config to `test/.eslintrc.json`)
- all files in `lib` pass eslint now
- using `mocha.opts` file for mocha options
  
  Let me know what you think.

This is related to comments in (and supercedes I believe) #314.

PR #515, opened 2017-04-23
[see comment] eslint cleanup, removed dead files

bradvogel:
Not ready to merge yet. See comment below


bradvogel:
@manast I just saw https://github.com/OptimalBits/bull/pull/339 after doing this. I like it better - it seems cleaner to not use gulp if we don't have to. What do you think? I can pull that PR into this PR (by removing gulp from here).

Also- what do you want to do with those dead files? Keep them in the repro under a different director (`/examples` like in the other PR)?

fearphage:
> I can pull that PR into this PR (by removing gulp from here).

Why not merge the other branch and then rebase? That would the set of changes here smaller and easier to grok.

bradvogel:
@fearphage you're right - let's work off your PR. Closing this PR. I left some comments on https://github.com/OptimalBits/bull/pull/339 for when you get a chance. If not, let me know and I can probably do it.

PR #517, opened 2017-04-25
Remove unused function `renewLock()` now that we call `extendLock`.

bradvogel:


test/test_job.js
```diff
@@ -218,16 +218,6 @@ describe('Job', function(){
...
-      return job.takeLock().then(function(lockTaken){
-        expect(lockTaken).to.be.truthy;
-      }).then(function(){
-        return job.renewLock().then(function(lockRenewed){
-          expect(lockRenewed).to.be.truthy;;
-        });
-      });
-    });
-
     it('can release a lock', function(){
```
manast: wouldn't be good to keep the unit test, but of course renaming renewLock to extendLock?

bradvogel: We could - but is the only value in making `extendLock` public for the unit tests? We wouldn't want it to be used publicly, since external code should be manipulating the job locks.

manast: sorry, I thought extendLock was already public.

PR #520, opened 2017-04-27
Script events. fixes #501

### LINKED ISSUE #501: Emit global events from inside the lua scripts
Currently we emit global events using an external module *disturbed*: https://www.npmjs.com/package/disturbed

This means that some operation, for example adding a job to the queue, is executed atomically via a lua script but the event is not emitted until the script has returned. Emitting the event inside the lua script itself has 2 advantages, firstly, it is a bit faster since a roundtrip to the redis instance is avoided, and secondly, it reduces a bit the chances for hazards, where at least we can guarantee that the event was emitted if the operation succeeded. It also makes the code more modular, since the events are actually emitted close to their context, instead of being emitted in an arbitrary place in javascript land.

manast:
In this PR we get rid of the "disturbed" dependency, and instead we publish global events from the lua scripts. This is faster and more reliable. For local events, we try to emit them directly. Local events and global are not identical, I will need to improve the documentation on this ASAP. For example while the local progress event sends the job as first parameter (since this can be get for free), in the global event we only send the jobId, to avoid fetching the job data everytime the progress is updated.

I also did some minor refactoring and removed the obsolete priority queue code and unit tests.


manast:
not sure why one of the tests fails, locally it passes all of the tests every time.

bradvogel:
Awesome!

MIGRATION.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+# Events
+
+All events are now published atomically in the scripts where they are relevant, this increases efficiency and
+reduces chances for hazards.
+
+'ready' event has been removed, you can use ```Queue##isReady()``` instead if you want to know when the queue
+has been initialized. Normally you will never need to wait for readyness since this is taken care internally
+by the queue methods that require the queue to be <ready class=""></ready>
```
bradvogel: `<ready class=""></ready>` might be a typo

MIGRATION.md
```diff
@@ -0,0 +1,39 @@
+# Migration from 2.x to 3.0.0
+
+Although version 3.0 is almost backwards compatible with 2.x, there are some important changes that needs
+to be taking in consideration before upgrading to 3.0.
+
+# Complete and failed sets.
+
+In 3.x, the jobs that are completed and failed end in two ZSETS, instead of a standard SET.
+This gives the posibility of retrieving a subset of the jobs in a high performant way, which
+is useful for grafical tools and scripts. However an old queue will not be compatible with 3.x.
```
bradvogel: graphical

MIGRATION.md
```diff
@@ -0,0 +1,39 @@
+# Migration from 2.x to 3.0.0
+
+Although version 3.0 is almost backwards compatible with 2.x, there are some important changes that needs
+to be taking in consideration before upgrading to 3.0.
+
+# Complete and failed sets.
+
+In 3.x, the jobs that are completed and failed end in two ZSETS, instead of a standard SET.
+This gives the posibility of retrieving a subset of the jobs in a high performant way, which
```
bradvogel: possibility

PR #536, opened 2017-05-20
Assorted fixes

manast:
small misc. fixes found during 3.0 development

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -34,10 +34,10 @@ Promise.config({
     - completed (set)
     - failed (set)
 
-        --> priorities      -- >completed
+        --> priorities      -- > completed
        /     |            /
     job -> wait -> active
-        |    ^            \
+       \     ^            \
         v    |             -- > failed
         delayed
 */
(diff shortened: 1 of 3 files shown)


lib/queue.js
```diff
@@ -82,7 +82,7 @@ var Queue = function Queue(name, url, opts){
   }
 
   if(_.isString(url)){
-    opts = _.defaults(redisOptsFromUrl, opts);
+    opts = {redis: _.defaults(redisOptsFromUrl(url), opts)};
```
bradvogel: I don't think this is right - it merges `opts` into the redis options, which is not what the user expects. I think it should be like https://github.com/OptimalBits/bull/pull/538/files#diff-b68fbb0f945cdf9cddec65ddf0fbe456R85 where we only override `opts.redis` and merge in the user-passed options.

bradvogel: I don't think this is right - it merges opts into the redis options, which is not what the user expects. I think it should be like https://github.com/OptimalBits/bull/pull/538/files#diff-b68fbb0f945cdf9cddec65ddf0fbe456R85 where we only override opts.redis and merge in the user-passed options.

manast: yes, that is completely wrong, I guess it was too late for coding when i did that :)

bradvogel:
this lgtm, though the tests are failing in CI

manast:
yes, it is strange, even without the change they fail I have to look into it.

manast:
maybe not so strange, there are other changes that breaks the tests

bradvogel:
this lgtm to be merged, if you can get tests to pass

bradvogel:
v3.0.0 is so close! Hope to start using it this week and use it on one of our high-volume production services :)

PR #537, opened 2017-05-20
better error propagation moveToFinished #499

manast:


lib/scripts.js
```diff
@@ -124,7 +124,12 @@ var scripts = {
 
   moveToFinished: function(job, val, propVal, shouldRemove, target, ignoreLock){
     var args = scripts.moveToFinishedArgs(job, val, propVal, shouldRemove, target, ignoreLock);
-    return job.queue.client.moveToFinished(args);
+    return job.queue.client.moveToFinished(args).then(function(result){
+      switch(result){
+        case -1: throw Error('Missing key for job ' + job.id);
+        case -2: throw Error('Missing lock for job ' + job.id);
```
bradvogel: shouldn't this fire the 'failed' event? I'm seeing in this PR that it doesn't

bradvogel: This probably also needs a test case too

manast: yes, the error is not being propagated, it is a bit tricky though since it is the script is executed in a multi command. Basically what is needed is to first mark that we run moveToFinished and then pick the last result from multi. Will work on it now.
https://github.com/OptimalBits/bull/blob/movetofinished-errors/lib/job.js#L186-L189

manast:
I think this is done. But it needs to be merged into v3. I will do it after merging v3-docs though

bradvogel:
looks good to me! Please get this into the RC :)

manast:
I made a final commit. Note that I had to change these tests as well: https://github.com/OptimalBits/bull/blob/movetofinished-errors/test/test_queue.js#L1496-L1541
Before it was listening to the 'failed' event, but in reality, in this case the job will not really fail, it will error since it cannot get the lock for the job, so therefore is the error event the one that should trigger (and fixed the code so that failed is not emitted anymore).

PR #538, opened 2017-05-21
Fix instantiation

bradvogel:
Fix the constructor so passing the redis connection works: (e.g. `var queue = new Queue('connstring', 'redis://localhost:6379');`). This wasn't caught by the unit tests because they happened to be using the same host and port number as the defaults.


manast:
hehe, this was quite funny, I actually fixed the same error yesterday, I guess great minds think alike :D. https://github.com/OptimalBits/bull/pull/536
Although I had to change a bit the test case so that I could close the queue, since normally .close waits for isReady, which will never be resolved if using a connection string to an un-existing redis DB. So I think I will keep my change instead of this because it also tests that the port is correct (by not using the default port value).

bradvogel:
Oh, i see, thanks! Your test cases are stronger since they use a different host and port. My test just used 'localhost' instead of 127.0.0.1 to let the queue actually be created. But I didn't test a changed port like you did.

PR #539, opened 2017-05-21
MoveToDelayed lock

manast:


lib/commands/moveToDelayed-3.lua
```diff
@@ -8,25 +8,28 @@
```
bradvogel: retryJob also needs a lock check

lib/commands/moveToDelayed-3.lua
```diff
@@ -8,25 +8,28 @@
...
 
   Events:
     - delayed key.
 ]]
 if redis.call("EXISTS", KEYS[3]) == 1 then
-  local score = tonumber(ARGV[1])
-  if score ~= 0 then
+  local lockKey = KEYS[3] .. ':lock'
+  local lock = redis.call("GET", lockKey)
+  if (not lock) or (lock == ARGV[3]) then
```
bradvogel: `not lock` should be removed since another worker might have already completed the job and deleted the lock. When I was trying to fix this locally, I realized that a bunch of unit tests broke when enforcing that a lock exists in this script. I fixed this by having `Job.prototype.moveToDelayed` take an `ignoreLock` parameter that passed in `0` as the token to the script, and checked it inside the script using:

```
  if ARGV[4] ~= "0" then
    local lockKey = KEYS[3] .. ':lock'
    if redis.call("GET", lockKey) ~= ARGV[4] then
      return -2
    end
  end
```

(this is how it's done in moveToFinished). That way the unit tests could use the special moveToDelayed() function that can ignore the lock.

lib/scripts.js
```diff
@@ -149,30 +149,30 @@ var scripts = {
...
-    return keys.concat([JSON.stringify(context), jobId]);
+    return keys.concat([JSON.stringify(timestamp), jobId, queue.token]);
   },
 
-  moveToDelayed: function(queue, jobId, context){
-    var args = scripts.moveToDelayedArgs(queue, jobId, context);
+  moveToDelayed: function(queue, jobId, timestamp){
+    var args = scripts.moveToDelayedArgs(queue, jobId, timestamp);
     return queue.client.moveToDelayed(args).then(function(result){
       if(result === -1){
```
bradvogel: check `-2` error code here?

lib/commands/moveToDelayed-3.lua
manast: ok. Will do that, thanks for the insight.

bradvogel:
Here are test cases:

```
    it('an unlocked job should not be moved to delayed', function(done) {
      var jobCallback;

      queue.process(function(job, callback) {
        jobCallback = callback;
      });

      queue.on('failed', function(job) {
        job.isDelayed().then(function(isDelayed) {
          expect(isDelayed).to.be.equal(false);
          done();
        });
      });

      queue.add({ foo: 'bar' }, { backoff: 1000, attempts: 2 }).then(function(job) {
        return job.releaseLock().then(function() {
          jobCallback(new Error('retry this job'));
        });
      });
    });

    it('an unlocked job should not be moved to waiting', function(done) {
      var jobCallback;

      queue.process(function(job, callback) {
        jobCallback = callback;
      });

      queue.on('failed', function(job) {
        job.isWaiting().then(function(isWaiting) {
          expect(isWaiting).to.be.equal(false);
          done();
        });
      });

      queue.add({ foo: 'bar' }, { backoff: 0, attempts: 2 }).then(function(job) {
        return job.releaseLock().then(function() {
          jobCallback(new Error('test'));
        });
      });
    });
```

manast:
ok, I fixed the code according to the review. However I have not been able to make you tests working yet.

bradvogel:
ok, I'll take a look

bradvogel:
oh - you'll also need to add the check to retryJob.lua

manast:
done!

test/test_queue.js
```diff
@@ -1492,6 +1492,50 @@ describe('Queue', function () {
         });
       });
     });
+
+    it.skip('an unlocked job should not be moved to delayed', function(done) {
```
bradvogel: Here are better test cases :)

```js
    it('an unlocked job should not be moved to delayed', function(done) {
      var queue = new Queue('delayed queue');

      queue.process(function(job, callback) {
        // Release the lock to simulate the event loop stalling (so failure to renew the lock).
        job.releaseLock().then(function() {
          // Once it's failed, it should NOT be moved to delayed since this worker lost the lock.
          callback(new Error('retry this job'));
        });
      });

      queue.on('failed', function(job) {
        job.isDelayed().then(function(isDelayed) {
          expect(isDelayed).to.be.equal(false);
          queue.close().finally(done);
        });
      });

      queue.add({ foo: 'bar' }, { backoff: 1000, attempts: 2 });
    });

    it('an unlocked job should not be moved to waiting', function(done) {
      var queue = new Queue('delayed queue');

      queue.process(function(job, callback) {
        // Release the lock to simulate the event loop stalling (so failure to renew the lock).
        job.releaseLock().then(function() {
          // Once it's failed, it should NOT be moved to waiting since this worker lost the lock.
          callback(new Error('retry this job'));
        });
      });

      queue.on('failed', function(job) {
        job.isWaiting().then(function(isWaiting) {
          expect(isWaiting).to.be.equal(false);
          queue.close().then(done, done);
        });
      });

      // Note that backoff:0 should immediately retry the job upon failure (ie put it in 'waiting')
      queue.add({ foo: 'bar' }, { backoff: 0, attempts: 2 });
    });
```

bradvogel:
This should be merged into v3.0.0 instead of v3-docs - right?

test/test_queue.js
manast: ok, they are passing now. Btw I moved back this ```queue.close().finally(done);``` to ```queue.close().then(done, done);``` The problem with finally is that it does not catch errors, it will always call done without arguments.

bradvogel:
lgtm to merge. Would you mind also pushing a new alpha release to NPM? I'll start testing on real data.

manast:
I think it should be RC, I do not have more features for this release.

bradvogel:
Awesome!

PR #548, opened 2017-05-26
Optimization: when removing from the wait list, start from the right …

bradvogel:
…side where the job is. This should reduce Redis CPU.

We're using Bull 3.0 in production (working great so far!), but we're noticing that Redis CPU is proportional to the number of jobs in `wait` (below: BytesUsedForCache at 400MB is about 400k jobs in `wait`): 
image

This hopefully fixes it. I can't think of any other place where were iterate the `wait` queue - can you?

### DIFF
--- a/lib/commands/moveToActive-4.lua
+++ b/lib/commands/moveToActive-4.lua
@@ -25,7 +25,7 @@ if jobId then
   
   -- get a the lock
   redis.call("SET", lockKey, ARGV[2], "PX", ARGV[3])
-  redis.call("LREM", KEYS[1], 1, jobId) -- remove from wait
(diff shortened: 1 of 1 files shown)


bradvogel:
Unfortunately I don't think this can be unit tested because it's functionally equivalent

manast:
Looks good. However for FIFO queues it would be better to use 1 instead of -1, so the value should be an argument. I can fix it myself.

bradvogel:
Oh actually, we can just use a RPOP. Moved discussion here: https://github.com/OptimalBits/bull/issues/550. FIFO queues wouldn't matter here because FIFO just means that the job is added to the right side of the wait list. Jobs are always moved to active from the right side, regardless of LIFO vs FIFO

PR #556, opened 2017-05-31
Fix promise not returned

juanborre:
This is a suggested fix to the error: 
```
Warning: a promise was created in a handler at Users/juanborreguero/dev/UpdateStaticData-Dispatcher/node_modules/bull/lib/queue.js:779:13 but was not returned from it, see http://goo.gl/rRqMUw
```
I had that when handling the:
- .on('completed', ...);
- .on('failed', ...);
events.

Feel free to comment or to fix it otherwise.
Thanks for the great job, guys.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -783,6 +783,7 @@ Queue.prototype.processJob = function(job){
 
     return job.moveToCompleted(result).then(function(){
       _this.emit('completed', job, result, 'active');
+      return null;
     });
   }
 
@@ -791,6 +792,7 @@ Queue.prototype.processJob = function(job){
 
     return job.moveToFailed(err).then(function(){
       _this.emit('failed', job, error, 'active');
+      return null;
     });
   }
 


bradvogel:
lgtm

lib/queue.js
```diff
@@ -783,6 +783,7 @@ Queue.prototype.processJob = function(job){
 
     return job.moveToCompleted(result).then(function(){
       _this.emit('completed', job, result, 'active');
+      return null;
```
manast: what about returning the result from .emit? should be null as well, looks a but nicer :D

juanborre: Maybe. I just applied the same fix I saw here: https://github.com/OptimalBits/bull/commit/1274f3ced5024c493c84bc4bb5f2246a28b44209
Also, there were others`return null;` in this file before this modification. We could replace all of them if you want.

We would be returning a promise. Can this be problematic to support both callbacks and promise based behavior from outside the package?

manast: I was wrong, emit does indeed return true/false.

juanborre: Thank you, guys!

Issue #557, opened 2017-06-01
[performance] `moveToActive` should only be called once per queue a job notification


bradvogel:
Moving conversation from https://github.com/OptimalBits/bull/issues/553#issuecomment-304788478 here

We currently have an inefficiency where when a queue receives a new job notification, it calls the `moveToActive` script `concurrency` times. This creates a lot of extra Redis traffic (and Node CPU burn) for queues with high-concurrency.

To repro:
1) Above line https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L719, add:
```
          .then(job => {
            console.log('getNextJob',!!job)
            return job;
          })
```
2) Instantiate a new queue with concurrency=40
3) Submit a job to that queue

You'll see `getNextJob true` logged 1 time and `getNextJob false` get logged 39 times.

A better approach is to have the queue itself listen for job notifications and then dispatch it to an available worker.

manast:
Yeah, it is a bit tricky actually. I was thinking about an approach where you could have a special event emitter that only emits to one of its listeners. For that to work I would need to setup the listener directly after the call to moveToActive when it returns null instead of a job(indicating that no job was in the queue, however I am not sure if a job could slip between the returning of moveToActive and my listener setup. If this does not happen vey often, the job that slips would be catched by the guard a few seconds later... I have a vague memory that some tests failed with a timeout if I moved the listener setup to after moveToActive, I will make some experiments and let you know.

bradvogel:
You're right. This will take some rearchitecting of how jobs are processed. I can take a look tomorrow.

bradvogel:
@ randallm working on this

bradvogel:
@randallm working on this

manast:
It would be great if we share his thoughts before starting the implementation.

manast:
I think that the easiest thing to test is to change the listener from line https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L839 to line https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L865
And use a special listener that only dispatches an incoming event to one of its listeners. Since we only listen when the worker actually is idling, and only dispatches to one of the workers, the problem should be solved. My only reservation is that maybe some event could be missed, but maybe it is just paranoia.  After we return from https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L855 I think that no event would have time to be received before we setup the listener since we are in the same event loop.

randallm:
[...]

PR #559, opened 2017-06-03
[proof of concept] eliminate duplicate moveToActive calls

randallm:
With regards to: https://github.com/OptimalBits/bull/issues/557

I took @bradvogel's idea of having the queue itself listen for added job events. This way, we do not have `moveToActive` called `concurrency` times on every `Job.process` call.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -118,6 +118,8 @@ var Queue = function Queue(name, url, opts){
   this.keyPrefix = redisOpts.keyPrefix || opts.prefix || 'bull';
   this.token = uuid();
 
+  this.numAdded = 0;
+
   //
   // We cannot use ioredis keyPrefix feature since we
   // create keys dynamically in lua scripts.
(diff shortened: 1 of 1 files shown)


bradvogel:
FYI you can close this @manast. We have more work to do and will open a new PR

manast:
@bradvogel I am wondering, what are you planing for this feature?

bradvogel:
Thanks for checking in. @randallm and I tested this change locally, and while we saw a significant improvement in CPU usage, it still wasn't down to Bull 1.0 benchmarks. So we're going to experiment some more and see if we can make more drastic changes.

manast:
I think a change than can have a nice impact in performance is to reduce the number of "added" events emitted. The 'added' event is only useful if the queue was drained and we have idling workers. As long as the queue has at least one element in the wait list, the 'added' event is redundant. So I think we can just check if the queue was empty before adding a job, and only if so emit the 'added' event.

Issue #558, opened 2017-06-02
Question About Expected Behavior for Repeating Jobs


andrewburgess:
Bull version: `3.0.0-rc.2`

I'm using repeating jobs with a specific job id to schedule a task to run at a specific hour every day. I have the process set up so that it runs `queue.add` on startup to ensure that the job exists. Looking through the `queue` implementation, every call to `queue.add` will execute `nextRepeatableJob`, which calculates the next iteration based on what is currently in bull.

Based on my current implementation, I always call `queue.add` to ensure the job is scheduled. If I restart the process 10 times in a row, then I will have 10 jobs scheduled out for the next 10 days, and each time a job completes, it will continue to maintain 10 future jobs.

Is this expected behavior? My initial assumption was that only the immediate next interval would be scheduled, especially when specifying a unique jobId.

andrewburgess:
This sample also seems to be producing potentially invalid results: https://gist.github.com/andrewburgess/38688712f94fa07756a84c8469bf1c48

The sample spins up four forks of the process, and each fork attempts to add a job to the queue with a repeating interval of 10 minutes.

If you look at the output, two jobs are created, that are on different interval schedules. One seems to be scheduled for every 10 minutes, but starting at the first exact 10 minute interval of the hour (like specifying `0/10/20/30/40/50 * * * *` for crontab), and the second job is every 10 minutes from the time the job was specified.

manast:
thanks for the report, I may have overlooked how cron-parse generates the next timestamp, I will take a deeper look into it.

andrewburgess:
Poking around a little more, I think this line might be an issue: https://github.com/OptimalBits/bull/blob/a9308a75052f6165bcd09f0d56ea0b9234600c1c/lib/queue.js#L448

Should it be `var delay = nextMillis - Date.now()` instead of `var delay = nextMillis - millis`?

PR #563, opened 2017-06-05
fix(queue): fix calculation for next repeat job

andrewburgess:
Fix `delay` calculation. Previously, it would not work if `queue.add` was called multiple times, since the delay would always be the interval length instead of the amount of time from now to the next call.

Add `isRepeat` flag to change which iteration is calculated. Only calculate based on queue `millis` when we repeat to get the next job, otherwise use the current time so the immediate next interval is used.

Not sure if the `isRepeat` flag is the best way to handle this, but I think there needs to be a differentiation between calling `queue.add` and adding the next repetition of the job.

Addresses #558

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -419,14 +419,16 @@ Queue.prototype.process = function(name, concurrency, handler){
 //
 var parser = require('cron-parser');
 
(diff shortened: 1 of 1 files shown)


manast:
Thanks for the patch. looks good so I will merge.

PR #565, opened 2017-06-06
Implementation of lazy events

manast:


manast:
This PR starts to be ready. It should reduce the amount of traffic related to events since only events that are really being listened to are registered.

manast:
If this looks good I can merge it today.

bradvogel:
This lgtm

PR #585, opened 2017-06-23
Mention Typescript definitions in README

marshall007:
I recently updated the Typescript definitions to reflect the API changes in v3.0 (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/17002). Thought it might be worth mentioning in the README.


manast:
great thanks!

PR #586, opened 2017-06-23
Lazily initialize redis clients, reduce side effects of `Queue` instantiation

### LINKED ISSUE #526: More lightweight queue clients.
I have a job processor running in one container -- and a CLI and web server both of which adds jobs.

The model seems to be that all processes that create Queues become workers (i.e., the node process doesn't return). What's the approach to build something that is client-only?

marshall007:
Fixes #526

This PR ensures that the `client` and `eclient` redis connections are not instantiated until they are absolutely needed. This in turn reduces the number of bound event listeners by default.

Additionally, I have added a `skipVersionCheck` option which, when `true`, avoids creating a redis connection and sanity checking the `MINIMUM_REDIS_VERSION` in the constructor.

### DIFF
--- a/lib/commands/index.js
+++ b/lib/commands/index.js
@@ -1,14 +1,14 @@
 /**
  * Load redis lua scripts.
  * The name of the script must have the following format:
- * 
+ *
  * cmdName-numKeys.lua
- * 
+ *
  * cmdName must be in camel case format.
- * 
+ *
  * For example:
  * moveToFinish-3.lua
- * 
+ *
  */
 'use strict';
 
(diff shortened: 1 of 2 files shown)


marshall007:
@manast there are currently two failing tests for `Queue.close()` related to lazily creating the redis connections. If you could take a look, I'd like some guidance on what we should assert in those tests now. In both of those tests, the assertions are now creating the client by accessing them only to check that they are in a certain state.

lib/commands/index.js
```diff
@@ -27,19 +27,16 @@ module.exports = function(client){
...
+  })
+  .map(function (file) {
+    var longName = path.basename(file, '.lua');
+    var name = longName.split('-')[0];
+    var numberOfKeys = parseInt(longName.split('-')[1]);
 
-      return fs.readFileAsync(path.join(dir, file)).then(function(lua){
-        client.defineCommand(name, { numberOfKeys: numberOfKeys, lua: lua.toString() });
-      }, function(err){
-        console.error('Error reading script file', err);
```
marshall007: The error callback was safely removed because we already attach an error handler to `_this._initializing`.

manast: Its true. The only reason I kept the console.error is to give information about the origin of the error, but I guess that using Bluebird and long stack traces would work just as well.

lib/queue.js
```diff
@@ -92,55 +92,64 @@ var Queue = function Queue(name, url, opts){
     throw Error('Options must be a valid object');
   }
 
-  var redisOpts = opts.redis || {};
+  this.name = name;
+  this.token = uuid();
 
-  _.defaults(redisOpts, {
+  opts.redis = opts.redis || {};
+
```
manast: The problem with this approach is that we are modifying the passed opts object which could create side effects on  the user's code. However I am seeing that we are already creating side effects since redisOpts is not a copy of opts.redis but a reference. I think it would be wise to clone the options before modifying them.

manast:
thanks for the nice patch. I would need to debug the test to understand why it is failing. However, something that I know is not correct in the current implementation is that clients that are created externally via the createClient option, are also closed when closing the queue, which is not the proper behaviour. Somehow we should mark external created clients and not close them.

marshall007:
> However, something that I know is not correct in the current implementation is that clients that are created externally via the createClient option, are also closed when closing the queue, which is not the proper behaviour.

@manast are you saying that was the previous behavior that this PR breaks or that it's just something we need to do?

manast:
@marshall007 something that we need to do, a known issue so to speak.

PR #587, opened 2017-06-24
job finished onCompleted should capture result from process

saviogrossi:
oncomplete event always pass result value (if exists), so looks like natural job onComplete handler capture that value


saviogrossi:
Sorry, I haven't tested against master, only against on -rc3 tag, I will review what's different here.

PR #588, opened 2017-06-25
onCompleted on job.finished should capture result from queue process

saviogrossi:
The result of a .process(), when exists, is always passed  on 'completed' event, so it's make sense to capture that value on onCompleted handler.

Now this should works:
```
service1Queue.add({hello: 'world'}).then(function(job) {
    return job.finished();
}).then(function(result) {
   // do something with result...
}).catch(function(err) {
   // catch err
})
```

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -315,18 +315,16 @@ Job.prototype.remove = function(){
 Job.prototype.finished = function(){
   var _this = this;
 
-  function getFailed(){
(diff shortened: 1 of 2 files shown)


manast:
Thanks for the patch, I think this was something we overlook. However there is one more place where the result must be returned, if ```status == 1``` the job was finished before calling .finished(), so we need to read the job and return the returnValue: https://github.com/OptimalBits/bull/blob/master/lib/job.js#L329

PR #589, opened 2017-06-26
Lazily initialize redis clients, reduce side effects of `Queue` instantiation

### LINKED ISSUE #526: More lightweight queue clients.

marshall007:
Supersedes #586

Sorry I had to create a new PR, pushing additional commits onto my branch wasn't updating the existing PR. I think I ran into https://github.com/isaacs/github/issues/591.

---

Fixes #526

This PR ensures that the client and eclient redis connections are not instantiated until they are absolutely needed. This in turn reduces the number of bound event listeners by default.

Additionally, I have added a `skipVersionCheck` option which, when `true`, avoids creating a redis connection and sanity checking the `MINIMUM_REDIS_VERSION` in the constructor.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -400,7 +400,7 @@ Queue.prototype.whenCurrentMoveFinished = function(){
 Queue.prototype.disconnect = function(){
   // TODO: Only quit clients that we "own".
   var clients = [this.client, this.eclient].filter(function(client){
-    return client.status === 'ready';
+    return client.status === 'ready' || client.status === 'connecting';
   });
 
   var ended = new Promise(function(resolve){
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -86,7 +86,7 @@ var Queue = function Queue(name, url, opts){
     opts = url;
   }
 
-  opts = opts || {};
+  opts = _.cloneDeep(opts || {});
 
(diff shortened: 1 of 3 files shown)


marshall007:
@manast I have those two failing tests passing now, but I had to modify one of them. My second commit provides some clarity as to why those tests were failing now. The issue is that in these tests, you're actually creating the `eclient` by accessing it whereas the previous assumption was that the client had already been created in the constructor.

lib/queue.js
```diff
@@ -92,55 +92,64 @@ var Queue = function Queue(name, url, opts){
     throw Error('Options must be a valid object');
   }
 
-  var redisOpts = opts.redis || {};
+  this.name = name;
+  this.token = uuid();
 
-  _.defaults(redisOpts, {
+  opts.redis = opts.redis || {};
+
```
manast: as mentioned in the previous PR, we need to clone opts and opts.redis, or we will be modifying the user's options object which could create side effects.

marshall007: @manast fixed.

PR #591, opened 2017-06-27
Cache Lua scripts

marshall007:
As a followup to #589, this PR caches the Lua scripts in memory so that file system calls can be avoided when creating many subsequent `Queue` clients.


marshall007:
@manast the first two failing tests are related to race conditions that I can't figure out (the client is "ready" faster now that the `fs` calls aren't being made, and that's causing problems).

The third, `Queue emits waiting event when a job is added` is an odd one because the client status is `connecting` (not `connect`) on my local using Node `v7.7.2`.

manast:
ok, I will try to reproduce them-

PR #593, opened 2017-06-28
Eliminate need for abstractions by exposing a better `getJobs` and `getJobCounts`

marshall007:
The goal of this PR is to consolidate all the logic for counting and listing various job sets. If this PR is merged, all the individual `getType()` and `getTypeCount()` can be considered deprecated as the user can now just use `getJobs([types])` and `getJobCounts(types...)` respectively.

The one breaking change is that the `asc` parameter of `getJobs()` is now the last optional parameter. This made more sense if `getJobs()` is to be considered a public API since most users won't care about the `asc` parameter.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -601,14 +601,7 @@ Queue.prototype.add = function(name, data, opts){
   Returns the number of jobs waiting to be processed.
 */
 Queue.prototype.count = function(){
-  var multi = this.multi();
-  multi.llen(this.toKey('wait'));
-  multi.llen(this.toKey('paused'));
-  multi.zcard(this.toKey('delayed'));
-
-  return multi.exec().then(function(res){
-    return Math.max(res[0][1], res[1][1]) + res[2][1];
-  });
+  return this.getJobCountByTypes('wait', 'paused', 'delayed');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/queue.js
```diff
@@ -946,139 +939,127 @@ Queue.prototype.getJob = function(jobId){
...
+        if(asc){
+          return _this.client.lrange(key, -(end + 1), -(start + 1)).then(function(result){
+            return result.reverse();
+          });
+        }else{
+          return _this.client.lrange(key, start, end);
+        }
+      case 'zrange':
+        return asc ? _this.client.zrange(key, start, end) : _this.client.zrevrange(key, start, end);
+    }
```
marshall007: @manast I was initially hoping we'd be able to get rid of all this logic, or at least implement it as a transaction like we do with counts.

After working on this PR it became clear that a lot of what I was doing might be better encapsulated by a Lua script. If you think there's a better solution here let me know.

manast: Yes. Whereas getCounts got nicer, getJobs became more complex. So I think it would be better to keep getCounts as it was in the old version.

marshall007: @manast part of the reason it's more complex is that `getJobs` supports passing multiple job types either as (`wait,completed` or `[ 'wait', 'completed' ]`). It also removes the `ZSET/LIST` parameter, which makes it consumable as a public API without concerning yourself with implementation details.

The initial reason I started working on this PR was to be able to do `.getJobs(['failed', 'active'])`, which I think is still valuable.

manast: ok. I understand. However I think I will need to merge this into the 3.1.0 branch since it is breaking backwards compatibility.

marshall007: @manast ok sounds good. I'll see about cleaning this up a bit further before the v3.1.0 release.

manast:
The breaking change is indeed on a private function, or at least undocumented.

lib/queue.js
```diff
@@ -946,139 +939,127 @@ Queue.prototype.getJob = function(jobId){
...
+// Job counts by type
+// Queue#getJobCountByTypes('completed') => completed count
+// Queue#getJobCountByTypes('completed,failed') => completed + failed count
+// Queue#getJobCountByTypes('completed', 'failed') => completed + failed count
+// Queue#getJobCountByTypes('completed,waiting', 'failed') => completed + waiting + failed count
+Queue.prototype.getJobCountByTypes = function() {
+  return this.getJobCounts.apply(this, arguments).then(function(result){
+    return _.chain(result).values().sum().value();
+  });
 };
```
manast: I did not notice this change before. Getting all the jobs and then counting them is very slow operation compared to using llen or zcard. I will fix this myself in another branch.

marshall007: @manast sorry for the late response. This doesn't actually fetch all jobs, it just sums the values from the object response of `getJobCounts()`. For example, `{ waiting: 3, active: 5 } -> 8`.

manast: oh sorry, I missed that commandByType actually used zcard and llen for counts. :)

PR #596, opened 2017-06-29
Emit `drained` event when queue is empty

### LINKED ISSUE #595: [Feature] emit `drained` event when Queue is empty
Currently, there does not appear to be a reliable way to be notified when the queue is done processing all active jobs and there are no other remaining jobs in the queue. It would be nice if a `drained` event were emitted in this case indicating that there is no remaining work to be done.

Use case is for knowing when it's safe to close/empty/clean up a Queue without missing any jobs.

The only workaround I'm aware of is to manually check the result of `getJobCounts()` after every `completed` event.

marshall007:
Fixes #595

I added a new `drainDelay` setting, which is the amount of time to wait (in milliseconds) for new jobs to be added to the queue before emitting the `drained` event.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -161,7 +161,8 @@ var Queue = function Queue(name, url, opts){
     stalledInterval: 30000,
     maxStalledCount: 1,
     guardInterval: 5000,
-    retryProcessDelay: 5000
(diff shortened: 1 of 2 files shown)


manast:
thanks for this nice patch.

Issue #594, opened 2017-06-29
Store the timestamp when the job was completed or failed.


manast:
This is useful information. Today only the creation date is stored in the job.

PR #602, opened 2017-07-04
Store finished timestamp

manast:


manast:
fixes #594

lib/queue.js
```diff
@@ -1079,12 +1079,25 @@ Queue.prototype.getJobs = function(queueType, type, asc, start, end){
...
+          }
+          return jobIds;
+        });
       break;
   }
 
   return jobs.then(function(jobIds){
-    var jobsFromId = jobIds.map(_this.getJobFromId);
+    var jobsFromId = jobIds.map(function(jobId){
+      return Job.fromId(_this, jobId, timestamps[jobId]);
```
marshall007: @manast it appears the new `timestamp` argument to `Job.fromId()` is not referenced. What was the intention here?

manast: Hmm, this code is obsolete since we now store the finishedOn timestamp in the job itself. It should be correct in master though, I will verify it.

PR #603, opened 2017-07-04
Allow for multiple repeatable jobs with same cron

### LINKED ISSUE #601: Add multiple repeatable jobs with the same cron pattern
In my use case I have multiple repeatable jobs with the same cron pattern that are created and removed programatically. I do not know beforehand how many jobs will be at any given time.

According to this line https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L506 the uniqueness of repeatable job keys is determined by the prefix, queue name, 'repeat', job name and cron string. Which means I can only try and differentiate the keys with the queue or job name variables, both for which I need to instantiate separate queue workers to process them. This works well when the jobs are predefined before but is less convenient when jobs are added and removed on the go.

If I try to add multiple jobs with the same cron pattern they get the same key and therefore only one will exist at a given time. JobId seems to not play a role with repeatable jobs.

Is there a way to accomplish this or a workaround for the existing code?

Maximization:
Fixes #601 

It was a bit tricky to do becase here https://github.com/OptimalBits/bull/blob/master/lib/queue.js#L547 the jobId is overwritten so if you use opts.jobId for the repeatKey/customId you're going to end up with a key that looks like this: `repeat:__default__:repeat:__default__:LHjHbS8qDxXEzpqQQPAH4m:1499171120000:1499171130000`

I came up with a solution to store the initial jobId provided by the user in a separate variable on the opts object. Let me know what you think.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -506,7 +506,8 @@ var parser = require('cron-parser');
 
 function nextRepeatableJob(queue, name, data, opts, isRepeat){
   var repeat = opts.repeat;
-  var repeatKey = queue.toKey('repeat') + ':' + name + ':' + repeat.cron;
+  var jobId = isRepeat ? opts.userCustomId : opts.jobId;
+  var repeatKey = queue.toKey('repeat') + ':' + name + ':' + (jobId ? jobId + ':' : '') + repeat.cron;
 
   //
   // Get millis for this repeatable job.
@@ -537,7 +538,7 @@ function nextRepeatableJob(queue, name, data, opts, isRepeat){
       //
       // Generate unique job id for this iteration.
       //
-      var customId = 'repeat:' + name + ':' + nextMillis;
+      var customId = 'repeat:' + name + ':' + (jobId ? jobId + ':' : '') + nextMillis;
 
       //
       // Set key and add job should be atomic.
@@ -545,6 +546,7 @@ function nextRepeatableJob(queue, name, data, opts, isRepeat){
       return queue.client.set(repeatKey, nextMillis).then(function(){
         return Job.create(queue, name, data, _.extend(_.clone(opts), {
           jobId: customId,
+          userCustomId: isRepeat ? opts.userCustomId : opts.jobId,
           delay: delay < 0 ? 0 : delay,
           timestamp: Date.now()
         }));
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -506,8 +506,11 @@ var parser = require('cron-parser');
 
 function nextRepeatableJob(queue, name, data, opts, isRepeat){
   var repeat = opts.repeat;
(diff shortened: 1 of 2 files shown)


manast:
yeah, I see what you mean. We are loosing a bit the advantage of using opts.jobId if we need to make a new variable in opts. Then it would be almost better to put a new variable inside opts.repeat.jobId that we use if defined, and we do not need all the extra logic.

manast:
Maybe better something like this for more consistent opts.jobId behaviour when using repeatable jobs (I am not 100% sure this is the best approach yet):
```js
function nextRepeatableJob(queue, name, data, opts, isRepeat){
  var repeat = opts.repeat;
  if(!isRepeat && opts.jobId){
    opts.repeat.jobId = opts.jobId;
  }
  var repeatJobId = (opts.repeat.jobId) ? ':' + opts.repeat.jobId : "";
  var repeatKey = queue.toKey('repeat') + ':' + name + ':' + repeat.cron + repeatJobId;

...

 var customId = 'repeat:' + name + ':' + nextMillis + repeatJobId;

...

```

Maximization:
This would work too. Wouldn't it be better for bull UI libs if the millis/cron would be switched with the JobId? So instead of:

`repeat:__default__:1499177290000:54229d8f43b07802002514c3`
`repeat:__default__:1499177300000:54229d8f43b07802002514c3`
`repeat:__default__:1499177310000:54229d8f43b07802002514c3`

and

`repeat:__default__:*/10 * * * * *:54229d8f43b07802002514c3`

it would be

`repeat:__default__:54229d8f43b07802002514c3:1499177290000`
`repeat:__default__:54229d8f43b07802002514c3:1499177300000`
`repeat:__default__:54229d8f43b07802002514c3:1499177310000`

and

`repeat:__default__:54229d8f43b07802002514c3:*/10 * * * * *`

Code wise I think you could clean it up a bit by putting the logic in the toKey function. But that's just aesthetics. I can't come up with a better solution right now to make this happen. I'm also missing the overall lib overview you have.

manast:
Why would it be better to switch places for the jobId? I am thinking that since it is an optional argument, it should be last.

Maximization:
To me it seems better hierarchical wise: repeat > jobName > JobId > timestamp. JobName is optional as well, maybe we could have a default value for JobId too. In any case, both options work.

manast:
ok, I like your reasoning regarding the order.

manast:
yes, I think it looks better now. We would need at leat one test case for this as well :)

manast:
great!. I will merge this as soon as 3.0.0 is released.

manast:
I changed my mind, I will merge this before 3.0, because there may be some other changes needed in the repeatable jobs functionality that will break even more backwards compatibility. I will have to make an apology for users already using the RC in production...

Issue #614, opened 2017-07-11
SyntaxError: Unexpected token o in JSON at position 0 


themuflon:
Hi!

I'm getting an error when JSON.parse tries to parse the `job.returnvalue` when processing `job.finished`. Actually, [this](https://github.com/OptimalBits/bull/blob/master/lib/job.js#L330) is where it fails. Typically that would be a JSON string, but rarely, when under heavy load and running on slow/few CPUs (our continuous integration), it's an object (thus the character is "o").

I was wondering if there is anything I'm doing wrong that should be common practice or if it's a bug? I went through the bull codebase but can't figure out what's going wrong.

Also, here's a code snippet of the code that produces the issue.
```
static postJob(queue:Queue.Queue, args:any):Promise<any> {
        var result = queue.add(args).then((_job) => {
            return _job.finished();
        });

        return Promise.resolve(result);
    }
```
The processing side is nothing special either, just some code that eventually calls `done()` with an object.

marshall007:
I believe this was fixed in #580, which was released in `v.3.0.0-rc.3`. @themuflon I assume you're still on `v.3.0.0-rc.2`?

themuflon:
I'm on `v.3.0.0-rc.4`. Like I said, this shows up only under specific circumstances. 99.99% of the time it works fine.

manast:
@themuflon could you by any chance get an error callstack? that would be quite useful to help finding the root cause of this.

themuflon:
```SyntaxError: Unexpected token o in JSON at position 1
    at Object.parse (native)
    at /usr/local/src/app/node_modules/bull/lib/job.js:330:24
    at tryCatcher (/usr/local/src/app/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/local/src/app/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/usr/local/src/app/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/usr/local/src/app/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/usr/local/src/app/node_modules/bluebird/js/release/promise.js:693:18)
    at Async._drainQueue (/usr/local/src/app/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/usr/local/src/app/node_modules/bluebird/js/release/async.js:143:10)
[...]

PR #617, opened 2017-07-12
fix job finished promise return value parsing when already complete

nkiraly:
Fixes Job.prototype.finished promise when job is already complete as determined by scripts.isFinished

returnvalue was already JSON.parse()'d during Job.prototype.fromJSON

this should address https://github.com/OptimalBits/bull/issues/614

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -327,7 +327,7 @@ Job.prototype.finished = function(){
         if(status == 2){
           throw new Error(job.failedReason);
         } else {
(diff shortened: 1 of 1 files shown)


nkiraly:
Upon further inspection, I see the additional JSON parsing is needed when reading job event data. Disregard.

manast:
A hack would be to try to parse, catch the error and return the unparsed returnvalue in that case. However, if we do that we will never know why this happens in the first place.

PR #622, opened 2017-07-19
job.finished: fix returnvalue already parsed and returnvalue in watchdog

saviogrossi:
more fixes in job.finished returns.


lib/job.js
```diff
@@ -7,7 +7,7 @@ var scripts = require('./scripts');
 var debuglog = require('debuglog')('bull');
 var errors = require('./errors');
 
-var FINISHED_WATCHDOG = 5000;
+var FINISHED_WATCHDOG = 1000;
```
manast: I guess the best would be to have the watchdog timeout as a configurable setting.

manast: (I can fix this later on)

lib/job.js
```diff
@@ -368,14 +368,14 @@ Job.prototype.finished = function(){
         //
         interval = setInterval(function(){
           scripts.isFinished(_this).then(function(status){
-            var finished = status > 0;
-            if(finished){
+            Job.fromId(_this.queue, _this.id).then(function(job){
+              removeListeners();
```
manast: I am thinking that maybe it is better to call removeListeners() on a .finally() call on ```scripts.isFinished```, and return Job.fromId so that the errors propagate correctly. A catch with a ```console.error()``` would also be nice :).

lib/job.js
```diff
@@ -368,14 +368,14 @@ Job.prototype.finished = function(){
...
               if(status == 2){
-                getFailed().then(resolve, reject);
+                reject(new Error(job.failedReason));
+              } else {
+                resolve(job.returnvalue); 
               }
-              removeListeners();
-              resolve();
-            }
+            });
```
manast: I do not think this code is correct. It is ignoring the status result from the call to isFinished, so it will resolve independently of if the job has been finished or not.

saviogrossi: Ok, I will work on this

saviogrossi: Sorry, I misunderstood isFinished, I thought it only resolve when job status is already finished. Will fix this, tks.

strax:
@saviogrossi @manast what's the status with this? I could submit a PR with only the return value fix if you're interested as we're already using a custom fork with that patch in.

manast:
@strax there are un-addressed comments yet...

saviogrossi:
sorry @manast <https://github.com/manast> @strax
<https://github.com/strax> , I
haven't much time to address Manuel comments. Sami, if you think you can
address, I will cancel this PR so you send yours. Tks

lib/job.js
amirrustam: Making this configurable would be helpful.

Issue #638, opened 2017-08-07
Commit b01f0419c breaks tests & repeatable jobs


ferronrsmith:
The following change https://github.com/OptimalBits/bull/commit/b01f0419cb19fe58246880aadb176a2eecbb8d8e broke a lot of functionality.

manast:
I know. I am fixing the issues as we speak.

ferronrsmith:
awesome thanks

Issue #637, opened 2017-08-07
Cron Jobs fail - only the initial job is executed


ferronrsmith:
```javascript
  it('should repeat every 2 seconds', function (done) {
    var _this = this;
    var date = new Date('2017-02-07 9:24:00');
    this.clock.tick(date.getTime());
    var nextTick = 2 * ONE_SECOND + 500;

    queue.add('repeat', {foo: 'bar'}, { repeat: {cron: '*/2 * * * * *'}}).then(function(){
      _this.clock.tick(nextTick);
    });

    queue.process('repeat', function(job, done){
      // dummy
      return done();
    });

    var prev;
    var counter = 0;
    queue.on('completed', function(job){
      _this.clock.tick(nextTick);
      if(prev){
        expect(prev.timestamp).to.be.lt(job.timestamp);
        expect(job.timestamp - prev.timestamp).to.be.gte(2000);
      }
      prev = job;
      counter ++;
      if(counter == 20){
        done();
      }
    });
  });
```

PR #641, opened 2017-08-09
Performance fixes

manast:
This PR includes various performance fixes. Some related to faster javascript, but the most important one is the one related to notifying waiting jobs, which now uses a blocking operation instead of pubsub. This solution completely removes the overhead of having many concurrent workers reacting to the pubsub 'added' event.

### DIFF
--- a/lib/commands/addJob-6.lua
+++ b/lib/commands/addJob-6.lua
@@ -12,16 +12,15 @@
          - LIFO
          - FIFO
          - prioritized.
-         - Emits a global event 'added' if not paused.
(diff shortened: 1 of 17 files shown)


manast:
@bradvogel the optimization is ready. Having many concurrent workers now should not affect the queue's performance as badly as it did before.

bradvogel:
Awesome! We're going to rerun the new benchmarks and post an update in https://github.com/bee-queue/bee-queue/issues/67

manast:
Also fixes #638

manast:
Issue #637 cannot be reproduced in this branch either.

PR #646, opened 2017-08-11
Update eslint dependency

udleinati:
Update eslint package.


manast:
thanks. Although nice with the formatting, it would be better to have a PR with only with the eslint package upgrade commit.

Issue #667, opened 2017-08-21
Add support for storing worker metadata


manast:
Currently only the name of the worker is stored. However, it is also quite useful to also store some extra metadata, such as the worker version number, etc.

marshall007:
Also the concurrency option it was passed. This would be useful since you don't necessarily know what the specified concurrency was when restarting queues.

manast:
And also store the queue token, this can be useful to know which worker is actually processing a given job (since the lock stores the worker's token)

PR #682, opened 2017-09-04
Added rate limiter support #362

manast:


marshall007:
@manast how does this relate to the concurrency option in `queue.process(...)`? Should that option be deprecated in favor of the new `limiter` option since it is persisted in the redis queue metadata?

manast:
@marshall007 not sure I understand what you mean. The limiter option is on the queue constructor, not on the process method, it does not affect the concurrent option. Also, the limiter is global, so if you configure it for 1 job per second, then it will not process more than 1 per second independently on the concurrency value or the amount of workers you have running.

marshall007:
@manast I get that the `concurrency` option is local to the worker, just wondering if it doesn't make sense to deprecate/merge that functionality into the new `limiter` option. We could have:

```ts
interface RateLimiter {
  max: number,      // Max number of concurrent jobs
  concurrency: number, // Max number of concurrent jobs within a single worker
  duration: number, // per duration in milliseconds
}
```

This would solve https://github.com/OptimalBits/bull/issues/667#issuecomment-324096544.

manast:
I agree that the concurrency option can be lifted to the queue constructor, but it is not related to the rate limiter as far as I understand it.

ritter:
If concurrency is lifted to the queue constructor, we lose the option of named processes having a specific concurrency. Is that correct?
I'd personally rather keep concurrency at the process method, because I have a single Queue with multiple named processes and I'd like to keep specifying different concurrencies depending on the process.

Please correct me if I'm wrong.

marshall007:
> I'd personally rather keep concurrency at the process method, because I have a single Queue with multiple named processes and I'd like to keep specifying different concurrencies depending on the process.

@ritter the problem with this approach is that it makes it difficult to restart dynamically created queues. If the `concurrency` option is not persisted in redis as queue metadata then you can't reliably restart queues (after process crash) with the appropriate concurrency.

marshall007:
> If concurrency is lifted to the queue constructor, we lose the option of named processes having a specific concurrency. Is that correct?

We could work around this by, optionally, allow you to specify key/value (name/concurrency) pairs in the constructor option. So specifying a number would be equivalent to `{ "__default__": <value> }`

manast:
@marshall007 The max parameter only states the maximum number of works processed by time unit. Whereas concurrency specifies how many processors can work concurrently on a given worker. Concurrency of 5 only means that jobs will be processed at most 5 in parallel, it does not state anything regarding how many jobs are allowed to be produced by unit of time... for example, a job could take 10 seconds to complete, other 20, other 5, other 1, other 3. All of them could at a given time be processed concurrently with a concurrency of 5, but how is that related to the rate limiter?

marshall007:
@manast ah, my mistake. I misunderstood the new options. I thought `max` specified the maximum number of concurrent jobs globally and `duration` was simply a minimum timeout between jobs (so if duration were 1000ms and a job took 800ms, it would wait an additional 200ms before freeing it from the processing queue).

Given that it's based on units of work / time, I agree that what I was saying doesn't make sense. Though, I still agree with what you initially said about somehow moving it to the constructor and storing it as metadata.

PR #686, opened 2017-09-04
Update job data

manast:


manast:
@bradvogel yes, that why there is an extra commit LOL :)

PR #689, opened 2017-09-11
implementation threaded processors

### LINKED ISSUE #488: [Feature] Support for running jobs in child processes

manast:
fixes #488

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,13 +79,13 @@ Are you developing bull sponsored by a company? Please, let us now!
 - [x] Concurrency.
 - [x] Pause/resume—globally or locally.
 - [x] Multiple job types per queue.
+- [x] Threaded (sandboxed) processing functions.
 - [x] Automatic recovery from process crashes.
 
 And coming up on the roadmap...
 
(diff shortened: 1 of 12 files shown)


robolivable:
Does this effectively solve the [second point](https://github.com/OptimalBits/bull/blame/develop/README.md#L428) in the "Important Notes" section of the README? I see that it references a ticket that this PR closes, so I'm curious if the README is slightly outdated as a result.

zyf0330:
Hello, is this feature released?

manast:
> Hello, is this feature released?

Yes, you can read about it in the documentation (Sandboxed processors)

PR #703, opened 2017-09-27
Fixed JobCounts interface description in REFERENCE docs

kroleg:
Number of waiting jobs contains in "waiting" property. There is no "wait" property.


manast:
good catch. Thanks!

PR #708, opened 2017-10-02
Fix duplicate return value parsing for already completed jobs

strax:
Calling `Job.finished()` on an already-completed job with a return value throws error, since the value is already parsed.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -351,7 +351,7 @@ Job.prototype.finished = function(){
         if(status == 2){
           throw new Error(job.failedReason);
         } else {
-          return job.returnvalue && JSON.parse(job.returnvalue);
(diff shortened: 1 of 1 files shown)


manast:
Is the title of this PR correct? seems like what the commit does is just to check that returnvalue is defined before trying to parse it...

strax:
@manast Quite the contrary, this PR removes duplicate parsing since `Job.finished` calls `Job.fromId` which in turn already parses the return value in https://github.com/optimalbits/bull/blob/master/lib/job.js#L493.

manast:
I merged this one https://github.com/OptimalBits/bull/pull/622/ which solve the same issue but included a few new unit tests. Thanks for your contribution anyway.

Issue #722, opened 2017-10-10
Ability to get a job's status


pintocarlos:
It would be great to have the ability to get a Job's status (completed, failed, delayed... etc).

We have a scenario where we would like to write custom logic depending on the job's status. The existing API to return all of the jobs that fall under a specific status isn't really fitting for our case scenario.

Is there any plans for adding this functionality?
Thanks!

manast:
No in the short term. That functionality does not fit so well in the current design, it would require an extra status variable that needs to be updated atomically when moving the job from state to state. Maybe if you explain what you are tying to achieve I can find a different path.

pintocarlos:
I actually found a `getStatus` method on the Job object that returns a Promise with the Job's status. This is exactly what we needed but it is not currently documented.

Now, after reading your reply, did you understand something different? 
Were you maybe referring to a string property on the object instead of a Promise-returning method? Just wanted to clarify.
Thanks.

manast:
No I did not, but we get a lot of patches and sometimes I forget that we already have some functions. However, this implementation is not very high performant, nor atomic, but as long as you queue does not have a very large amount of items I guess it is ok.

PR #744, opened 2017-10-18
Fixes #719

### LINKED ISSUE #719: Problem with concurency
When i have one queue, and i add named process to it, concurency is increased to the amount of the proccesors

For example:
const test = new Queue('test', {redis});

test.process('test1', require('./processors/test1));
test.process('test2', require('./processors/test2'));

I expect each proccessor run one at the time .. but it runs 2 at the time, if i add 3 proccesor - then it run 3 at the time.

The only work around i have found is to create seporate queue for each proccesor.

Is it supposed to work like that?

I use bull@3.3.1

weeco:
Even though one can expect that, it's clarified with a short sentence.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -111,7 +111,17 @@ __Warning:__ Do not override these advanced settings unless you understand the i
 ### Queue#process
 
 ```ts
-process(name?: string, concurrency?: number, processor: (job, done?) => Promise<any> | string)
+/**
+ * Consider these as overloaded functions. Since method overloading doesn't exist in javacript
+ * bull recognizes the desired function call by checking the parameters' types. Make sure you
+ * comply with one of the below defined patterns.
+ * 
+ * Note: Concurrency defaults to 1 if not specified.
+ */
+process(processor: (job, done?) => Promise<any> | string)
+process(concurrency: number, processor: (job, done?) => Promise<any> | string)
+process(name: string, processor: (job, done?) => Promise<any> | string)
+process(name: string, concurrency: number, processor: (job, done?) => Promise<any> | string)
 ```
 
 Defines a processing function for the jobs in a given Queue.
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -148,7 +148,27 @@ module.exports = function(job){
 You can return a value or a promise to signale that the job has been completed.
 
 
-A name argument can be provided so that multiple process functions can be defined per queue. A named process will only process jobs that matches the given name. If you define multiple named process functions in one Queue they will run concurrently (in case jobs are queued for these processors).
+A name argument can be provided so that multiple process functions can be defined per queue. A named process will only process jobs that matches the given name. If you define multiple named process functions in one Queue the defined concurrency for each process function stacks up for the Queue. See the following examples:
+```js
+  /***
+   * For each named processor concurrency stacks up, so any of these three process functions
+   * can run with a concurrency of 125. To avoid this behaviour you need to create an own queue
+   * for each process function.
+   */
+  const loadBalancerQueue = new Queue('loadbalancer')
+  loadBalancerQueue.process('requestProfile', 100, requestProfile)
+  loadBalancerQueue.process('sendEmail', 25, sendEmail)
+  loadBalancerQueue.process('sendInvitation', 0, sendInvite)
+
+  const profileQueue = new Queue('profile')
+  // Max concurrency for requestProfile is 100
+  profileQueue.process('requestProfile', 100, requestProfile)
+
+  const emailQueue = new Queue('email')
+  // Max concurrency for sendEmail is 25
+  emailQueue.process('sendEmail', 25, sendEmail)
+```
+
 
 **Note:** in order to determine whether job completion is signaled by
 returning a promise or calling the `done` callback, Bull looks at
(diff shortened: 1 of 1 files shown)


REFERENCE.md
```diff
@@ -138,7 +138,7 @@ module.exports = function(job){
 You can return a value or a promise to signale that the job has been completed.
 
 
-A name argument can be provided so that multiple process functions can be defined per queue. A named process will only process jobs that matches the given name.
+A name argument can be provided so that multiple process functions can be defined per queue. A named process will only process jobs that matches the given name. If you define multiple named process functions in one Queue they will run concurrently (in case jobs are queued for these processors).
```
manast: hmm, there is a risk for misunderstanding if we say "the will run concurrently" in the sense that they still be in the same queue and processed in a given order (FIFO by default). So for instance if we have a queue like this, with job type A and job type B, the job A will not be processed until all B jobs are processed, but the queue will have a concurrent value of 2:
```AAAAAAABBBBBB```, 

I think that even if this is by design maybe it is not so logical after all...

Comment (weeco) [reply REFERENCE.md]: Sorry I didn't understand your comment.

Do you mean that I should further clarify, that even though there may be multiple named process functions in the exact same queue (which may be set to FIFO or LIFO) it doesn't mean it's necessarly going to meet the FIFO/LIFO order? If so then I understood the concept correctly and I can further clarify it.

**Comment by weeco** [comment]
I probably should have created another pull request for that change, but I wasn't sure how to do so since I was working on the same file. What do you think about these changes to the process function documentation?

Personally I was confused about the possible call patterns. I think it's much clearer with that documentation.

File: REFERENCE.md (hunk 1 above)
Comment (manast) [reply REFERENCE.md]: I think that the current behaviour may not be very logical, but anyway it works like this: when you define a process function, you have an optional concurrency parameter, this means that  several processors can run concurrently when processing the queue. But when you define new processors for named jobs, then for every processor you increase the concurrency value, so if you had a concurrency value of 4 in the first processor and a concurrency value of 5 in the second processor now the queue can process 9 jobs concurrently, but it does not matter if they are 9 of the same job name or not.

Comment (weeco) [reply REFERENCE.md]: I see, indeed that's not very intuitive. Not even easy to explain :D.

Comment (weeco) [reply REFERENCE.md]: By the way would that also mean if someone wants to run 200 concurrent requests in total for all the named processors, he could simply setup ONE process function with a concurrency of 200 and setup the other processor functions with a concurrency of 0? Just for clarification

File: REFERENCE.md [hunk 2]
```diff
@@ -111,7 +111,17 @@ __Warning:__ Do not override these advanced settings unless you understand the i
...
+ * bull recognizes the desired function call by checking the parameters' types. Make sure you
+ * comply with one of the below defined patterns.
+ * 
+ * Note: Concurrency defaults to 1 if not specified.
+ */
+process(processor: (job, done?) => Promise<any> | string)
+process(concurrency: number, processor: (job, done?) => Promise<any> | string)
+process(name: string, processor: (job, done?) => Promise<any> | string)
+process(name: string, concurrency: number, processor: (job, done?) => Promise<any> | string)
 ```
```
manast: LGTM

weeco:
@manast I also took care of #722 (jobState). Once this pull request gets merged, I will promise to create a new branch for each fix so that I can create multiple pull requests the next time.

REFERENCE.md
Asp3ctus: Can you explain why is like that by design? is it posible to change? the whole point of named proccesors is lost - then it behaves like that.

manast: @Asp3ctus the behaviour can be changed but why is it so bad as it is now?

Asp3ctus: @manast well as you stated in your comments in this topic you agree this is no very logical.
i will explain my case - of couse their is 100 ways to do anythink - so it's just for example, of cause i can rewrite my logic to that is changed in docs' but personaly i will probably just end up not using bull for my current task - instead make my own much less feture complete implementation.

Let's say you have a queue named "video" - this queue has many task like:
* "record-finder" - find records in db which don't have local processed videos - lets say it selects 10 records then it compleates it creates new job to "video-downloader" - this must run 1 time - only if no other tasks in video "queue"
* "video-downloader" - downloads videos - this can run in 1-4 concurency - so it doesn't use all of the bandwith - then it compleates - it is call "video-converter"
* "video-converter" - this can run in 1-4 concurency (dependant on cpu cores - so the ffmpeg doesn't use all of the system resorces) - then it's done - "db-updater" job is created
* "db-updater" - updates db with converted files - this can have 1-10 concurency - depends on db poll connections
once the db-updater finished - it checks that video queue does not have any other jobs - no mater which named procceeor - is it's empty - first named task is called - if not it will check later
then i finished this case - and started to test it - it ended up eating all of the bandwith and staling my cpu - as it started more and more ffmpeg proccesses as it progressed -it took me a lot of time to find this "feature/bug" before you updated the docs - and i was very disapointed then it ended up beeing a feature - not a bug - as it does not fit in my religion :) 

and this is only explain one queue - i will have 5-6 of diferent types  - this is very neat to organaize it in queus and create named proccesors - which have it's on concurency setting

if i will go the way a new documentation is stated - i will end up having 20-30 queues instead of 5-6 - the code will be more complex - the administration will be more coplex - more connections to redis(but it's not the issue)

My point is that if each named proccessor will follow it's own concurency settings - it will give more power to bull - and probably make some other person happy - not just me :)

PR #749, opened 2017-10-22
Fix #739 properly

leontastic:
Failing test output:

```
leonli:bull (master) npm test

> bull@3.3.2 pretest /Users/leonli/Documents/Dev/packages/bull
> npm run lint


> bull@3.3.2 lint /Users/leonli/Documents/Dev/packages/bull
> eslint lib test


> bull@3.3.2 test /Users/leonli/Documents/Dev/packages/bull
> NODE_ENV=test mocha



  sandboxed process
Child process exited 36222 0 null
    1) should remove exited process


  0 passing (6s)
  1 failing

  1) sandboxed process should remove exited process:

      AssertionError: expected [ 'undefined' ] to have a length of 0 but got 1
      + expected - actual

      -1
      +0

      at test/test_sandboxed_process.js:111:65
  From previous event:
      at Queue.<anonymous> (test/test_sandboxed_process.js:110:28)
      at lib/queue.js:866:13
  From previous event:
      at handleCompleted (lib/queue.js:865:40)
  From previous event:
      at Queue.processJob (lib/queue.js:896:23)
  From previous event:
      at lib/queue.js:804:12
  From previous event:
      at lib/queue.js:802:43
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)



npm ERR! Test failed.  See above for more details.
```

PR #750, opened 2017-10-22
Failing test: Process after exit

leontastic:
This PR includes a test for processes being run sequentially after a job has been killed. It's failing right now because the same process thread is retained for queued jobs despite the previous iteration calling `process.exit()` within an async callback, causing subsequent jobs to die ungracefully and without warning. Possible solutions:
- Add a handler to restart jobs that exit in the middle of processing on a new thread
- Add a feature to kill jobs automatically upon completion, forcing a new thread to be retained every time

Depends on https://github.com/OptimalBits/bull/pull/749.

### DIFF
--- a/lib/process/child-pool.js
+++ b/lib/process/child-pool.js
@@ -5,59 +5,63 @@ var path = require('path');
 var Promise = require('bluebird');
 var _ = require('lodash');
 
-module.exports = function ChildPool() {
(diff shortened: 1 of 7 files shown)


leontastic:
@manast I think I ran into issues trying to test this change, but I couldn't find more time to work on this so I closed the PR. I will reopen if I return to this work.

PR #755, opened 2017-10-23
Supresses possible memory leak warnings

weeco:
Since many listeners can be attached when calling job.finished() Bull should set the max listeners to unlimited

Otherwise you will get possible memory leak warnings like these:

> (node:592) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 global:completed listeners added. Use emitter.setMaxListeners() to increase limit
> (node:592) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 global:failed listeners added. Use emitter.setMaxListeners() to increase limit

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -79,6 +79,7 @@ var MAX_TIMEOUT_MS = Math.pow(2, 31) - 1; // 32 bit signed
 // Queue(name: string, url?, opts?)
 var Queue = function Queue(name, url, opts){
   var _this = this;
+  _this.setMaxListeners(0);
   if(!(this instanceof Queue)){
     return new Queue(name, url, opts);
   }


manast:
I do not feel completely comfortable suppressing this warning, since it may mask real leaks that we could have in the future. As far as I know the only way to produce this warning is by calling finished multiple times, so the use would have the chance to suppress it manually in this case. Documentation for this use case may be enough...

ritter:
I agree with @manast, I don't think Bull should be masking what could be legitimate warnings. I've not had any of these warnings since setting a higher reasonable default myself in my project, and I have over a dozen queues.

```
const EventEmitter = require( "events" );
EventEmitter.defaultMaxListeners = 20;
```

weeco:
Since I "promisified" most of my enqueued jobs I get lots of warning and it's also dynamically. It completely spams the log with these warnings. Since they are expected when one works with job.finished() I think it's mandatory to surpress them by default in that case.

What ideas do you have to make it more specific for job.finished() users?

PR #797, opened 2017-11-21
Replicate LUA scripts effects

cyberdelia:
Replicate LUA scripts effects rather than using pure scripts, it speeds up replication and avoid potential nasty consistency problems and having to design Lua scripts to be deterministic.

You can read more about this here: https://redis.io/commands/eval#scripts-as-pure-functions

### DIFF
--- a/lib/commands/addJob-6.lua
+++ b/lib/commands/addJob-6.lua
@@ -34,6 +34,8 @@
       ARGV[10] LIFO
       ARGV[11] token
 ]]
+redis.replicate_commands()
+
 local jobId
 local jobIdKey
 local rcall = redis.call
--- a/lib/commands/extendLock-2.lua
+++ b/lib/commands/extendLock-2.lua
@@ -12,6 +12,8 @@
   Output:
     "1" if lock extented succesfully.
 ]]
+redis.replicate_commands()
+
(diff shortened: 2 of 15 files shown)


manast:
hmm, currently all LUA scripts in bull are pure functions, so what is the use of this PR?

cyberdelia:
It appears to not be the case, as I have a replica blocking on a `brpoplpush` because of a bull lua script having two different behaviour on two different server. 

It's hard to replicate unless you reach a critical mass of jobs, but still happen like clockwork.

manast:
If you write a non-pure lua script, redis will error, try for instance:
```lua
local time = redis.call('TIME')[0]
redis.call('SET', 'time', time)
```
So if there are non-pure functions we should get redis to complain which it doesn't. Means that the error must be related to something else. Furthermore, activating ```replicate_commands``` makes the lua execution much slower, so we do not want to do that. My proposal is to try to understand what causes you issue and address that specifically.

cyberdelia:
I didn't have much data at the time, but I found another case, where the replica crashes because of a `BRPOPLPUSH` issued by this library LUA scripts.

You can read more about the crash issue here: https://github.com/antirez/redis/issues/4440

PR #798, opened 2017-11-28
Adding ability to limit by stacktrace

joshuacullen:
If a job fails multiple times the stacktrace can grow very large. 
This is to limit the amount of memory used by a failed job. 

This enables a job to be created with a max stacktrace.


joshuacullen:
Yeah sure thing. I've added the option to the REFERENCE.md.

Issue #693, opened 2017-09-19
global:completed event handler receives only job id and not job object


fabiomargarido:
Hi there,

I'm using 'global:completed' event handlers and they were working on 3.0.0-rc versions. I've updated recently to 3.1.0 and have observed that instead of receiving the job object, my handlers started to receive only the job id in the 'job' parameter. If I change the handler to local 'completed' events, the full job object is received, as expected.

manast:
Yes, that is true. I have to update the documentation. The reason is that for returning the full job a ```Job.fromId``` is required, but since this is not always needed I did not want to have that extra performance penalty. So the solution is just to use ```Job.fromId```  in your handler.

fabiomargarido:
I've tried it but I always get 'null' when I do 'Job.fromId'. Is this because I'm using 'removeOnComplete: true'?

manast:
yes. The job is deleted after emitting the event.

fabiomargarido:
So my use case is no longer supported in 3.0.0. Isn't there a way to make the id vs object parameter configurable?

fabiomargarido:
Otherwise what are my options? To periodically clean completed jobs from the queue? I have no need to keep old jobs around, so that's why we use removeOnComplete.

manast:
yes it could be done. an alternative is to remove the job after getting it in your global:completed handler.

manast:
I can add this feature as en enhancement.

maxbates:
this bit me as well, +1 for documenting

cxcorp:
It seems like the result is passed via a second argument to the handler:

https://github.com/OptimalBits/bull/blob/7ad18f5f08cdca3746045d8c7b6c15f4ac06be7c/lib/queue.js#L387-L390

Can this be trusted to actually provide the result or can it be null? What about when `removeOnComplete` is `true`?

manast:
as long as the job returns a non-undefined result, the result should be defined independently of the ```removeOnComplete``` option.

PR #801, opened 2017-11-28
Mention global events only passing job ID in docs

cxcorp:
The documentation in its current state only briefly covers global events. As seen in #693, there's no mention of global event listeners only receiving the job ID instead of the job itself. This PR attempts to fill that gap in the reference.

Additionally, this PR also highlights their existence by adding a subsection for global events in the table of contents.


manast:
Great, Thanks!.

PR #817, opened 2017-12-22
Update Patterns Documentation to Clarify Redis Setup

peterpme:
:wave: awesome library!

I would just like to add a few clarifications to the documentation. This took me about 30 minutes to figure out. Not a long time, but I would hate to have somebody else stumble upon and go through what I did.

Two things:

- I tried using `redis` library, bull uses `ioredis`
- passing in a `REDIS_URL`. Doesn't have to happen here, but I would like to at least make a NOTE of it, especially since Heroku is mentioned elsewhere and I am in fact using Heroku.

Thanks 😄


manast:
thanks!.

PR #826, opened 2018-01-02
Support custom backoff strategies

holm:
This adds support for custom backoff strategies. 

We have a need to do a backoff that adds some jitter to the delay, so we don't get many retries at the exact same time. Instead of hard-coding another strategy, I came up with this solution instead.

I originally wantes to support a custom method directly on the job, but that didn't work as it has to be serialized to json. So I put it on the queue settings instead.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -109,7 +109,7 @@ __Warning:__ Do not override these advanced settings unless you understand the i
 
 `retryProcessDelay`: Time in milliseconds in which to wait before trying to process jobs, in case of a Redis error. Set to a lower value on an unstable Redis connection.
 
-`backoffStrategies`: An object containing custom backoff strategies. The key in the object is the name of the strategy and the value is a function that should return the delay in milliseconds. For a full example see [Patterns](./PATTERNS.md).
+`backoffStrategies`: An object containing custom backoff strategies. The key in the object is the name of the strategy and the value is a function that should return the delay in milliseconds. For a full example see [Patterns](./PATTERNS.md#custom-backoff-strategy).
 
 ```js
 backoffStrategies: {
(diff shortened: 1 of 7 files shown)


lib/job.js
```diff
@@ -190,10 +192,11 @@ Job.prototype.moveToFailed = function(err, ignoreLock){
     // Check if an automatic retry should be performed
     if(_this.attemptsMade < _this.opts.attempts && !_this._discarded){
       // Check if backoff is needed
-      var backoff = _this._getBackOff();
-      if(backoff){
+      if(_this.opts.backoff){
```
manast: in the previous implementation backoff could be 0, meaning that in this case it will be retried immediately. This differs from the new implementation where it is still sent to the delay set.

REFERENCE.md
```diff
@@ -89,6 +89,7 @@ interface AdvancedSettings {
   maxStalledCount: number = 1; // Max amount of times a stalled job will be re-processed.
   guardInterval: number = 5000; // Poll interval for delayed jobs and added jobs.s
   retryProcessDelay: number = 5000; // delay before processing next job in case of internal error.
+  backoffStrategies: {}; // A set of custom backoff strategies keyed by name. The value should be a function that when invoked, returns the delay in milliseconds.
```
manast: maybe an example would be useful since it may a bit difficult to understand that you have to define the strategy in the queue and then select it when you add the job to the queue.

lib/job.js
holm: Good spot. I changed it now to behave as before, but I am not sure how best to create a testcase for this.

REFERENCE.md
holm: I added more details here and also added a full example in PATTERNS.md.

PR #829, opened 2018-01-04
Fix reference interface name

vhf:
This PR corrects a typo in the reference docs, signature showed `QueueOptions` but the interface name was `QueueOpts`.


manast:
thanks!

PR #835, opened 2018-01-08
add es2015 module support.

### LINKED ISSUE #786: Queue.process("path/to/module") doesn't handle ES6 default export
```
/app/node_modules/bluebird/js/release/method.js:9
        throw new Promise.TypeError("expecting a function but got " + util.classString(fn));
              ^
TypeError: expecting a function but got [object Object]
    at Function.Promise.method (/app/node_modules/bluebird/js/release/method.js:9:15)
    at process.<anonymous> (/app/node_modules/bull/lib/process/master.js:35:29)
    at emitTwo (events.js:126:13)
    at process.emit (events.js:214:7)
    at emit (internal/child_process.js:772:12)
    at _combinedTickCallback (internal/process/next_tick.js:141:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

Problem is that typescript produces ES6 default export:
```
Object.defineProperty(exports, "__esModule", { value: true })
exports.default = processor;
```

So, Bull should not expect that `exports` will be `function`, but instead use `default` if defined.

Workaround: give some name to default function and
```
module.exports = processor
```

marshalys:
fix #786

### DIFF
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -29,6 +29,9 @@ process.on('message', function(msg) {
   switch(msg.cmd){
     case 'init':
       processor = require(msg.value);
+      if (processor.default) { // support es2015 module.
+        processor = processor.default;
+      }
       if(processor.length > 1){
         processor = Promise.promisify(processor);
       }else{


marshalys:
write code like this in es2015 modules:
```
 // processor.js
exports default function(job){
  // Do some heavy work

  return Promise.resolve(result);
}
```

manast:
great, thanks!.

marshalys:
You are welcome. 😄

PR #843, opened 2018-01-16
Quitting own Redis clients and removing listeners on others

Equals182:
I'm pretty sure it's not an ideal approach, but it should properly work with any kind of clients. I've tested it with `npm test` and live, it looks good.


Tin-Nguyen:
+1, I'm waiting for this as well

PR #852, opened 2018-02-04
 Added repeatable limit feature

alexkh13:
The problem: 
When dealing with clustered processors, it is hard to achieve an easy implementation to limit the number of times a repeatable job should run without diving into Bull's nextRepeatableJob method.

The solution:
An extended attribute on the RepeatOpts interface to indicate how many times the job should run at max. It will increment a counter on the RepeatOpts itself ("count") to help keep track of what is the limit and how many jobs were already initiated.

I've added a test case and updated the Queue.add method documentation accordingly.

Works like a charm on my clustered app :)


manast:
Looks great, thanks!.

PR #860, opened 2018-02-15
Add defaultJobOptions to Queue

### LINKED ISSUE #706: Change default job options
Is it possible to change the default behaviour/options for jobs? I would like to define that my jobs get removed on completion / fail by default instead of passing job options everywhere.

mxstbr:
I'd like to pass `removeOnComplete` and `removeOnFail` only once, instead of every time I run `.add()` on a queue. This patch adds a new option to queues called `defaultJobOptions` which, as the name suggests, sets the default job options.

Also adds two tests verifying that the option is passed through correctly and that it works with `removeOnComplete`. I wasn't sure whether to more tests were necessary, if so, please just let me know and I'll work on it.

Closes #706

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -59,6 +59,7 @@ var MAX_TIMEOUT_MS = Math.pow(2, 31) - 1; // 32 bit signed
     limiter?: RateLimiter,
     redis : RedisOpts, // ioredis defaults,
     createClient?: (type: enum('client', 'subscriber'), redisOpts?: RedisOpts) => redisClient,
+    defaultJobOptions?: JobOptions,
 
     // Advanced settings
     settings?: QueueSettings {
@@ -100,6 +101,10 @@ var Queue = function Queue(name, url, opts){
 
   if(opts.limiter){
     this.limiter = opts.limiter;
+  }
+
+  if (opts.defaultJobOptions) {
+    this.defaultJobOptions = opts.defaultJobOptions;
   }
 
   this.name = name;
@@ -621,6 +626,9 @@ Queue.prototype.add = function(name, data, opts){
     data = name;
     name = Job.DEFAULT_JOB_NAME;
   }
+  if (!opts) opts = {};
+  _.defaults(opts, this.defaultJobOptions);
+
   if(opts && opts.repeat){
     var _this = this;
     return this.isReady().then(function(){
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -626,10 +626,10 @@ Queue.prototype.add = function(name, data, opts){
     data = name;
     name = Job.DEFAULT_JOB_NAME;
   }
-  if (!opts) opts = {};
+  opts = _.cloneDeep(opts || {});
   _.defaults(opts, this.defaultJobOptions);
 
-  if(opts && opts.repeat){
+  if(opts.repeat){
     var _this = this;
     return this.isReady().then(function(){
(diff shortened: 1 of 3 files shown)


lib/queue.js
```diff
@@ -621,6 +626,9 @@ Queue.prototype.add = function(name, data, opts){
     data = name;
     name = Job.DEFAULT_JOB_NAME;
   }
+  if (!opts) opts = {};
```
manast: this should not be needed since we already take care of it here: https://github.com/mxstbr/bull/blob/81d7540ac663ff4dcab30a9807067bd7d58105cb/lib/queue.js#L96

lib/queue.js
```diff
@@ -621,6 +626,9 @@ Queue.prototype.add = function(name, data, opts){
...
+  _.defaults(opts, this.defaultJobOptions);
```
manast: why merge the job options into the standard opts object? besides it may create side effects since it is modifying inputs.

test/test_queue.js
```diff
@@ -332,6 +332,15 @@ describe('Queue', function () {
         });
       }, done);
     });
+
+    it('creates a queue with default job options', function (done) {
+      var defaultJobOptions = { removeOnComplete: true };
```
manast: where you are at it, why not test also ```removeOnFailed``` ?

lib/queue.js
mxstbr: This is in `Queue.prototype.add`, this is not the main queue options. `opts` is undefined here quite often since the tests (and users too probably) do `Queue.add(data)` without specifying any special options.

lib/queue.js
mxstbr: This is in `Queue.prototype.add`, this is not the main queue options. `opts` here refers to the options passed to `Queue.add(data, opts)`, not `new Queue(opts)`, so I'm merging the default job options into the passed job options.

test/test_queue.js
mxstbr: Sure, happy to add a test for that!

lib/queue.js
manast: ah ok, sorry, I did not see it right due to github removing all the lines between the queue constructor and this line.
But please change the line to:
```
opts = _.cloneDeep(opts || {});
```
to avoid sideeffects.
Also in line 632 we do not need to check for opts anymore.

mxstbr: No worries, I know how it goes :blush: Will clean that up, one second!

mxstbr: Done, good catch

mxstbr:
The test failure only happened in Node 6 and seems to be unrelated to this patch. Race condition maybe? Can you try restarting that CI run?

mxstbr:
Restarting the CI run with a force push, let's hope it passes this time.

mxstbr:
Now a different test failed in Node 6 :confused: Any clues why @manast?

screen shot 2018-02-15 at 14 18 08

screen shot 2018-02-15 at 14 18 17

manast:
Seems like a hazard, probably was there before since it does not have to do with your code.

mxstbr:
:100: Would appreciate a release with this change

mxstbr:
Ping @manast, a release would be awesome :blush:

manast:
I'll try this weekend!.

mxstbr:
Thank you! :pray:

PR #865, opened 2018-02-16
Allow cleaning paused jobs

gabegorelick:
The underlying cleanInJobsSet supports it. Or is that a legacy thing?

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1111,9 +1111,8 @@ Queue.prototype.toKey = function(queueType) {
  * grace period.
  *
  * @param {int} grace - The grace period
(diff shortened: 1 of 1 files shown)


lib/queue.js
```diff
@@ -1127,7 +1126,10 @@ Queue.prototype.clean = function(grace, type, limit) {
   }
 
   if (
-    _.indexOf(['completed', 'wait', 'active', 'delayed', 'failed'], type) === -1
+    _.indexOf(
+      ['completed', 'wait', 'active', 'paused', 'delayed', 'failed'],
```
Tin-Nguyen: is there any tests on this? we should update the current tests or add new one if the test doesn't exist

PR #877, opened 2018-02-27
 fix typescript define error of Queue#process in document

marshalys:
someone make same mistake: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23806
so I think we should make the document right.

sorry for my bad english 😄


manast:
Good catch, thanks.

marshalys:
😄 you are wellcome

PR #887, opened 2018-03-12
Document Queue#getJobs

### LINKED ISSUE #876: Easier way to get all jobs of all statuses
## Description

I'm looking to list all jobs of all statuses. The only way that I've found is to use Queue#getActive, Queue#getWaiting, etc. and then concatenate the return values of each of those calls. Is there a simpler way than this? If not, it would be really useful.

## Test code to reproduce

N/A

## Bull version

3.3.10

denizdogan:
Fixes #876


manast:
Looks good, thanks.

PR #894, opened 2018-03-20
Fixes #893: A promise was rejected with a non-error.

### LINKED ISSUE #893: Warning 'a promise was rejected with a non-error' for external queues in case of an error.
## Description

If you run a bull queue as external process you get a bluebird warning in case of an error about a promise that was rejected with a non-error.
This is due to the sandbox implementation that rejects simply with an error message as string.

I will create a pull request for this. Happy to hear if that one is OK.

## Bull version
3.3.10

jpschober:
Instead of returning the error message in sandbox.js we return an Error object. To have a clean error message in the queue later on we must use the message property of the err object (if present) as error value in master.js.

### DIFF
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -59,7 +59,7 @@ process.on('message', function(msg) {
           function(err) {
             process.send({
               cmd: 'failed',
-              value: err
+              value: err.message || err
             });
           }
         )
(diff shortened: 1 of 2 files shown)


manast:
Looks good, thanks.

Zolmeister:
@manast please cut v3.3.11

PR #908, opened 2018-04-05
Fix process retain when reusing a previous process

igncp:
I was reviewing this code and it seems that it was missing to add the child process id to the `retained` map. It all started while trying to add coverage for that if block.

I am not aware of many details, the previous implementation may be intended. If so, I could update this test to at least cover that condition (without the new line in `lib/process/child-pool.js`).

Hope it helps, cheers


manast:
@igncp thanks for the PR.
Currently, the use of the retain object is used when the pool is cleaned. So this may create leaks in scenarios where queues are created and closed.

Issue #915, opened 2018-04-15
Prioritise delayed jobs


alexkh13:
## Description
When adding delayed jobs with priority to the queue, the priority is not affecting its placement in the waiting list.

## Test code to reproduce
```
var normalPriority = [],
      highPriority = [];

highPriority.push(queue.add({p: 1}, {priority: 1, delay:2000}));

for(var i = 0; i < 4; i++){
   normalPriority.push(queue.add({p: 2}, {priority: 2}));
}

Promise.all(normalPriority, highPriority).then(function(){
   queue.process(function(job, jobDone){
      console.log(job.data.p);
      setTimeout(jobDone, 1000);
   });
});
```

The expected output of the above should be 2,2,1,2,2
But indeed we get 2,2,2,2,1

## Bull version
Latest

## Additional information
This is, of course, relevant to repeatable jobs as well

roggervalf:
supported since v3.4.0 https://github.com/OptimalBits/bull/blob/master/CHANGELOG.md#v340

Issue #881, opened 2018-03-02
queue is ignoring priority if job has delay option too.


assafdd:
## Description
Priority queues are working great. 
But if I add delay option to job, priority value seems to get ignored.

## Test code to reproduce
```js
const Queue = require('bull');
var priorityQueue = new Queue('priorityQueue', 'redis://127.0.0.1:6379');

priorityQueue.process(function(job){
    console.log(`processing job ${job.id} priority: ${job.data.p}`);
    return Promise.resolve();
});

priorityQueue.add({p: 2}, {priority:2 ,delay:10});
priorityQueue.add({p: 1}, {priority:1 ,delay:10});
```

## Bull version
3.3.9

manast:
This is actually a design flaw in bull. Currently the priority is only used when adding a job to the queue, not in any other case when a job goes back to the wait state, this is true for delay, but also to retries in the case of failure, etc. Since it is not completely trivial to fix this, the question is how important is to support this case?

assafdd:
Not too much. 
Yet, I think it's worth being documented.  I spent quite a lot of time trying to figure out why priorities are not working in my use case. 
If you feel fine with documenting this behaviour, I can send a PR on the REFERENCE.md#queueadd doc.

PR #916, opened 2018-04-15
Prioritize delayed jobs

alexkh13:
#915 #881

### DIFF
--- a/lib/commands/updateDelaySet-4.lua
+++ b/lib/commands/updateDelaySet-4.lua
@@ -6,6 +6,7 @@
       KEYS[1] 'delayed'
       KEYS[2] 'active'
       KEYS[3] 'wait'
+      KEYS[4] 'priority'
 
       ARGV[1]  queue.toKey('')
       ARGV[2]  delayed timestamp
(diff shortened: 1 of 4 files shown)


lib/commands/updateDelaySet-4.lua
```diff
@@ -21,7 +24,25 @@ if (score ~= nil) then
   if (math.floor(score) <= tonumber(ARGV[2])) then
     redis.call("ZREM", KEYS[1], jobId)
     redis.call("LREM", KEYS[2], 0, jobId)
-    redis.call("LPUSH", KEYS[3], jobId)
+
+    local priority = tonumber(redis.call("HGET", ARGV[1] .. jobId, "priority"))
+
```
manast: I am a bit concerned about what happens if "priority" is not defined, for example in the case that you have a queue already and you upgrade bull, so that some old jobs where added without the priority field. Will ```tonumber``` coerce to 0?

manast: ok, I read the documentation of tonumber :) and it seems that it will return ```nil``` if the number could not be converted, so the if condition should be changed to:
```lua
if priority ~= nil or priority == 0 then
```

lib/commands/updateDelaySet-4.lua
```diff
@@ -6,13 +6,16 @@
       KEYS[1] 'delayed'
       KEYS[2] 'active'
       KEYS[3] 'wait'
+      KEYS[4] 'priority'
 
       ARGV[1]  queue.toKey('')
       ARGV[2]  delayed timestamp
+      ARGV[3]  token
```
manast: ```ARGV[3]``` is not used

lib/scripts.js
```diff
@@ -314,10 +314,10 @@ var scripts = {
     It checks if the job in the top of the delay set should be moved back to the
     top of the  wait queue (so that it will be processed as soon as possible)
   */
-  updateDelaySet: function(queue, delayedTimestamp) {
-    var keys = [queue.keys.delayed, queue.keys.active, queue.keys.wait];
 
-    var args = [queue.toKey(''), delayedTimestamp];
+  updateDelaySet: function(queue, delayedTimestamp){
+    var keys = [queue.keys.delayed, queue.keys.active, queue.keys.wait, queue.keys.priority];
+    var args = [queue.toKey(''), delayedTimestamp, queue.token];
```
manast: ```queue.token``` does not seem to be needed.

alexkh13: You're right. The token is useless there. I thought I'll emit the same waiting event like when adding a job, but then I realised it's not a good idea right now.

So I removed the corresponding line but forgot to remove the arguments.

lib/commands/updateDelaySet-4.lua
alexkh13: i think you meant `priority == nil`
nevertheless, i moved it to the variable assignment `local priority = tonumber(...) or 0`
i think it is more clean that way

alexkh13:
i've added the unit test inside test_repeat to exploit the fake timers it already has.

lib/commands/updateDelaySet-4.lua
```diff
@@ -21,7 +22,27 @@ if (score ~= nil) then
   if (math.floor(score) <= tonumber(ARGV[2])) then
     redis.call("ZREM", KEYS[1], jobId)
     redis.call("LREM", KEYS[2], 0, jobId)
-    redis.call("LPUSH", KEYS[3], jobId)
+
+    local priority = tonumber(redis.call("HGET", ARGV[1] .. jobId, "priority")) or 0
```
manast: I am not an expert in LUA, but doesn't ```or``` evaluate to a boolean?

manast: ah ok, seems correct: https://www.lua.org/pil/3.3.html

manast:
I forgot to mention, you need to run prettier before commit in order to pass the unit tests.

Issue #911, opened 2018-04-10
Seperate process worker fails to launch when Node is started with --inspect flag


nirmalvp:
## Description
When the node process is started with the --inspect flag, bull fails to start a child process for the process worker with the error message : " error: unknown option `--inspect' "

## Test code to reproduce
nodemon index.js  --inspect  --exec babel-node
## Bull version
3.3.7

ryedin:
Confirmed this behavior.

It looks to me like you could blacklist the `--inspect` option (ie. remove it when the child processes are spawned). I don't think `--inspect` is "cluster friendly" anyway. It should probably be shed from the command that spawns the children

manast:
Could we close this now that we have a workaround in 3.4.2 https://github.com/OptimalBits/bull/blob/master/CHANGELOG.md#v342 ?

PR #929, opened 2018-05-08
when --inspect is present, do not pass it on to child processes

ryedin:
Workaround for #911 

This still isn't _ideal_ because it means the debugger won't be attached to the children processes, but at least it makes it so that bull actually works when using `--inspect` in the main process.

The issue was simply that it was trying to bind to the same host:port as the main process. A robust solution might be some kind of auto-port incrementing algorithm for the children, but that seems like it can become a rather complicated bit of logic.

This at least gets us moving forward in a way that still allows us debug our main app and have the job processors pick up their work.

### DIFF
--- a/lib/process/child-pool.js
+++ b/lib/process/child-pool.js
@@ -20,7 +20,15 @@ module.exports = function ChildPool() {
       return child;
     }
 
-    child = fork(path.join(__dirname, './master.js'));
+    // if node process is running with --inspect, don't include that option
+    // when spawning the children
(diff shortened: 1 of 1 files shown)


ryedin:
Your builds seem to failing on this:

`The command "npm run prettier -- --list-different" exited with 1.`

:(

ryedin:
So... maintainer(s)... thoughts on this PR?

I've published this version to our private npm registry because we need to use it immediately. Just hoping you can merge it in so we can get back on an upgrade path.

Issue #942, opened 2018-05-15
Support more interval definitions for repeatable jobs


cuchi:
Hi!

We are currently migrating from [agenda](https://github.com/agenda/agenda), but there is this little problem keeping us from doing that.
The **cron** notation is a very good idea to schedule repeatable jobs, but there are cases it is impossible to use, like representing an interval definition of _every 90 minutes_. 

To support this, `RepeatOpts` could be modified to accept other formats besides **cron**, which could be:
- How many seconds or millis, like in `setInterval`
- A string to be parsed by [human-interval](https://github.com/agenda/human-interval), which is used by agenda

I do not know the complexity of this enhancement, but it could be a good idea for the next feature release, what do you think?

Maybe related to #912

manast:
It is quite easy to implement, so we can do this for the next release.

manast:
I think we already cover this functionality so I will close the issue, re-open if necessary.

PR #944, opened 2018-05-16
Support for milliseconds intervals

cuchi:
With this, the interval can be defined using a more flexible manner than **cron**, which was discussed in #942

For example, when I need a job to execute every 90 minutes, I would do like that:
```javascript
queue.add({ ... }, { repeat: { seconds: 5400 } })
```

I am not sure if it is the best API design for this, maybe the contributors have a better opinion on that. I need to update the reference and the TS definitions too.

We at @rung-tools will be testing this branch with our application to see if it is stable enough, I'll report here if something is not right.

### DIFF
--- a/lib/repeatable.js
+++ b/lib/repeatable.js
@@ -27,7 +27,7 @@ module.exports = function(Queue) {
     var now = Date.now();
     now = prevMillis < now ? now : prevMillis;
 
-    var nextMillis = getNextMillis(now, repeat.cron, repeat);
+    var nextMillis = getNextMillis(now, repeat);
     if (nextMillis) {
       var jobId = repeat.jobId ? repeat.jobId + ':' : ':';
       var repeatJobKey = getRepeatKey(name, repeat, jobId);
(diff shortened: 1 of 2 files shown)


lib/repeatable.js
```diff
@@ -127,13 +127,23 @@ module.exports = function(Queue) {
...
+      : String(repeat.milliseconds || repeat.seconds * 1000);
+
+    return name + ':' + jobId + endDate + suffix;
   }
 
-  function getNextMillis(millis, cron, opts) {
+  function getNextMillis(millis, opts) {
+    var millisInterval = opts.milliseconds || opts.seconds * 1000;
+    if (millisInterval) {
+      return new Date(millis + millisInterval);
```
manast: Can't we just return ```millis+millisInterval``` ?

lib/repeatable.js
```diff
@@ -27,7 +27,7 @@ module.exports = function(Queue) {
     var now = Date.now();
     now = prevMillis < now ? now : prevMillis;
 
-    var nextMillis = getNextMillis(now, repeat.cron, repeat);
+    var nextMillis = getNextMillis(now, repeat);
```
manast: maybe an exception should be thrown if the user uses both the milliseconds and the cron expression at the same time.

manast:
Great job!, thanks for the contribution. Regarding the api, another possibility could be to have an option called ```every``` and then only support milliseconds, or allowing a postfix s for seconds, ms for ms, etc.

manast:
@cuchi bump!

test/test_repeat.js
```diff
@@ -14,7 +14,7 @@ var ONE_HOUR = 60 * ONE_MINUTE;
 var ONE_DAY = 24 * ONE_HOUR;
 var MAX_INT = 2147483647;
 
-describe('repeat', function() {
+describe.only('repeat', function() {
```
manast: remove .only

cuchi: I did it as soon as I realized in [8ae7a58](https://github.com/OptimalBits/bull/pull/944/commits/8ae7a589fff078e8f77c3e06e378cca21a5e9444).

PR #960, opened 2018-05-26
Update types array to include paused

brendann993:
getJobsCounts without arguments didn't return paused jobs as 'paused' was not included in types array


manast:
good catch, thanks.

Issue #966, opened 2018-06-05
Race conditions reviving repeatable jobs


alanclarke:
If you run `removeRepeatables`, and the remove happens just after a moveToX function has finished, the job will get zadded back into the queue.

I think this is an example of https://github.com/OptimalBits/bull/issues/904

alanclarke:
```js
it.only('should not re-add a repeatable job after it has been deleted', function() {
    var _this = this;
    var date = new Date('2017-02-07 9:24:00');
    var nextTick = 2 * ONE_SECOND;
    var repeat = { cron: '*/2 * * * * *' };
    var nextRepeatableJob = queue.nextRepeatableJob;
    this.clock.tick(date.getTime());

    var afterRemoved = new Promise(function(resolve) {
      queue.process(function() {
        queue.nextRepeatableJob = function () {
          var args = arguments;
          // In order to simulate race condition
          // Make removeRepeatables happen any time after a moveToX is called
          return queue.removeRepeatable(_.defaults({ jobId: 'xxxx' }, repeat))
            .then(function() {
              // nextRepeatableJob will now re-add the removed repeatable
              return nextRepeatableJob.apply(queue, args);
            })
            .then(function(result) {
              resolve();
              return result;
            })
        }
      });

      queue
        .add({ foo: 'bar' }, { repeat: repeat, jobId: 'xxxx' })
        .then(function() {
          _this.clock.tick(nextTick);
        })

      queue.on('completed', function() {
        _this.clock.tick(nextTick);
      })
    })

    return afterRemoved.then(function() {
      return new Promise(function(resolve) {
        queue.on('completed', resolve);
        _this.clock.tick(nextTick);
      })
      .then(function() {
        return queue.getRepeatableJobs().then(function(jobs) {
          // No repeatable jobs even though removeRepeatable was called
          expect(jobs.length).to.eql(0);
        })
      })
    })
  });
```

alanclarke:
PR here: https://github.com/OptimalBits/bull/pull/967

PR #967, opened 2018-06-05
Check if a repeatable was deleted before re-adding it

alanclarke:
Fixes https://github.com/OptimalBits/bull/issues/966

### DIFF
--- a/lib/repeatable.js
+++ b/lib/repeatable.js
@@ -8,7 +8,12 @@ var crypto = require('crypto');
 var Job = require('./job');
 
 module.exports = function(Queue) {
-  Queue.prototype.nextRepeatableJob = function(name, data, opts) {
+  Queue.prototype.nextRepeatableJob = function(
+    name,
+    data,
+    opts,
+    skipCheckExists
+  ) {
     var _this = this;
     var client = this.client;
     var repeat = opts.repeat;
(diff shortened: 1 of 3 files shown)


lib/repeatable.js
```diff
@@ -34,33 +39,44 @@ module.exports = function(Queue) {
...
-            data,
-            _.extend(_.clone(opts), {
-              jobId: customId,
-              delay: delay < 0 ? 0 : delay,
-              timestamp: now,
-              prevMillis: nextMillis
-            })
-          );
+        .zscore(_this.keys.repeat, repeatJobKey)
+        .then(function handleZScore(repeatableExists) {
```
manast: although better than before, there is still a hazard here, since the job could have been deleted after the call to zscore and before the call to create a new job. The only way to fix this correctly is using a LUA script... thats much more job, I think we should at least comment this fact so that it can be fixed in the future.

alanclarke:
Hey @manast thanks so much! I'm not sure why the build is failing actually, tests, coverage and prettier seem to be passing, note that the current commit on master is also failing. The error seems to be something to do with the process hanging? Maybe an unclosed connection somewhere? Any help much appreciated

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

alanclarke:
@manast thanks for merging, any chance of a release?

PR #982, opened 2018-06-20
Add events 'completed' and 'failed' to a Job

scinos:
This PR makes a Job an EventEmitter, so it can trigger events too. I implemented two events, 'completed' and 'failed'. These events are triggered *after* the queue's events, because there is little value in doing it before: in that case, you could as well run the code handling the event in the Job processor.

Our use case:

We have a WebServer that creates jobs, and a Worker that picks those jobs, generates the output, and then has to do some expensive cleanup before accepting the next job. We can control it with `pause`/`resume`, but it is hard to schedule the cleanup and not interfere with the "main loop". For example, if we use `setTimeout` inside the processor to schedule the clean up, it might end up running before the job is marked as done in Redis, delaying the whole process.

Having a event we can listen to that signals "the job has been processed and handled by the queue" is a great solution for us.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -190,7 +190,7 @@ const emailQueue = new Queue('email')
 emailQueue.process('sendEmail', 25, sendEmail)
 ```
 
-Specifying `*` as the process name will make it the default processor for all named jobs.  
+Specifying `*` as the process name will make it the default processor for all named jobs.
 It is frequently used to process all named jobs from one process function:
 ```js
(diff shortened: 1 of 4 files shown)


manast:
Thanks for the addition. The only thing that needs to be considered (and was the main reason we do not support jobs events yet), is the fact that only the job instance that actually moves the job to completed or failed will emit the event, another job instance not related to the job processing will not emit any event, and users will report it as a bug...

samuelgoldenbaum:
Any progress on job events - really critical for our scenario

PR #992, opened 2018-07-09
Update PATTERNS.md

matthiasfeist:
Updated documentation so the attemptsMade parameter is documented for the custom backoff strategies.


manast-apsis:
good catch, thanks!

Issue #994, opened 2018-07-11
Question: queue.getJobs() race condition


fyodorvi:
From `getJobs()` implementation I can see that unlike `getJobCounts()` you don't use transaction to get the jobs, which is potentially very dangerous since you might have jobs transitioning state while you are getting every state separately:

```
queue.getJobs(['active', 'waiting']);
```
Job 1 is in waiting state.
getJobs gets 'active' jobs => [] (empty array)
Job 1 transitions to active state
getJobs gets 'waiting' jobs => [] (empty array)

Could you please tell me if I'm wrong, or clarify why it's done without a transaction or tell me it's a bug.

Thanks.

manast:
More than a bug it has been overlooked, since the getters have not been considered very as critical methods.

fyodorvi:
@manast PR is out

PR #995, opened 2018-07-12
Improve queue.getJobs getter to use multi to avoid concurrency issues

fyodorvi:
Solves #994


fyodorvi:
Not sure how to unit test it... At least existing tests are passing.

manast:
Looks good, thanks.

Issue #980, opened 2018-06-19
Ability to backoff with an exact duration


moltar:
# Feature Request

## Use Case

I am legally scraping an API that is throttled. Once I reach the limit, API announces via HTTP headers when the usage resets. I'd like to be able to delay the job until that time passes instead of some arbitrary number of seconds.

## Solution Discussion

Not sure how the API would look like, since there is only the rejection triggers the repeat. I was thinking maybe a special instance of rejection error that could then be passed to the backoff strategy.

Or even just pass any error to the backoff strategy.

moltar:
What if we pass `err` as 4th argument to `backoffs.calculate(`?

https://github.com/OptimalBits/bull/blob/5144962c13670b36918268151fb459397c15763d/lib/job.js#L214-L218

And then pass it to `strategy(attemptsMade, err);`?

https://github.com/OptimalBits/bull/blob/5144962c13670b36918268151fb459397c15763d/lib/backoffs.js#L47

PR #1001, opened 2018-07-18
Feature/provide error to custom backoff strategy

fyodorvi:
Pushes the error that job failed with to the backoff strategy.

Should solve #980


manast:
Fantastic!, thanks.

moltar:
Thanks!

Issue #764, opened 2017-10-27
Job which obey delay and cron both


sudhirbitsgoa:
Hi,

```javascript
var Queue = require('bull');
var cronDelay = new Queue('cronDelay');
cronDelay.add({
  everysec: 'every3sec'
}, {
  delay: 10000,
  repeat: {
    cron: '*/3 * * * * *'
  }
}).catch(err => console.log(err))

cronDelay.process(5, (job, done) => {
  console.log('%j', job.data);
  done()
});

var i = 0;
setInterval(() => {
  console.log(i++);
}, 1000)
```
I created a job with delay of 10sec and repeat after 10sec for every 3 sec.
This is not supported.

nirpeer:
@manast Will it be supported in the future?

PR #1061, opened 2018-09-25
Add ability for cron repeatable job with startDate

nirpeer:
In this following example, the repeatable job will start run at 02.02.2020, 12:00, for every 2 minutes.
```
queue
      .add(
        'repeat',
        { foo: 'bar' },
        {
          repeat: {
            cron: '*/2 * * * * *', // every 2 seconds
            startDate: new Date('2020-02-02 12:00:00')
          }
        }
      )
```

This solution handles https://github.com/OptimalBits/bull/issues/764 issue.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -272,8 +272,8 @@ interface JobOpts{
 interface RepeatOpts{
   cron?: string; // Cron string
   tz?: string, // Timezone
-  startDate?: Date | string | number; // Start data when the repeat job should start repeating (only with cron).
-  endDate?: Date | string | number; // End data when the repeat job should stop repeating.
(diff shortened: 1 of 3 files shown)


REFERENCE.md
```diff
@@ -272,6 +272,7 @@ interface JobOpts{
 interface RepeatOpts{
   cron?: string; // Cron string
   tz?: string, // Timezone
+  startDate?: Date | string | number; // Start data when the repeat job should start repeating (only with cron).
```
manast: while we are at it we could correct the typo "data" -> "date"

manast:
Great addition, thanks!

REFERENCE.md
nirpeer: Sure

nirpeer:
@manast done :)

PR #1062, opened 2018-09-25
Propagate stack traces from child processes

### LINKED ISSUE #985: [BUG] Failed jobs don't have original stacktrace when run in separate processes
## Description
This is a regression introduced here https://github.com/OptimalBits/bull/commit/7160c06b96ae0d60b6f5358c94f90a9bd759ad00
(in 3.4.2), it results in loss of the original stack trace. Here is an example I grabbed from Arena UI for one of our failed jobs:

```js
Error: [object Object]
    at ChildProcess.handler ([....masked....]/node_modules/bull/lib/process/sandbox.js:25:22)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at emit (internal/child_process.js:772:12)
    at _combinedTickCallback (internal/process/next_tick.js:141:11)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

and before 3.4.2
```js
Error: email sending failed  
    at [....masked....]/server/app/email/job/send-email.js:9:8
    at Object.statsAsync ([....masked....]/server/lib/hermes/create-hermes-with.js:75:10)
    at statsAsync.next (<anonymous>)
    at statsAsync.tryCatcher ([....masked....]/node_modules/bluebird/js/release/util.js:11:23)
    at PromiseSpawn._promiseFulfilled ([....masked....]/node_modules/bluebird/js/release/generators.js:90:49)
    at Object.statsAsync ([....masked....]/node_modules/bluebird/js/release/generators.js:182:15)
    at P.method ([....masked....]/server/proc/lib/create-job-processor/index.js:121:4)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

I tried to understand the original motivation and I sounds like that change was made so promise is not getting rejected with just a message and not error object. But I am still not sure why do we need to do

```diff
           process.send({
               cmd: 'failed',
-              value: err
+              value: err.message || err
             });
```

Use separate process to process jobs with `process('*', 'path-to-file.js')`. Queue job that throws error, observe that logged error lacks original stack-trace. (Easy to see using Arena UI)

## Bull version
3.4.2

## Additional information

@manast @jpschober

nightpool:
This is a followup to #894 that fixes #985 (prematurely closed because the reporter fixed the issue on a fork).

the rationale here:

the subprocess will always bubble up a JSON-serialized representation of an Error object, and we use err.message to maintain the same behavior from #894. Because we have no way of knowing what specific type this error was or even guarantee that we'd have access to that constructor (it may be a library or function that lives only in the child process), we need to create a normal Error object. However, we can use the properties that are bubbled up, including stack, message, code, etc, so we Object.assign them.

### DIFF
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -57,9 +57,12 @@ process.on('message', function(msg) {
             });
           },
           function(err) {
+            if (!err.message) {
+              err = new Error(err);
+            }
             process.send({
               cmd: 'failed',
-              value: err.message || err
+              value: err
             });
           }
         )
--- a/lib/process/sandbox.js
+++ b/lib/process/sandbox.js
@@ -20,7 +20,9 @@ module.exports = function(processFile, childPool) {
             case 'failed':
             case 'error':
               child.removeListener('message', handler);
-              reject(new Error(msg.value));
+              var err = new Error();
+              Object.assign(err, msg.value);
+              reject(err);
               break;
             case 'progress':
               job.progress(msg.value);
--- a/test/test_sandboxed_process.js
+++ b/test/test_sandboxed_process.js
@@ -230,6 +230,7 @@ describe('sandboxed process', function() {
         expect(job.data).eql({ foo: 'bar' });
         expect(job.failedReason).eql('Manually failed processor');
         expect(err.message).eql('Manually failed processor');
+        expect(err.stack).include('fixture_processor_fail.js');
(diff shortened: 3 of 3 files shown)


nightpool:
test failures seem unrelated?

manast:
Yes, unrelated, I will need to fix the test in master before we merge this.

gcox:
@manast Those tests are fixed by #1094

Issue #892, opened 2018-03-19
Process only 1 job using queue.process


juan-cb:
Hi

Rather than an issue, this is a question about Bull. I'm coding a queues system where I don't need to process the queue in a continuous way (like it does `queue.process()`) but I need retrieve only one job each time I want (instead of processing every job available). 

I tried to do this using `process` by this way:
```
const consumeJob = function(queue) {
  return new Promise(
    function(resolve, reject){
          queue.isReady().then(() =>{
            queue.process(function(job){ 
              queue.close();
              resolve(job);
            });
          })
          .catch(function(error){
            reject(error);
          });
    }
  );
};
```

In this case, using `queue.close()` I destroy the Redis connection, so if I want to retrieve other job (or continue operating using Bull) without reinitiating the script (or Node server) I get the error `Error: Connection is closed`. Knowing that, I would like to ask:
  * If there is another way of obtaining only 1 (or N) job(s) using the `queue.process` method.
  * Is it possible to reconnect Redis after the `queue.close()` to continue using bull in other functions and parts of my code?

Thanks in advance

manast:
this may be a duplicate of this issue: https://github.com/OptimalBits/bull/issues/519

Issue #519, opened 2017-04-27
Non-persistent listener for processing?


doprdele:
I am wondering if there is a way to do the following in Bull. I have two queues, of which I need information from both queues.

I want to pop a job from Queue A, and then somewhere in processing, pop a job from Queue B, but, only once. I only want a job dequeued from Queue B inside the logic of Queue A's processing, kind of like this in a perfect world.

qA.process(function(item) {
  qB.processOne(function(item) {
 }
}

manast:
If we had a process function that only can be used once I guess your example would work, but it is a bit unorthodox usage of the queue. Can you explain more about the context where you want to do this? maybe there are other ways that matches bull better.

doprdele:
Sure.

Basically, I have a primary queue that has jobs and the qA.process() function works just fine for that. These jobs are added to the queue by another process, that works fine.

During the processing function, I need an element from a second queue. The reason I made this as a queue is because these second elements are time sensitive, meaning that I only want to grab them when I absolutely need them and I use a bull function to selectively remove all non-active keys that have passed the mark of 1 and a half minutes. The element of the second queue is necessary for the function to proceed and I have another process generating elements for this second queue. Unfortunately, I can't simply run both simultaneously and create an event when both are present to proceed with the work, it needs to be as fast as possible

I can do this with simple redis blpop but I wanted the extra security of using bull since it has a lot of built in acking, persistence, etc.

manast:
ok. I have never had the need, so I do not know if it makes sense that bull includes a processOnce method that just processes one job. But it is certainly possible to implement, basically its a matter of changing the run function so that instead of calling ```processJobs```, it just calls ```getNextJob and then ```processJob```. All this methods returns promises, so your ```processOnce``` will return a promise that you can use in your main process handler to wait for this second queue to finalize. To make it as stable as the current ```process```you will also need to take care of stalled jobs, calling ```movetUnlockedJobsToWait``` appropriately. I can look more into this, but I have a lot of high priority issues that needs to be implemented before 3.0

doprdele:
That's fine, it's not an urgent request, but it would be great. In the meantime I can use redis's brpop. The second queue that requires the processOnce functionality can be unstable, as in, I don't require stalled jobs to be reaped and placed back in the queue since each item only has a valid lifetime of 2 minutes.

manast:
btw, what happen if your main process function fails after processOne has finished? wouldn't you end with a dangling result somewhere?

[...]

PR #1066, opened 2018-09-27
Add processOne method to the library

cbjuan:
As discussed in #519, #892 and across different commits, the Bull library lacks a method to process the queue consuming individual jobs instead of doing the same retrieving all the jobs available sequentially. 

This PRs proposes a `processOne` method to do this task, allowing users to retrieve one job from the queue in each call for the function. To implement this, we have continued some methods yet available in Bull as described in https://github.com/OptimalBits/bull/issues/519#issuecomment-421407186, https://github.com/OptimalBits/bull/issues/519#issuecomment-422060733, and https://github.com/OptimalBits/bull/issues/519#issuecomment-422790555

@pacomf, @jelcaf and me co-author this PR.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -186,7 +186,7 @@ Job.prototype.releaseLock = function() {
  * @param ignoreLock {boolean} True when wanting to ignore the redis lock on this job.
  * @returns {Promise} Returns the jobData of the next job in the waiting queue.
  */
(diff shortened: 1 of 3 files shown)


cbjuan:
Closing this since with the last versions most of the contributions are already in the library.

moyarich:
what is the right way to implement this for my application? I would like to be able to process jobs one at a time instead of all at once

PR #1073, opened 2018-10-03
Improved docs for guide

j05u3:
I lost some time finding out why the async/await example wasn't working in my setup (the job was getting stalled because I was never calling the done function).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -71,16 +71,15 @@ that defines a process function like so:
 ```js
 const myFirstQueue = new Bull('my-first-queue');
 
-myFirstQueue.process(async (job, data) => {
-  return doSomething(data);
+myFirstQueue.process((job, done) => {
+  done(null, doSomething(data));
 });
 ```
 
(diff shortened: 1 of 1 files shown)


ritter:
@j05u3 I believe if you don't specify a 2nd parameter to the process function, bull expects it to be a promise and you should just return a promise.  No need to define and call done.

j05u3:
> @j05u3 I believe if you don't specify a 2nd parameter to the process function, bull expects it to be a promise and you should just return a promise. No need to define and call done.

It is true, but in the example there was a second parameter and it is not clear that it is a function you should call (the done function) and its name is misleading :) 

`myFirstQueue.process(async (job, data) => {
  return doSomething(data);
});`

ritter:
> `myFirstQueue.process(async (job, data) => { return doSomething(data); });`

I don't think `data` is correct as the 2nd parameter and should be:
`myFirstQueue.process(async (job) => { return doSomething(job.data); });`
Where `doSomething()` returns a promise.

manast:
Thanks for the improvement in the docs. I will try to improve it even a bit more.

PR #1074, opened 2018-10-03
Provide the Job to the custom backoff strategy

ronnievdv:
This request adds the job as an additional parameter to the custom backoff strategy.

Use Case:
We are using the Gmail API to batch process archives / deletes etc.  The API will respond with a status for each request in the batch, some of which may have failed while others succeeded.  We only want to retry the failed items, not the entire batch.  Adding the job as a parameter to the backoff strategy allows us to do this, by updating the job data with the failed ids (as per the simplified example in PATTERNS.md of this pull request)

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -130,7 +130,7 @@ NODE_DEBUG=bull node ./your-script.js
 Custom backoff strategy
 -----------------------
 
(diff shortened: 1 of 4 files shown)


manast:
Closing since this functionality has been added recently.

PR #1075, opened 2018-10-03
Fix for getRepeatableJobs when job type is 'every'

j05u3:
I couldn't remove a repeatable job of type 'every' when obtaining its options from `getRepeatableJobs`, then took a look at `getRepeatableJobs` and the decoding process wasn't prepared for an 'every' type job.


manast:
Closing since this functionality has already been added.

PR #1078, opened 2018-10-05
Remove .only from test

gabegorelick:
This was causing only one test to be run.

### DIFF
--- a/test/test_repeat.js
+++ b/test/test_repeat.js
@@ -644,7 +644,7 @@ describe('repeat', function() {
       );
   });
 
-  it.only('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
+  it('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
     var _this = this;
     var date = new Date('2017-02-07 9:24:00');
     this.clock.tick(date.getTime());


gabegorelick:
You may also want to add https://github.com/lo1tuma/eslint-plugin-mocha/blob/47f624abfac84b963007423cf5a35ca46ad4d981/docs/rules/no-exclusive-tests.md to prevent this from happening in the future.

test/test_repeat.js
```diff
@@ -644,7 +644,7 @@ describe('repeat', function() {
       );
   });
 
-  it.only('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
+  it('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
```
gabegorelick: Perhaps this was supposed to be a `.skip` since this test is failing?

gabegorelick:
Build is failing because this test doesn't pass. Perhaps it should be skipped?

test/test_repeat.js
manast: must have been committed while trying to fix the reason for the test to fail

manast:
I am gonna change our merge procedure, having a "develop" branch and only merge to master when we do a new release, so that master is kept green all the time.

PR #1079, opened 2018-10-05
Only pass `err.cause` to `failed` handlers if it's a Bluebird error

### LINKED ISSUE #1077: Bull passing error.cause to failed handlers
## Description

When failing a job with an Error, Bull attempts to grab the error's cause:

https://github.com/OptimalBits/bull/blob/cf2c6883fc2d861397bd59df7186b730800655e0/lib/queue.js#L1026-L1027

However, there's no guarantee `cause` will be an instance of Error. In fact, if you pass a [`VError`](https://github.com/joyent/node-verror), `cause` will be a _function_ that _returns_ the cause. This causes (no pun intended) Bull to pass incorrect values to any `failed` handler.

~I'm also not sure why Bull would want to return the cause and obscure the higher level error in the first place. If I'm passing a wrapped Error as the failure, surely that means I want that error to be returned, and not the lower level error.~ EDIT: there's a handy test case that explains the rationale: https://github.com/OptimalBits/bull/blob/cf2c6883fc2d861397bd59df7186b730800655e0/test/test_queue.js#L1213

```js
const VError = require('verror');

const queue = new Queue('foo');
queue.process('job', (job, callback) => {
    callback(new VError(new Error('root cause'), 'wrapper message'));
});
queue.on('failed', (job, err) => {
    console.log(err.stack); // prints `undefined`
    process.exit();
});

queue.add('job', {});
```

## Bull version
3.4.8

gabegorelick:
Hiding the internal `Bluebird.OperationalError` makes sense, but Bull should not unwrap user-supplied errors.

Fixes #1077

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1024,7 +1024,13 @@ Queue.prototype.processJob = function(job) {
   }
 
   function handleFailed(err) {
-    var error = err.cause || err; //Handle explicit rejection
+    var error = err;
+    if (
+      error instanceof Promise.OperationalError &&
+      error.cause instanceof Error
+    ) {
+      error = error.cause; //Handle explicit rejection
+    }
 
     return job.moveToFailed(err).then(function(jobData) {
       _this.emit('failed', job, error, 'active');
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -1299,6 +1299,37 @@ describe('Queue', function() {
       });
     });
 
+    it('process a job that rejects with a nested error', function(done) {
+      var cause = new Error('cause');
+      var jobError = new Error('wrapper');
+      jobError.cause = function() {
+        return cause;
+      };
+
+      queue.process(function(job) {
+        expect(job.data.foo).to.be.equal('bar');
+        return Promise.reject(jobError);
+      });
+
+      queue.add({ foo: 'bar' }).then(
+        function(job) {
+          expect(job.id).to.be.ok;
+          expect(job.data.foo).to.be.eql('bar');
+        },
+        function(err) {
+          done(err);
+        }
+      );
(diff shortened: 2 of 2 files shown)


test/test_repeat.js
```diff
@@ -644,7 +644,7 @@ describe('repeat', function() {
       );
   });
 
-  it.only('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
+  it.skip('should emit a waiting event when adding a repeatable job to the waiting list', function(done) {
```
gabegorelick: This test is failing on master.

lib/queue.js
```diff
@@ -1024,7 +1024,13 @@ Queue.prototype.processJob = function(job) {
   }
 
   function handleFailed(err) {
-    var error = err.cause || err; //Handle explicit rejection
+    var error = err;
```
manast: I prefer to just make the code simpler and just pass the err argument directly to the user in the failed event.

gabegorelick: @manast That would actually cause the user's error to *not* be returned in the following case:
```js
queue.process(function(job, callback) {
  jobDone(new Error());
});
```

Instead, a `Bluebird.OperationalError` would be passed to the `failed` handler. This arguably leaks implementation details to the caller and would be a change in API since it's not currently returned. It would also break this test case that explicitly checks for this:
https://github.com/OptimalBits/bull/blob/cf2c6883fc2d861397bd59df7186b730800655e0/test/test_queue.js#L1208-L1232

My PR was an attempt to have the best of both worlds: unwrap Bull's internal error but don't unwrap too much.

manast: ok, I get it.

PR #1084, opened 2018-10-15
Fix copy paste mistake for stacktrace in job toData

yyfearth:
I do believe this is a small mistake by copy paste but forgot to modify, which spotted when I read the code.
It does not hurt since the stacktrace will be overridden when error.
But still I think it should be corrected.


manast:
great, thanks!.

PR #1085, opened 2018-10-15
Add Job State Diagram to Documentation

jbutz:
It looked like the documentation had needed a job flow diagram for a while so I went ahead and created one. I used [draw.io](https://draw.io), and the XML file of the [diagram is in this gist](https://gist.github.com/jbutz/5a536d55dcc9fa056ff2bb46534a1b14). If you would like any changes let me know.


manast:
Looks great!. For completeness you could write an arrow between Failed and Delayed for the cases when you have auto retry enabled.

PR #1086, opened 2018-10-15
Update REFERENCE.md

popenkomaksim:
Hi,

I want to add useful note about pausing queue.
I had totally missed that during development phase, so I believe it could help someone take it into account sooner than later.

Thanks for great project!

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -301,6 +301,8 @@ Returns a promise that resolves when the queue is paused. A paused queue will no
 
 Pausing a queue that is already paused does nothing.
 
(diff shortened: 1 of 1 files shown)


popenkomaksim:
Just to make sure that we on the same page:
https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L58
function `pause() `do swap between `wait` <-> `paused`.

manast:
pause renames the wait list so that jobs are not being processed until renamed back to wait. But if the delayed jobs puts the jobs in a list called wait (as it is doing now), they will be processed directly not respecting the pause status.

popenkomaksim:
Yeah, I can confirm that behavior.
Thanks.

PR #1089, opened 2018-10-17
Fix wrong repeatable count updates

alexkh13:
I found that due the the un-pure nature of the nextRepeatableJob function the repeatable count wasn't reliable in all circumstances. 
The function should be pure whatsoever.


lib/repeatable.js
```diff
@@ -74,6 +74,9 @@ module.exports = function(Queue) {
               name,
               data,
               _.extend(_.clone(opts), {
+                repeat: _.extend(_.clone(opts.repeat), {
```
manast: This starting to get quite "hairy" with so many extends and clones. I think we can do better, maybe something like (untested):
```
return Job.create(
              _this,
              name,
              data,
              _.defaultsDeep( {
                repeat: {
                  count: currentCount
                },
                jobId: customId,
                delay: delay < 0 ? 0 : delay,
                timestamp: now,
                prevMillis: nextMillis
              }, opts)
            );
```

alexkh13: I didn't know there was a defaultsDeep in lodash. Good to know :)

I changed it and also added a test case to support my scenario and a documentation for the count attribute on the RepeatOpts interface.

alexkh13: I didn't know there was a defaultsDeep in lodash. Good to know :)

I changed it and also added a test case to support my scenario and a documentation for the count attribute on the RepeatOpts interface.

alexkh13: I didn't know there was a defaultsDeep in lodash. Good to know :)

I changed it and also added a test case to support my scenario and a documentation for the count attribute on the RepeatOpts interface.

alexkh13:
I didn't know there was a defaultsDeep in lodash. Good to know :)

I changed it and also added a test case to support my scenario and a documentation for the count attribute on the RepeatOpts interface.

PR #1091, opened 2018-10-18
Catch processor crashes

GabrielCastro:
Currently if a processor crashes or calls `process.exit` then the job hangs


lib/process/master.js
```diff
@@ -75,6 +75,17 @@ process.on('message', function(msg) {
...
 
+process.on('uncaughtException', function(err) {
+  if (!err.message) {
+    err = new Error(err);
+  }
+  process.send({
+    cmd: 'failed',
+    value: err
+  });
+  throw err;
```
manast: is this throw needed?

GabrielCastro: I would say so, if there's an uncaught exception, we have no guarentees as to the state of the worker so it's best to just let it crash after we pass the error up to the parent

PR #1096, opened 2018-10-22
1070 Add discard config for rate-limiter

### LINKED ISSUE #1070: Config option for rate-limiter to do nothing when enabled and hit
## Description

* As apart of the async multi-repo pattern, I would like a config for the rate-limiter to do nothing when it gets hit.
* Currently it throws jobs into the delayed queue.

```typescript
// enabling the rate limiter
const queueOptions = {
  rate: {
      max: 5,
      duration: 1000
    },
};
const queue = new Queue(description, queueOptions);
```

## Bull version
3.4.2

## Additional information

This has been discussed on this issue thread: https://github.com/OptimalBits/bull/issues/1053.
This issue is for tracking the new feature. I will be working on this.

aleccool213:
- Meant for Manually Fetching Jobs pattern

- closes #1070

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -1,4 +1,6 @@
-# Patterns
+
+Patterns
+========
 
 Here are a few examples of useful patterns that are often implemented with Bull:
 
@@ -12,26 +14,28 @@ Here are a few examples of useful patterns that are often implemented with Bull:
 
 If you have any other common patterns you want to add, pull request them!
 
-## Message Queue
+
+Message Queue
+-------------
 
 Bull can also be used for persistent message queues. This is a quite useful
-feature in some usecases. For example, you can have two servers that need to
-communicate with each other. By using a queue the servers do not need to be online at the same time, this creates a very robust communication channel. You can treat `add` as _send_ and `process` as _receive_:
+feature in some use cases. For example, you can have two servers that need to
+communicate with each other. By using a queue the servers do not need to be online at the same time, so this creates a very robust communication channel. You can treat `add` as *send* and `process` as *receive*:
 
 Server A:
 
 ```js
 var Queue = require('bull');
 
-var sendQueue = new Queue('Server B');
-var receiveQueue = new Queue('Server A');
+var sendQueue = new Queue("Server B");
+var receiveQueue = new Queue("Server A");
 
-receiveQueue.process(function(job, done) {
-  console.log('Received message', job.data.msg);
+receiveQueue.process(function(job, done){
+  console.log("Received message", job.data.msg);
   done();
 });
 
-sendQueue.add({ msg: 'Hello' });
+sendQueue.add({msg:"Hello"});
 ```
 
 Server B:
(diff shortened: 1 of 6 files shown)


test/test_rate_limiter.js
```diff
@@ -69,18 +69,71 @@ describe('Rate limiter', function() {
...
     );
 
-    queue.on('failed', done);
+    queue.on('failed', function(e) {
+      assert.fail(e);
+    });
+  });
+
+  it('should put a job into the delayed queue when limit is hit', function() {
+    queue.on('failed', function(e) {
```
csingh: Might be better to create a new queue here as you did in the test below, since this test assumes that the queue max is set to 1.

test/test_rate_limiter.js
```diff
@@ -69,18 +69,71 @@ describe('Rate limiter', function() {
...
+        return newQueue.add({});
+      })
+    ).then(function() {
+      Promise.all(
+        [1, 2, 3, 4].map(function() {
+          return newQueue.getNextJob();
+        })
+      ).then(function() {
+        return newQueue.getDelayedCount().then(function(delayedCount) {
+          expect(delayedCount).to.eq(0);
```
csingh: Should also do an assert on the main queue to make sure 3 objects are there. Might also want to assert the ordering of the jobs to make sure that is untouched?

lib/commands/moveToActive-8.lua
```diff
@@ -48,8 +49,12 @@ if jobId then
   if(ARGV[6]) then
     local jobCounter
     local maxJobs = tonumber(ARGV[6])
+    local discard = ARGV[8]
     jobCounter = tonumber(rcall("GET", KEYS[6]))
     if jobCounter ~= nil and jobCounter >= maxJobs then
+      if discard == 'true' then
```
csingh: Are there any tests for these .lua scripts?

PATTERNS.md
```diff
@@ -1,6 +1,4 @@
-
-Patterns
-========
+# Patterns
```
csingh: Do we need all these formatting changes in this PR? :) I would remove if unnecessary, its cluttering things up imo

REFERENCE.md
```diff
@@ -68,14 +66,17 @@ interface QueueOptions {
...
 
+When a job gets hit with the rate limiter, it is moved to the delayed queue.
+
 ```typescript
 interface RateLimiter {
-  max: number,      // Max number of jobs processed
-  duration: number, // per duration in milliseconds
+  max: number; // Max number of jobs processed
+  duration: number; // per duration in milliseconds
+  discard: boolean = false; // When jobs get rate limited, nothing happens
```
csingh: `discard` might not be the best name for this.. because we're not entirely discarding the rate limiter, we are just disabling the delayed queue. So maybe `disableDelyedQueue` or something similar?

test/test_rate_limiter.js
```diff
@@ -69,18 +69,71 @@ describe('Rate limiter', function() {
...
+      assert.fail(e);
+    });
+  });
+
+  it('should put a job into the delayed queue when limit is hit', function() {
+    queue.on('failed', function(e) {
+      assert.fail(e);
+    });
+
+    return Promise.all(
```
manast: maybe it is just better to write this just as:
```Promise.all([queue.add({}), queue.add({}), queue.add({}), queue.add({})])```

REFERENCE.md
manast: ok, now I full understand the use-case and discard may indeed not be the best name. I thought the idea was to really throw away the jobs that get "throttled". so this name will be misleading. On the other hand, a name like ```disableDelayedQeuue``` is a bit techie and exposes implementation details.

manast: Not sure about this name but what about this alternative ```bounceBack``` ? the option is quite special since it only has use in the manual processing pattern...

PATTERNS.md
aleccool213: will make another pr

aleccool213: https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint was used and automatically formatted it for me on save

REFERENCE.md
aleccool213: I like `bounceBack` 👍

aleccool213:
@manast new review is up 👍

PR #1102, opened 2018-10-24
Fix Warning: cancellation is disabled

alolis:
Bluebird configuration must have cancellation enabled (false by default) in order to be able to use promise.cancel().


manast:
hmm, I guess this can be merged, however I have started to replace bluebird by native promises since it gives a nice performance boost.

PR #1105, opened 2018-10-25
Implement 'onCancel' callback in sandbox environment

alolis:
The onCancel callback will automatically discard the job, move it to failed state and kill the child process.


alolis:
After some tests, I will close this PR myself since the flow is completely wrong and will break things.

PR #1106, opened 2018-10-25
Reset 'failedReason', 'finishedOn' and 'processedOn' fields on job retry

alolis:
The class members will be reset to null, failedReason field will be removed from Redis and the other two will be set to null.

### DIFF
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -1411,7 +1411,17 @@ describe('Queue', function() {
         expect(job.data.foo).to.be.eql('bar');
         expect(err).to.be.eql(notEvenErr);
         failedOnce = true;
-        job.retry();
+        job.retry().then(function() {
+          expect(job.failedReason).to.be.null;
+          expect(job.processedOn).to.be.null;
+          expect(job.finishedOn).to.be.null;
+
+          retryQueue.getJob(job.id).then(function(updatedJob) {
+            expect(updatedJob.failedReason).to.be.undefined;
+            expect(updatedJob.processedOn).to.be.undefined;
+            expect(updatedJob.finishedOn).to.be.undefined;
+          });
+        });
       });
 
       retryQueue.once('completed', function() {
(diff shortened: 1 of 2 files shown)


alolis:
Hm I can not understand why one of the tests fails

```
1) Queue
       Retries and backoffs
         should not retry a job that has been removed:

      AssertionError: expected 'Couldn\'t retry job: The job has been already retried or has not failed' to equal 'Couldn\'t retry job: The job doesn\'t exist'
      + expected - actual

      -Couldn't retry job: The job has been already retried or has not failed
      +Couldn't retry job: The job doesn't exist
```

manast:
Looks great. Could you improve some retry test case to verify the new functionality?

alolis:
@manast updated one of the tests, let me know if it will do.

test/test_queue.js
```diff
@@ -1411,7 +1411,11 @@ describe('Queue', function() {
         expect(job.data.foo).to.be.eql('bar');
         expect(err).to.be.eql(notEvenErr);
         failedOnce = true;
-        job.retry();
+        job.retry().then(function() {
+          expect(job.failedReason).to.be.null;
+          expect(job.processedOn).to.be.null;
+          expect(job.finishedOn).to.be.null;
```
manast: would be great to actually do a ```Queue.getJob()``` and check that they are also null after reading the job from redis.

alolis: After the retry() promise resolves (inside `then`) you mean? Because, I do not think they will be null after the promise resolves (?)

alolis: I updated the test, let me know if you meant something else.

manast: yes, it should only be null the failedReason since now it has been completed.
This makes me think however that the user would loose the information on why a job failed before the retry, what do you think?

alolis: They can always check failedReason before they execute `retry`. They have the opportunity to check it - if they want - before doing anything else so I think that the information is not lost since nothing automatically happens.

How about the other two fields though? Why aren't they timestamp integer after the retry completes?

manast: it is only returning the job as it is in that very moment

alolis: Ok got it now, thanks for clarifying :)

alolis:
@manast , do you need anything else or can this be merged?

PR #1127, opened 2018-11-15
process.NODE_ENV -> process.env.NODE_ENV

williamfeng91:
I'm getting errors about bluebird `longStackTraces` in production due to this typo. Is this going to be a breaking change?

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -42,8 +42,8 @@ var uuid = require('uuid');
 var commands = require('./commands/');
 
 Bluebird.config({
-  warnings: process.NODE_ENV !== 'production',
-  longStackTraces: process.NODE_ENV !== 'production',
+  warnings: process.env.NODE_ENV !== 'production',
+  longStackTraces: process.env.NODE_ENV !== 'production',
   cancellation: true
 });
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -44,7 +44,6 @@ var commands = require('./commands/');
 Bluebird.config({
   warnings: process.env.NODE_ENV !== 'production',
   longStackTraces: process.env.NODE_ENV !== 'production',
-  cancellation: true
 });
 
 /**
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -16,7 +16,7 @@ Bluebird.config({
   // Enable warnings.
   // warnings: true,
   // Enable long stack traces.
-  longStackTraces: process.NODE_ENV !== 'production',
+  longStackTraces: process.env.NODE_ENV !== 'production',
   // Enable cancellation.
   cancellation: true
 });


alolis:
Sorry guys; my bad about this...

manast:
Thanks for the fix. @alolis Wouldn't it work if we remove the Bluebird out from bull so that you can config it yourself as you need?

alolis:
@manast , for the `cancellation` specifically it seems that you can not.

As per [docs](http://bluebirdjs.com/docs/api/promise.config.html):

```
Cancellation is always configured separately per bluebird instance.
```

manast:
hmm, ok, but I do not think we use cancellation at all.

alolis:
Yeah I am aware, it was something that it helped me personally in my project to cancel a job. No worries you do what you have to do :) I will either do it in my own form or use the already in-place events system I have to `process.exit()` from within the processor.

But i really think a proper cancellation mechanism is needed at some point in order for the processors to do cleanups/graceful exists and to actually be able to remove a running job and not just waiting ones.

manast:
ok, I don't know your code, but I wonder if you could not just enable cancellation for the promises that you generate in your processor?

alolis:
With the current design there are two ways (or at least thats what I have figured out) to properly remove a running job. 

**Way 1:**

1. Change the `sandbox.js` file, add an `onCancel` handler in order to `discard()` the job whenever `jobPromise.cancel()` is called and then throw an error if cancellation is detected in order to trigger the `handleFailed` handler inside `queue`. You can find the whole change [here](https://github.com/OptimalBits/bull/pull/1128).

2. I keep track of all `jobPromise` objects returned by the `queue` active event and whenever I need to remove a specific job, I fetch the `jobPromise`, `jobPromise.cancel`.

3. Release the job lock and remove normally.

The above requires `cancellation: true`

**Way 2:**

This was the original thing I did but i then started working on something cleaner although I will switch back to this since cancellation is now completely out of the question and you are removing Bluebird completely.

1. Keep track of the `job` object from the `queue` `active` event. Wrap all my processors in order to listen to a global custom event `job:cancelled` via Redis pub/sub which will simply call `process.exit()` if the job id matches the one from the event.

2. Whenever I want to remove a job, i just fetch the tracked `job` object, I first `job.discard()` and then emit a `job:cancelled` event. Process will exit and then in `sandbox.js` , `child.on('kill')` will catch the kill, reject the promise and `handleFailed` will be called normally.

3. Release the job lock and remove the job.

Both of the above ways are working just fine but I hate the glue code.

**Way 3:** (Under thinking)

I want to refactor the processor handlers within the `queue` a bit and be able to accept `onCancel`, `onPause` and `onResume` hooks and each processor will be able to cleanup and exit gracefully. If this is done, then someone would be able to `queue.cancelJob(jobId)` or `queue.pauseJob(jobId)`. I just need some free time to get some code working and show it to you.

williamfeng91:
@alolis Sorry I didn't realize my second commit went into the pr. I disabled cancellation because I was getting `cannot enable cancellation after promises are in use` error, probably caused by other libs that uses bluebird before bull is loaded.

alolis:
@williamfeng91 no worries, the whole `Promise.config` block was removed anyway :)

PR #1128, opened 2018-11-15
Implement 'onCancel' handler in sandboxed environments

alolis:
This is my second attempt on the issue since PR #1105 wasn't correct and was breaking/affecting the flow of the library. This time cancellation will not impact the flow of the queue and the job will be cancelled nice and smoothly, plus it will ensure that the child process is killed.

### DIFF
--- a/lib/process/sandbox.js
+++ b/lib/process/sandbox.js
@@ -10,7 +10,7 @@ module.exports = function(processFile, childPool) {
         job: job
       });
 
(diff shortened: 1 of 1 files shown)


alolis:
I will close this since `Promise.config` was removed completely and `Bluebird` will also be removed so it's a pointless!

alolis:
I will close this since `Promise.config` was removed completely and `Bluebird` will also be removed so it's a pointless!

PR #1153, opened 2018-12-09
WIP: Replace bluebird function usages with native Promise

pc-jedi:
- [x] `.return` is a short hand by Bluebird. [Bluebird .return](http://bluebirdjs.com/docs/api/return.html)
If the ioredis client is created by the user without using Bluebird Promises the `.return` function is not available.
- [x] `.spread` is Bluebird implementation for destructuring an array of promises. [Bluebird .spread](http://bluebirdjs.com/docs/api/spread.html)
It can be replaced by ES6 destructuring.
- [x] `.timeout` [Bluebird .timeout](http://bluebirdjs.com/docs/api/timeout.html)
- [x] `.map` [Bluebird .map](http://bluebirdjs.com/docs/api/map.html)
- [x] `Promise.map` replaced by `Promise.resolve().map()`.
- [x] `.reduce` [Bluebird .reduce](http://bluebirdjs.com/docs/api/reduce.html)
- [x] `.filter` [Bluebird .filter](http://bluebirdjs.com/docs/api/filter.html)
- [x] `.each` [Bluebird .each](http://bluebirdjs.com/docs/api/each.html)
- [x] `.delay` [Bluebird .delay](http://bluebirdjs.com/docs/api/delay.html)
- [x] `Promise.delay` [Bluebird Promise.delay](http://bluebirdjs.com/docs/api/promise.delay.html)
- [x] `.asCallback` [Bluebird .asCallback](http://bluebirdjs.com/docs/api/ascallback.html)
- [ ] `.reflect` [Bluebird .reflect](http://bluebirdjs.com/docs/api/reflect.html)
- [ ] `Promise.promisify` [Bluebird Promise.promisify](http://bluebirdjs.com/docs/api/promise.promisify.html)
- [ ] `Promise.method` [Bluebird Promise.method](http://bluebirdjs.com/docs/api/promise.method.html)

Fix #1027

PR #1156, opened 2018-12-12
Remove bluebird

btd:
It does not change Promise.prototype if not need


pc-jedi:
@btd Nice work. I think I can then close my PR #1153 in favor of this one, since you're already done.

manast:
> It is a bit not clear what you are using yarn or npm, because i see both `yarn.lock` and `package-lock.json`. Also it should be possible to upgrade `ioredis` without issues.
Ok, I just learned that you can use package.lock.json in yarn https://yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/
so no need to keep yarn.lock, at least for the users of the module.

btd:
@manast i noticed your minimum supported node version is 6 (which is currently in maintaince mode). Currentl LTS versions is 8 and 10. But even with 6 supported, it is possible to modernize code base of bull (arrow functions, const/let, destructuring etc). This will greatly improve readability.
With minumum node version 8 it would be possible to use async/await.

pc-jedi:
@btd I agree. Maybe we should create an issue with a task list of what could be modernized.

manast:
@btd I have already started a new codebase in typescript. Bull 4.0 will be typescript only and use some redis 5.0 features as streams for events, etc.

btd:
@manast does this mean that you will not merge anything new to existing code base?

manast:
> @manast does this mean that you will not merge anything new to existing code base?

Of course I will. Bull 4.0 is long way to go anyway, and I will continue making fixes on the 3.x branch.

btd:
Ahh, ok. Thank you.

btd:
@manast what about suggesting about modernizing a bit existing code base (arrow functions, let/const etc)?

manast:
@btd would require babel to keep backwards compatibility with node. Thats why I think it is better to focus the effort on rewriting to typescript, a non-backwards compatible codebase that do not need to inherit many of the ugly hacks we made on the old one.

btd:
No it should not be required to use babel. See node.green. it shows that 6.x supports huge set of es6.

manast:
I am most interested in async/await, which would make most of the codebase much cleaner.

pc-jedi:
@manast Will we get a new bull release which includes this PR?

manast:
@pc-jedi I can release soon, but this is just an internal improvement as far as I understand.

zefexdeveloper:
@manast I'm just getting this error when using Bull, I hope this get merged soon.

PR #1168, opened 2019-01-10
Fix breakpoints in child processes

Vortec4800:
Bull currently strips out the inspect flags from the parent process when spawning child processes. This means if you run the program in an IDE, any breakpoints you set in your sandboxed processors will not fire, which can make development and debugging them much more difficult.

The solution is to detect if the inspect flag is set, then modify it to automatically select an open debug port. This keeps from conflicting with ports that are already in use by other processes. In a production environment where this flag is not set, there is no change to the arguments.

Tested by setting a breakpoint in a sandboxed process and running with WebStorm. Breakpoint in the child process was hit as expected.

If there are side-effects to this change that I'm not thinking of then please let me know, otherwise I think this is a pretty big quality-of-life improvement during development of the processors.


manast:
I agree. We need some green tests though :)

Vortec4800:
I ran the tests on my local machine and got the results 196 passing (1m), 6 pending. Do you know why they would be failing on CI?

twrobel3:
@Vortec4800 the problem here is that the behavior when a port isn't provided isn't to "automatically select an open debug port", but instead uses the default port of `9229`.  So in the tests you can see that once the second child process attempts to fork, you're getting errors of the form `Starting inspector on 127.0.0.1:9229 failed: address already in use`.

If @manast is okay with adding another dependency to the project, there are some pretty lightweight libraries out there to pick a random unused port, with [get-port](https://www.npmjs.com/package/get-port) being the most popular.  Conveniently, like Bull, it is MIT-licensed, so shouldn't add any complications to add here from a licensing perspective.

Using that, you could add a utility function to convert the parent's `execArgv` array to one with new, unused ports for the child process:

```js
const convertExecArgv = function(execArgv) {
  const standard = [];
  const promises = [];

  _.forEach(execArgv, arg => {
    if (arg.indexOf('--inspect') === -1) {
      standard.push(arg);
    } else {
      const argName = arg.split('=')[0];
      promises.push(
        getPort().then(port => {
          return `${argName}=${port}`;
        })
      );
    }
  });

  return Promise.all(promises).then(convertedArgs => {
    return standard.concat(convertedArgs);
  });
};
```

And you could restructure the retain function to use that helper as part of a promise chain:

```js
ChildPool.prototype.retain = function(processFile) {
  let child = this.getFree(processFile).pop();

  if (child) {
    this.retained[child.pid] = child;
    return Promise.resolve(child);
  }

  return convertExecArgv(process.execArgv).then(execArgv => {
    child = fork(path.join(__dirname, './master.js'), {
      execArgv
    });
    child.processFile = processFile;

    this.retained[child.pid] = child;

    child.on('exit', this.remove.bind(this, child));

    return initChild(child, child.processFile).then(() => {
      return child;
    });
  });
};
```

Vortec4800:
Thanks for the info. When I was implementing this and researching how breakpoints are triggered in webStorm (which is the IDE I'm using), a post on their bug tracker stated that running that flag without a port automatically selected an open port. While that does seem to work in webStorm, it obviously doesn't work outside of there which is why this is happening, and that makes sense. That also explains why my breakpoints work fine when I debug in webStorm, but these tests are failing.

We'll have to see if @manast is okay with adding a framework to manage this. While I think it's a good idea as the dev experience is important, I can see the argument for keeping the list of external dependencies as small as possible.

manast:
I see that the proposed module is really tiny and has no external dependencies so I am fine with it if it resolves this issue.

Vortec4800:
Okay, let me put this together and you guys can check it out.

Vortec4800:
Green!

manast:
LGTM

PR #1173, opened 2019-01-14
fix(pool): added optional setting for killing processes and not retaining them

DavisJaunzems:
Created optional setting for child pool, default is set to `true` - always retains child processes, but when set to `false` the child process is killed and not retained after finishing processing.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -48,7 +48,7 @@ function setDefaultOpts(opts) {
   var _opts = Object.assign({}, opts);
 
   _opts.attempts = typeof _opts.attempts == 'undefined' ? 1 : _opts.attempts;
-  _opts.delay = typeof _opts.delay == 'undefined' ? 0 : _opts.delay;
+  _opts.delay = typeof _opts.delay == 'undefined' ? 0 : Number(_opts.delay);
   _opts.timestamp =
     typeof _opts.timestamp == 'undefined' ? Date.now() : _opts.timestamp;
 
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -229,7 +229,8 @@ var Queue = function Queue(name, url, opts) {
     guardInterval: 5000,
     retryProcessDelay: 5000,
     drainDelay: 5,
-    backoffStrategies: {}
+    backoffStrategies: {},
+    keepProcesses: true
   });
 
   this.settings.lockRenewTime =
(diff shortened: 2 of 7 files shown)


DavisJaunzems:
@manast would appreciate if you could take a look at this PR

lib/job.js
```diff
@@ -48,7 +48,7 @@ function setDefaultOpts(opts) {
   var _opts = Object.assign({}, opts);
 
   _opts.attempts = typeof _opts.attempts == 'undefined' ? 1 : _opts.attempts;
-  _opts.delay = typeof _opts.delay == 'undefined' ? 0 : _opts.delay;
+  _opts.delay = typeof _opts.delay == 'undefined' ? 0 : Number(_opts.delay);
```
manast: this fix seems to have been slipped in from an unrelated issue.

test/test_child-pool.js
```diff
@@ -142,4 +142,20 @@ describe('Child pool', function() {
...
+
+  it('should kill child after processing is finished and not retain it', function() {
+    var processor = __dirname + '/fixtures/fixture_processor_bar.js';
+    var child;
+
+    pool.setKeepProcesses(false);
+
+    return pool.retain(processor).then(function(_child) {
+      expect(_child).to.be.ok;
+      child = _child;
```
manast: Why this assignment?

DavisJaunzems: Must have been from copy-paste, you are right no need for re-assigned the variable, could `pool.release(_child);` directly

DavisJaunzems: Updated PR, removed the unnecessary reassignment

lib/queue.js
```diff
@@ -250,6 +251,13 @@ var Queue = function Queue(name, url, opts) {
   this.processJob = this.processJob.bind(this);
   this.getJobFromId = Job.fromId.bind(null, this);
 
+  // Check settings to see if processes will be kept after finishing processing, default set to true
+  this.keepProcesses =
```
manast: Since we are setting a default value ```true``` a few lines above, this is not necessary.

lib/queue.js
```diff
@@ -672,6 +680,10 @@ Queue.prototype.start = function(concurrency) {
   });
 };
 
+Queue.prototype.setKeepProcesses = function(keepProcesses) {
```
manast: I do not think we need a setter for this setting, as with the other settings.

DavisJaunzems:
@manast what command did you run to get these errors, because I can run all of the tests locally?
Also looking at the format issues (using `let` or `const` instead of `var` and using arrow functions), all of the tests should have the same formatting issues..

DavisJaunzems:
@manast updated the code, anything else before this can be merged?

ks-s-a:
The PR was spoiled, but the feature is good. @manast may it be merged for the next release?

manast:
In order to merge this we need to rework it according to the comments and make it pass the tests.

PR #1180, opened 2019-01-17
Upgrade codebase to improve readability 

btd:
All changes should be safe without any possible breaking changes for module users.


aleccool213:
@btd All good changes but I am 99% sure this break compatibility with old versions of Node.

btd:
@aleccool213 it does not break anything up to 6.x which is lowerst supported (according to travis config)

manast:
yes, it passed all the tests so it means it is compatible with at least version 6.

manast:
Looks great. Thanks!.

PR #1182, opened 2019-01-17
Fix #1148 job dropping when rate-limiter hits during discard mode

### LINKED ISSUE #1148: Job is dropped when rate limit hit and bounceback is set to true

- set bounceBack to true for a queue
- hit the rate limit
- job is dropped instead of it being in the `waiting` queue

## Bull version

latest

## Additional information

- this bug is minimal impact currently because bounceBack was just recently introduced
- Here is where the code removes the job from the active queue which is correct: https://github.com/wishabi/bull/blob/632ad7c95cae9bb0fc273b6243ffb9bd1e7888a1/lib/commands/moveToActive-8.lua#L67
- What doesn't happen is that it is added back to the waiting queue

mehdimu:
### Notes
- This is a fix for #1148
- Fixed the issue where when rate limit hits, the jobs were been dropped
- Fixed the original test case for `should not put a job into the delayed queue when discard is true`, where it was equating `getActiveCount` instead of `getWaitingCount`

cc: @aleccool213

### DIFF
--- a/lib/commands/moveToActive-8.lua
+++ b/lib/commands/moveToActive-8.lua
@@ -62,9 +62,12 @@ if jobId then
         -- put job into delayed queue
         rcall("ZADD", KEYS[7], timestamp * 0x1000 + bit.band(jobCounter, 0xfff), jobId)
         rcall("PUBLISH", KEYS[7], timestamp)
+        -- remove from active queue
+        rcall("LREM", KEYS[2], 1, jobId)
+        return
       end
-      -- remove from active queue
-      rcall("LREM", KEYS[2], 1, jobId)
+      -- move from active to wait
+      rcall("RPOPLPUSH", KEYS[2], KEYS[1])
       return
     else
       jobCounter = rcall("INCR", rateLimiterKey)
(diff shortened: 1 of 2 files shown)


lib/commands/moveToActive-8.lua
```diff
@@ -62,9 +62,12 @@ if jobId then
         -- put job into delayed queue
         rcall("ZADD", KEYS[7], timestamp * 0x1000 + bit.band(jobCounter, 0xfff), jobId)
         rcall("PUBLISH", KEYS[7], timestamp)
+        -- remove from active queue
+        rcall("LREM", KEYS[2], 1, jobId)
+        return
       end
-      -- remove from active queue
-      rcall("LREM", KEYS[2], 1, jobId)
+      -- move from active to wait
```
manast: use ```else``` for making intent more explicit.

aleccool213: I am thinking actually if we put this code in an `if` statement and the move the code in the `if` statement right now outside, this would be more clear.

PR #1191, opened 2019-01-23
Added parameter to optional fetching next job

rysi3k:


aleccool213:
Not fetching the next job should be the default behaviour imo. Seems magical and when I first started using these methods, I did not realize it was doing this until much later.

PR #1230, opened 2019-02-26
Throwing error when trying to insert object with existing ID

MicahMartin:
Now we can do something like 

```
  try {
    const job = await JobQueue.add('storyJob', data, {jobId: `${id}:${lastMod}`});
    res.json({message: `storyID ${id}`});
  } catch (e) {
    console.log(`Error adding job!: ${e}`);
    res.status(500).json({message: `${e}`});
  }
```

### DIFF
--- a/lib/commands/addJob-6.lua
+++ b/lib/commands/addJob-6.lua
@@ -47,7 +47,7 @@ else
   jobId = ARGV[2]
   jobIdKey = ARGV[1] .. jobId
   if rcall("EXISTS", jobIdKey) == 1 then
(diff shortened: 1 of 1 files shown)


manast:
this is a breaking change that may be challenging for existing users to adapt. 
Two alternatives:
1) emit an event, like 'ignored'.
2) add a status variable to the object instance, ```ignored``` that is true if the job has been ignored.

MicahMartin:
I thinking adding a status variable to the object instance is a good approach, I'll give it a shot

PR #1242, opened 2019-03-13
Correct call signature for 'process' in guide

stefanpl:
The call signature for the `process` function, as shown in the Quick Guide, is highly misleading: There is no real need to provide a second parameter, and its naming (`data`) suggests that is contains the job's data (which is however stored in `job.data`). If a second parameter is passed, it will be `done` callback, which is not really necessary and should be omitted.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -71,8 +71,8 @@ that defines a process function like so:
 ```js
 const myFirstQueue = new Bull('my-first-queue');
 
-myFirstQueue.process(async (job, data) => {
-  return doSomething(data);
+myFirstQueue.process(async (job) => {
+  return doSomething(job.data);
 });
 ```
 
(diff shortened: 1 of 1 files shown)


balexandre:
shouldn't the `readme.md` from the root also be updated?

stefanpl:
@balexandre it mostly uses `queue.process(function(job, done){ …`, but also mentions the possibility to return a promise, leaving out the `done` from the call signature: 

```
pdfQueue.process(function(job){
  // Processors can also return promises instead of using the done callback
  return pdfAsyncProcessor();
});
```
For me that's okay, nothing confusing there. The `done` callback is named properly here and used in a way that's fairly easy to understand. My main problem with `docs/README.md` it that it suggests the second parameter to `.process` would contain the job's data, which is very confusing.

manast:
thanks for the patch. The documentation was wrong.

PR #1252, opened 2019-03-25
Update sandbox exit handler to log signals

GabrielCastro:


manast:
thanks for the contribution

PR #1256, opened 2019-03-28
docs: lockRenewTime in AdvancedSettings

Eomm:
Added `lockRenewTime` parameter in `AdvancedSettings` overview


manast:
thanks for the improvement!

PR #1259, opened 2019-03-29
Don't use `latest` for get-port version

gabegorelick:
This fixes churn in package-lock.json as well as preventing future breakage when get-port releases a backwards incompatible version.

The `get-port` dependency was added in #1168.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1518,9 +1518,9 @@
       "dev": true
     },
     "get-port": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/get-port/-/get-port-4.1.0.tgz",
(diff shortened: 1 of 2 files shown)


manast:
thanks, an overlook from my side when code reviewing the pr.

gabegorelick:
@manast There is a new major version of get-port since this was merged. So there's the possibility of breaking changes. It may be wise to release this.

gabegorelick:
> however we deliver package-lock and yarn.lock so in practice it should not be a problem

Those are not used when installing a package. They're only used when hacking on the package itself.

https://docs.npmjs.com/files/package-lock.json:
> One key detail about package-lock.json is that it cannot be published

https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-current-package-only:
> During install Yarn will only use the top-level yarn.lock file and will ignore any yarn.lock files that exist within dependencies.

Issue #1262, opened 2019-04-03
ERR unknown command ' client '


minzojian:
## Description  
node_modules\bull\lib\worker.js; line 17  
![image](https://user-images.githubusercontent.com/361781/55461394-e9a28900-5626-11e9-9f76-db208d4b23ac.png)

an error thrown out while check redis is ready  
actual error message is **ERR unknown command \' client \'**   
but the condition is   
`if (err.message !== 'ERR unknown command \'client\'') throw err;`  
notice in the actual error message, there are spaces around the "client", so err has been throw out.  
 
as i  run my app in docker containner, when this error throw out, app just restart. that's  too bad.  
is that a redis bug or something?

manast:
your redis server does not support that command, if you check the issues you will find workarounds.

minzojian:
yes，setname command is not supported in my cloud server，is there a way to skip the setname command in worker.js

minzojian:
created a pull request, for adding a plus condition for my strange situation
https://github.com/OptimalBits/bull/pull/1263

minzojian:
created a pull request, for adding a plus condition for my strange situation
https://github.com/OptimalBits/bull/pull/1263

minzojian:
created a pull request, for adding a plus condition for my strange situation
https://github.com/OptimalBits/bull/pull/1263

PR #1263, opened 2019-04-03
Update worker.js

minzojian:
sometimes, the 'ERR unknown command \'client\'' check is not enough, there is another like "ERR unknown command ' client '". there are spaces around the "client" word.

check the issues
https://github.com/OptimalBits/bull/issues/1262


foxling:
#1240

PR #1265, opened 2019-04-04
Improve Job#progress reference

ArnaudValensi:
The reference was missing the detail Job#progress can be used to retrieve job progress and not only set it.


manast:
thanks for the fix!

PR #1266, opened 2019-04-04
bumped up the version of depended node_modules to the newest ones

zhaow-de:


zhaow-de:
updated the minimal required version of Node.js to 8.x, because 6.x is about to be out of maintenance by April 30, 2019. According to [the official schedule](https://github.com/nodejs/Release)

manast:
thanks for the patch. I always update the packages before every release, as described in the CONTRIBUTING readme, so I will close this PR, in the next release all packages will be uptodate.

PR #1267, opened 2019-04-04
fixed the wrong invocation parameters in a dead function

zhaow-de:
Luckily `scripts#moveToFailed` has never been called, because it invokes `scripts#moveToFailedArgs` with wrong parameters. Seems like a copy&paste mistake :)

### DIFF
--- a/lib/scripts.js
+++ b/lib/scripts.js
@@ -208,11 +208,8 @@ const scripts = {
     const args = scripts.moveToFailedArgs(
       job,
       failedReason,
-      'failedReason',
       removeOnFailed,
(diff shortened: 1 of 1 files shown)


ks-s-a:
@manast I'v found a problem with removing task after lock missing due to having long sync action inside. When there is a combination of missing lock and an error as a result of task processing - queue ignore removeOnFail setting and save the task as 'failed'. May it be fixed by the PR?

I could post a separate issue it you want.

manast:
@ks-s-a better a separate issue :).

PR #1268, opened 2019-04-05
docs: add link to prometheus exporter

GabrielCastro:
We created a Prometheus exporter for bull,
we've been using in prod for several months now and think more users would benefit from it.

https://github.com/UpHabit/bull_exporter

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,12 @@ There are a few third-party UIs that you can use for monitoring:
 
 ---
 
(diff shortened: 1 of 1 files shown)


manast:
Great, thanks!

rvalim:
Hi guys,

Would like to use the bull_exporter, seems to fit what I need, do you guys have some documentation, cos I didn't find it.

Tks

PR #1269, opened 2019-04-06
moved scripts#cleanJobsInSet to Lua

zhaow-de:


manast:
good job, and thanks for the clean ups.

PR #1271, opened 2019-04-07
removed a redundant parameter at Job#addJob

zhaow-de:
`scripts#addJob` accepts 4 parameters, but the invocation at `Job#addJob` sents 5. The last `queue.token` is not used and not necessary.


manast:
thanks for the fix.

PR #1292, opened 2019-04-23
Storing finished date in job object

rysi3k:
Small change for storing in job object date of finished job.
Previuosly it was only stored in redis.


manast:
looks good thanks

PR #1293, opened 2019-04-23
Taking lock without checking queue token

rysi3k:


manast:
I would like to know more about this use case, you are aware that if the lock is still alive some worker is processing the job?

rysi3k:
It happens in case, when worker started async job, but master node.js process is restarted for some reason. So queue has other token and job is marked as stalled, then it is restarted from begining. I can't restart worker job, so I must force take lock to process job to the end.

Issue #1222, opened 2019-02-18
Redis network bandwidth consumption


abriosi:
## Description

I have a cluster with 30 workers consuming jobs from redis using bull (around 0.5 tasks per second, and each task lasts for 1 minute). Each worker restarts every hour and is on one of 3 different nodes (which might be different from the one redis is hosted).

I want to debug or understand why the network consumption is around 25mb/s. Is this normal behavior from bull and how do I debug this problem?

Each job payload is a simple string.

## Bull version

`"bull": "^3.5.2"`

## Additional information

Each queue has around 10k of tasks waiting for completion and/or delayed. We can see the traffic is between redis and the workers although only a string is fetched

manast:
It is ok to leave the issues even if you discovered that the bug was in your side, it can be useful to other users :).

abriosi:
@manast I thought I had solved the problem, but I guess not. Can you provide some guideline on how to debug this issue? I was thinking about booting up a wireshark micro service

manast:
ok, so lets see if I understand the workload: 30 workers, each one processing 0.5 jobs/second, which means 15 jobs/second in total, since thew payload is around 10k, the total amount of network bandwidth consumption should lie around 150Kb + some small queue overhead, right?

abriosi:
@manast yes you are about right. I have managed to run `redis-cli monitor` and it seems that there is a lot of talk between `bull` and `redis` regarding `delayed tasks`

manast:
can you try to upgrade to 3.7.0? there are some important improvements regarding delayed jobs, if you are using a lot of them it may help.

abriosi:
@manast I will. Right now it seems `drainDelay: 500` solved the problem. It seems that there is some logic that is always checking redis for jobs switching context

abriosi:
@manast I have tried `"bull": "^3.7.0` and my redis is still getting flooded with messages like this

![bandwidth_consumption](https://user-images.githubusercontent.com/23288462/53035952-88579980-346e-11e9-811c-48f7ab909a32.png)

From what I understand this seems like routine everyone of my 30 workers is running checking the delayed queue, hence de 20mb/s network traffic

abriosi:
Is this line responsible for the flood?

https://github.com/OptimalBits/bull/blob/develop/lib/commands/updateDelaySet-6.lua#L24

abriosi:
[...]

Issue #1226, opened 2019-02-25
[3.7.0] Higher CPU/Memory consumption


roychri:
## Description

We recently upgraded to 3.7.0 but we had to rollback to 3.6.0 due to CPU and Memory consumption growing over time to the point of affecting the event loop so much everything would get deplayed.

![CPU/Memory graph](https://content.screencast.com/users/christian.s.roy/folders/Snagit/media/1e383033-eef6-43d8-86b4-96cdbbca08c8/02.25.2019-09.44.png)

As you can see from this graph, both cpu and memory consumption increased dramatically. The sudden drops is when the services/processes were restarted. Near the end, it was being restarted by a script every 2 hours to keep it somewhat steady. It took me a few days to suspect something else than my own code was doing this.

## Minimal, Working Test code to reproduce the issue.

I have no proof but I suspect one of our main queue is the culprit. It's the one with the most load (and its not that much).
![Arena](https://content.screencast.com/users/christian.s.roy/folders/Snagit/media/d0b3f298-50be-46aa-923a-a8db9ed297a0/02.25.2019-09.47.png)
![Job JSON](https://content.screencast.com/users/christian.s.roy/folders/Snagit/media/8978692b-a583-4aaf-8264-43eb532af6a8/02.25.2019-09.47.png)

```
const queue = new Queue( 'queue_name', redis_url, {
    limiter: {
        max: 99,
        duration: 1000
    }
});
queue.add({
    member_id: event.member_id,
    force: true
}, {
    delay: 5 * 60 * 1000, // 5 minutes
    attempts: 5,
    backoff: {type: 'fixed', delay: 60000}
});
```
## Bull version
3.7.0

## Additional information

The problem starts appearing after ~3-6 hours of working with new jobs appearing every minutes, sometimes every seconds.
I suspect the limiter being the problem. This option is what has been causing the most traffic (see #1222). Whenever I had problems with the slave not being able to keep up is when the limiter was being used by most of my queue. When I disabled the limiter for those that could work fine without it, the Redis slaves stopped lagging like crazy.

manast:
thanks for the report. I would guess it is related to the rate limiter. The aim was to actually make it better in 3.7 so that it took less resources, but it seems I failed at that. Will try to implement a better rate limiter that is not based in the delay set, which seems to create a lot of issues.

gcox:
Seeing the same behavior. You probably already know what the problem is but here's the limiter options I'm using and my use case:

```typescript
{
[...]

PR #1294, opened 2019-04-23
Rewrite timing logic for updating delay set

holm:
This rewrites the logic around updating delay set to be a bit safer and simpler (imo).

In the previous code there was a race-condition due to the timer being cleared before the call to process the delay set, and in the time until the processing was done, another processing could have been started.

The new code does away with the guardian timer, and instead combines the guardian timer with the updates coming from new jobs being posted.

Refs https://github.com/OptimalBits/bull/issues/1222

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -42,7 +42,7 @@ const commands = require('./commands/');
     job -> wait -> active
        \     ^            \
         v    |             -- > failed
-        delayed 
+        delayed
 */
 
 /**
@@ -56,7 +56,6 @@ const commands = require('./commands/');
   delayed job is processed after timing out.
 */
 const MINIMUM_REDIS_VERSION = '2.8.18';
-const MAX_TIMEOUT_MS = Math.pow(2, 31) - 1; // 32 bit signed
 
 /*
   interface QueueOptions {
@@ -304,31 +303,6 @@ function redisOptsFromUrl(urlString) {
   return redisOpts;
 }
 
-function setGuardianTimer(queue) {
-  return setInterval(() => {
-    const now = Date.now();
-    if (
-      queue.delayedTimestamp < now ||
-      queue.delayedTimestamp - now > queue.settings.guardInterval
-    ) {
-      scripts
-        .updateDelaySet(queue, now)
-        .then(timestamp => {
-          if (timestamp) {
-            queue.updateDelayTimer(timestamp / 4096);
-          }
-          return null;
-        })
-        .catch(err => {
-          queue.emit('error', err);
-        })
-        .then(() => {
-          return null;
-        });
-    }
-  }, queue.settings.guardInterval);
-}
-
 util.inherits(Queue, EventEmitter);
 
 //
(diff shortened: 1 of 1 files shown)


manast:
thank you for your attempt, I tried myself to reduce the complexity of the code before, however it is quite difficult to fix all the edge cases, for instance with this solution most repeatable job tests are failing. As a side note you may be interested in the new implementation for bull 4, however this requires redis 5.0 and above: https://github.com/taskforcesh/bull/pull/1/files#diff-e43f718fe879603152e9c16fba64d4cfR49

holm:
This seems to just be due to the way the tests are written. I can change the code a little bit to make the tests pass as they are.

holm:
Should be all green now

lib/queue.js
```diff
@@ -851,49 +814,38 @@ Queue.prototype.run = function(concurrency) {
...
+Queue.prototype.updateDelayTimer = function() {
+  return scripts
+    .updateDelaySet(this, Date.now())
+    .then(nextTimestamp => {
+      this.delayedTimestamp = nextTimestamp
+        ? nextTimestamp / 4096
+        : Number.MAX_VALUE;
+
+      // Clear any existing update delay timer
+      if (this.delayTimer) {
```
manast: this if is not needed since clearTimeout(0) is a noop

manast:
good job. I will merge this but I will wait a couple of days before doing a release so that you can run this in production and check if it works better than the previous version in your scenario.

holm:
Thanks! I will update our production environment tomorrow and report back what I see there.

gcox:
@holm @manast This has been working well for me for about 12 hours now.

holm:
@gcox That is good to hear. Did you have problems before, and what were they exactly?

gcox:
@holm Yes, I suffered from #1226 as well as #1222

#### Setup
A rate limited queue (1 job every 3 seconds), with about 1000 jobs.

#### Issues

**Excessive bandwidth** - Bandwidth for my redis instance was averaging 5Mbps in & out. Interestingly, in 3.6, it was averaging 5Mbps in & 1.5Mbps out. In any case, with your fix, it's now averaging 0.25Mbps in & <1Mbps out.

**High CPU** - Once a worker was started, CPU usage increased exponentially. After a couple of minutes it was at 100%. Now it's back to a consistent low usage.

I believe both of those issues were made worse due to using the rate limiter.

---

So, thanks for your fix 👏

holm:
Thats great, thanks for the update. 

I must admit I find queue logic really hard to get right, so I would definitely keep an eye on the jobs and make sure all the delayed and rate limited jobs gets properly processed, just to be on the safe side :)

lib/queue.js
```diff
@@ -851,49 +814,38 @@ Queue.prototype.run = function(concurrency) {
...
+      ]);
 
-    const update = delay => {
-      if (delay) {
-        delay = delay < MAX_TIMEOUT_MS ? delay : MAX_TIMEOUT_MS;
-        this.delayTimer = setTimeout(delayUpdate, delay);
+      // Schedule next processing of the delayed jobs
+      if (delay <= 0) {
+        // Next set of jobs are due right now, process them also
+        this.updateDelayTimer();
```
gabegorelick: I'm getting warnings about unhandled promises stemming from this. It seems like this should return here?

gabegorelick: Nevermind, I see `updateDelayTimer` handles errors itself. I must be running into something else.

PR #1305, opened 2019-05-02
Update feature comparison table

tamlyn:
Thanks for the useful feature comparison table. I notice that Bee Queue has gained some powers since the table was created though.

Bee Queue now supports [global events](https://www.npmjs.com/package/bee-queue#queue-pubsub-events), [delaying jobs](https://www.npmjs.com/package/bee-queue#jobdelayuntildatetimestamp) and a web UI via [Arena](https://github.com/bee-queue/arena) which is now hosted in the same org.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ There are a few third-party UIs that you can use for monitoring:
 **Bull v3**
 
 - [Taskforce](https://taskforce.sh)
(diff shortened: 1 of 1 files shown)


manast:
The problem is that Bee queue keep stealing code from Bull without crediting, so I will not make free publicity for them here.

PR #1311, opened 2019-05-06
Fix check for supported file types

michaltk:
[There was a check for supported file types added](https://github.com/OptimalBits/bull/pull/1186) but it’s [checking](https://github.com/OptimalBits/bull/blob/7e10fd206bf6aaa8a1886170f3621d483c290b16/lib/queue.js#L632) for `supportedFileTypes.indexOf(path.extname(handler)) !== '-1' `,
`indexOf` returns a number `-1` not a string `'-1'`, so this check was always true.

Updated to  number `-1` instead of string.


michaltk:
#1320

PR #1320, opened 2019-05-16
Use correct type for !== comparison

ianp:
Sorry it's taken me so long to get around to this!

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -626,10 +626,10 @@ Queue.prototype.setHandler = function(name, handler) {
   this.setWorkerName();
 
   if (typeof handler === 'string') {
-    const supportedFileTypes = ['js', 'ts', 'flow'];
+    const supportedFileTypes = ['.js', '.ts', '.flow'];
     const processorFile =
       handler +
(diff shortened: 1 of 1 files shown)


manast:
tests are broken though with errors like ```(node:4734) UnhandledPromiseRejectionWarning: Error: File /home/travis/build/OptimalBits/bull/test/fixtures/fixture_processor_slow.js.js does not exist
```

ianp:
Ha! Yes, turns out they were only passing before because the comparison failing was hiding the fact. that the list of supported extensions was incorrect. My bad for not checking.

I've also replaced the `indexOf !== -1` test with `includes` which was added in Node 6, which is the minimum supported version according to the package file, and reads better IMHO.

ianp:
As far as I can see the CI tests are failing because of a module version incompatibility, unrelated to this PR. The version of `get-stdin` (added 10 days ago) looks like it needs Node 8. Not sure if you want to make 8 the minimum supported version of do something about that dependency?

manast:
@ianp yes I think I will remove nodejs 6 support for the next version because it is quite old by now anyway. Thanks for the fix.

PR #1333, opened 2019-06-03
docs: fixed global listener example

stansv:
Just checked, if you pass `"global: completed"` as event name, the callback won't be triggered. The correct is `"global:completed"`.


manast:
very good catch thank you.

Issue #1340, opened 2019-06-11
queue.add still throw exception, when ioredis reconnected


HsinHeng:
## Description
I want to test reconnect mechanism of bull/ioredis, my test program as bellow
```
const queue = new Queue("test", {
		redis: {
			maxRetriesPerRequest: null,
			retryStrategy: function (times) {
				console.log("connection retry...");

				return Math.min(times * 50, 2000);
			},
		},
	});

setInterval(() => queue.add({ msg: "hello world" }), 3000);
```

1. Use Case 1
turn off redis -> start my program ( ioredis try to reconnect...) -> turn on redis
however, queue.add still throw Error
```
{ Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1113:14)
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379 }
```

2. Use case 2
redis is on -> 
start program ->
turn off redis ->
my program connection retry.. -> 
turn on redis-server ->
reconnected & queue.add work fine.

How can fix the use case 1  or is it  a bug?

## Bull version
3.10.0

joewagner:
I ran into this issue today.  Wanted give a +1 for fixing and to present a potential workaround.  
The idea is to just kill the process and fail as loud as possible if the redis connection can't be established during startup, which is the cause of this issue afaik.

```
const queue = new Queue('worker queue', {
    redis: this.config.redisOptions
});

let initialConnection;
queue.isReady().then(() => {
    initialConnection = true
},
err => {
    if (!initialConnection) {
        console.error('Cannot establish connection to Redis during startup, exiting...')
        console.error(err);

        process.exit(err);
    }
});
```

Any input or alternatives to this approach would be much appreciated

manast:
as promises works it will either resolve or reject, so in your example above you do not have any use for the "initialConnection" variable.

joewagner:
@manast good call, thank you!

zyf0330:
I just cannot find doc about `isReady` in REFERENCE.md

PR #1343, opened 2019-06-11
Renew _initializing promise when ioredis fails to apply lua scripts

stansv:
Please review, I don't like this solution but it seems to work.

Fixes https://github.com/OptimalBits/bull/issues/1340.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -147,16 +147,18 @@ const Queue = function Queue(name, url, opts) {
     client.on('error', this.emit.bind(this, 'error'));
 
     if (type === 'client') {
-      this._initializing = commands(client).then(
-        () => {
-          debuglog(name + ' queue ready');
-        },
-        err => {
-          this.emit('error', new Error('Error initializing Lua scripts'));
-          throw err;
-        }
-      );
-
+      function initScripts(instance) {
+        return commands(client).then(
+            () => {
+              debuglog(name + ' queue ready');
+            },
+            err => {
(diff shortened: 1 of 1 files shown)


manast:
The problem is that ```_initializing``` is also been used by```queue##isReady``` so it may create some problems, need to think more about it. Maybe it can be recreated with a test case.

stansv:
@manast OK I think this PR can be closed, I hope you or someone would have time to redesign initialization mechanism and fix this better way.

lib/queue.js
```diff
@@ -147,16 +147,18 @@ const Queue = function Queue(name, url, opts) {
...
+              debuglog(name + ' queue ready');
+            },
+            err => {
+              instance.emit('error', new Error('Error initializing Lua scripts'));
+              instance._initializing = initScripts(instance);
+            }
+        );
+      }
+      this._initializing = initScripts(this);
       this._initializing.catch((/*err*/) => {});
```
JesperMadsen123: Maybe remove the "instance._initializing = initScripts(instance);" from the error handling part of initscripts, and 
add it to the catch instead.

Maybe the "instance.emit('error', new Error('Error initializing Lua scripts'));" also could be handled in the catch part.

this._initializing.catch((/*err*/) => {
   instance.emit('error', new Error('Error initializing Lua scripts')); // maybe use err here..
   setImmediate(()=> {
      instance._initializing = initScripts(instance);
   });
});

Issue #1344, opened 2019-06-11
queue.isReady() throws error if client not initialized


gabegorelick:
## Description

If `queue._initializing` is not set, `queue.isReady()`will throw a `TypeError: Cannot read property 'then' of undefined` at https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/lib/queue.js#L478 

`_initializing` is set when the main Redis client is initialized: https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/lib/queue.js#L150

Setting `skipVersionCheck: true` is the only way to trigger this as far as I can tell. Otherwise, `getRedisVersion(this.client)` will cause `._initializing` to be set before the `Queue` constructor returns.

## Minimal, Working Test code to reproduce the issue.

```js
const queue = new Queue('foo', {skipVersionCheck: true});
queue.isReady();
```

## Bull version
`develop`

## Additional information

Anything that calls `isReady` will also throw, e.g. `queue.close()`, `queue.add()`, etc.

ifokeev:
same issue here with `node-redis`

PR #1347, opened 2019-06-11
fix: whenCurrentJobsFinished shouldn't initialize bclient

### LINKED ISSUE #1346: whenCurrentJobsFinished initializes bclient unnecessarily
## Description

Bull typically does a good job of lazily connecting to Redis. But the following code causes the blocking client to be initialized unnecessarily:

https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/lib/queue.js#L1140-L1147

If `bclient` hasn't been initialized yet, then we're connecting for no reason.  

```js
const Redis = require('ioredis');

const queue = new Queue('foo', {
  createClient (type) {
    if (type === 'bclient') {
      // enforce that we don't waste connections for read-only queues
      throw new Error('This queue is read-only');
    }

    return new Redis();
  }
});
queue.whenCurrentJobsFinished();
```

## Bull version
`develop` branch

## Additional information
Also reproducible when calling `queue.pause(true)` since that calls `whenCurrentJobsFinished` internally. Workaround for that is to call `queue.pause(true, true)` so that `whenCurrentJobsFinished` is not called. But that only works if you're sure there are no jobs to wait for.

gabegorelick:
Fixes #1346

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -279,6 +279,12 @@ function redisClientGetter(queue, options, initCallback) {
         return connections[type];
       }
       const client = (connections[type] = createClient(type, options.redis));
+
+      // Since connections are lazily initialized, we can't check queue.client
+      // without initializing a connection. So expose a boolean we can safely
+      // query.
+      queue[type + 'Initialized'] = true;
+
       if (!options.createClient) {
         queue.clients.push(client);
       }
@@ -1139,6 +1145,11 @@ Queue.prototype.clean = function(grace, type, limit) {
  */
 Queue.prototype.whenCurrentJobsFinished = function() {
   return new Promise((resolve, reject) => {
+    if (!this.bclientInitialized) {
+      // bclient not yet initialized, so no jobs to wait for
+      return resolve();
+    }
+
     //
     // Force reconnection of blocking connection to abort blocking redis call immediately.
     //
--- a/test/test_connection.js
+++ b/test/test_connection.js
@@ -98,8 +98,15 @@ describe('connection', () => {
 
     const testQueue = utils.buildQueue('external connections', opts);
 
-    return testQueue
-      .isReady()
+    return new Promise(resolve => {
+      if (subscriber.status === 'ready') {
+        return resolve();
+      }
+      subscriber.once('ready', resolve);
+    })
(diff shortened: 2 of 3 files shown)


lib/queue.js
```diff
@@ -279,6 +279,12 @@ function redisClientGetter(queue, options, initCallback) {
         return connections[type];
       }
       const client = (connections[type] = createClient(type, options.redis));
+
+      // Since connections are lazily initialized, we can't check queue.client
+      // without initializing a connection. So expose a boolean we can safely
+      // query.
+      queue[type + 'Initialized'] = true;
```
manast: Couldn't you store this information in a separate data structure? furthermore, what other queue types would benefit from storing this data?

gabegorelick: > Couldn't you store this information in a separate data structure?

Not sure I follow. Do you mean you want a dedicated object on the queue to save this? Something like
```js
queue._clientStatuses = {}; // in Queue constructor

queue._clientStatuses[type] = 'initialized'; // after initialization
```

Or are you thinking there's a more complex state machine that should be used?

> what other queue types would benefit from storing this data?

Do you mean Redis connection types? There may be instances where `client` and `eclient` are also initialized unnecessarily, but I haven't tracked them down yet. That's also blocked by #1344 which means that `client` is always initialized.

gabegorelick:
The failing test appears to be an existing race condition:

https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/test/test_connection.js#L82-L117

`isReady` only waits for `client` to be ready, not the `subscriber` client. `subscriber` client is initialized in the `Queue` constructor when we call `this.on('error')`, so it's a race condition whether it's ready or not by the time the test gets to that assertion. I don't see anything in my PR that would change this.

lib/queue.js
gabegorelick: @manast Have you had a chance to look at this?

manast: yes, sorry. I meant a separate data structure.

mgreenw:
This is awesome, thanks for fixing this @gabegorelick!

Issue #883, opened 2018-03-07
Repeatable job's delay option doesn't work. Is it a bug or I didn't have correct config?


MacIt:
## Description
I'm using bull to do repeatable jobs, and if not receiving success message from client within 1 min then this job should be executed again.
This logic works, but all repeatable jobs have more delays (around 50 seconds) although I specified the _delay_ option for every job, i.e. _delay: 2_.

The job delay will be 0 if run it as non-repeatable, even if the _delay_ option was given 10.

I noticed "Note that for accurate delays, both server and clients should have their clocks synchronized." in the bull document. But don't really understand which server and client they are. I'm testing on a Mac OS High Sierra 10.13.3. The client and server are both local machine. The only difference is that Redis sits in another machine in the same network. Redis server has almost the same time as Mac.

Also tried to config timezone as CST where I'm locating for repeatable jobs. My Mac and redis server are in the same timezone. But it didn't make any difference.
 
## Test code to reproduce
Here is code:
`
deliveryQueue.add('newDelivery', {name: 'lala'}, {repeat: {cron: '*/1 * * *'}, tz: 'CST', delay: 10})
`
## Node version: 8.9.4

Thanks a lot in advance if anyone give hints.

![img_2791](https://user-images.githubusercontent.com/1609994/37094929-71b76d4a-224f-11e8-97bc-eb9f0bc16adb.PNG)

manast:
Actually, repeatable jobs use internally the "delay" option in order to schedule when to be processed. Therefore you cannot mix the delay option and the repeat option, I will raise an error to make this behaviour more clear.

MacIt:
Too busy to come back for updates.... Many thanks to @manast for your clarification. Just wonder if it is possible to let repeatable jobs accept delay option from configuration? Maybe it takes more time. Anyway I know the right usage of delay option now, so happy working with bull.

ancmikesmullin:
it would be nice if delay could be use with repeat to introduce an initial stagger/offset, for the case where you have many jobs being enqueued at the same time with repeatable schedules, and perhaps they are all set to the same interval, but you want them evenly spread over that interval, so you could then specify a `delay: Math.random()*interval` to act as a delay that is only respected the first time a job is scheduled, and the interval begins after that.

as a workaround, i guess i could enqueue a stagger job using the delay, and then a scheduled job using repeat after that.

PR #1363, opened 2019-06-25
add support for a staggered interval

ancmikesmullin:
it would be nice to introduce an initial stagger/offset, for the case where you have many jobs being enqueued at the same time with repeatable schedules, and perhaps they are all set to the same interval, but you want them evenly spread over that interval, so you could then specify a `stagger: Math.random()*interval` to act as a delay that will offset when jobs of similar intervals get started, so they don't all start at the same time every time.

see also: https://github.com/OptimalBits/bull/issues/883#issuecomment-505552546

my use case is, i have 50 jobs running on a one minute interval, and without this change, the consumer's CPU utilization spikes at the beginning of every minute, and is idle after that. not as efficient.

### DIFF
--- a/lib/repeatable.js
+++ b/lib/repeatable.js
@@ -16,6 +16,7 @@ module.exports = function(Queue) {
     const client = this.client;
     const repeat = opts.repeat;
     const prevMillis = opts.prevMillis || 0;
+    const stagger = opts.stagger || 0;
 
     if (!prevMillis && opts.jobId) {
       repeat.jobId = opts.jobId;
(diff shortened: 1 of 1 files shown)


manast:
maybe it would be better to call it "jitter", which is a more established term in this context. However, before being able to merge this I would need a couple of unit tests that demonstrates it works as expected.

ancmikesmullin:
i considered jitter, but this has a more specific meaning which implies randomness. in my case, i let the user define the strategy, so it could be random, or it could be perfectly evenly spaced, or the first two overlapping, or anything. so i like stagger.

i will see if i can learn how the unit tests are working and make one.

ancmikesmullin:
ended up on custom bull queue implementation.

PR #1365, opened 2019-06-28
fix: expose getNextMillis as getNextRepeatableTime

jaulz:
Expose function so it's possible to know when the next execution of a job is planned.


manast:
This should not be needed, you can use ```queue##getRepeatableJobs``` to get all the repeatable jobs and their next execution time.

manast:
If somebody needs this function, in order to merge it we will need 1) a test case, 2) update the REFERENCE.md

PR #1370, opened 2019-07-04
Fixed REFERENCE.md faulty Job#log description

alexandrugheorghe:
Removed copied arguments section from Job#log


manast:
thanks!.

Issue #1357, opened 2019-06-20
Ability to limit size of job data


gabegorelick:
## Feature request

Many queues have upper limits for the size of messages that you can send through them. For example, SQS limits messages to [256 KB](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html).

Such limits are really important for a queue backed by Redis, since memory may be limited: you don't want someone enqueuing 1GB jobs!

Bull should check the size of `jobData` and return an error if it's larger than a configurable limit:
https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/lib/job.js#L61-L64

Since it has to serialize job data to a string anyway, Bull should be able to approximate the size a job would occupy in Redis by checking the length of the serialized string:
https://github.com/OptimalBits/bull/blob/c0ee6be03c6d47d60465d2bae0d2566d5bf0d661/lib/job.js#L132-L135

joshterrill:
I can give this a shot if that's okay. Seems to be a fairly simple introduction into getting my feet wet with contributing to the repo.

joshterrill:
@gabegorelick Would you like the default size to be set to 256kb in the `setDefaultOpts` method?

gabegorelick:
> Would you like the default size to be set to 256kb in the setDefaultOpts method

256 KB is fine by me, but wouldn't setting that as a default be a breaking change?

joshterrill:
Yeah I was just thinking about that. So I won't have it set a default, but I'll check for the existence of the sizeLimit, if it exists and the data goes over the limit, I can throw an exception.

joshterrill:
@gabegorelick I can either let the user define the `sizeLimit` in bytes or I can implement a small library like https://www.npmjs.com/package/bytes and create a utility method which will let the user either enter a number in bytes, or a string such as `1mb` or `1gb` and that library will automatically convert it to bytes. I don't know what the rule is for adding extra dependencies, but this one is pretty small - if you want to look at the source code for it, it's here: https://github.com/visionmedia/bytes.js/blob/master/index.js

gabegorelick:
> I can either let the user define the sizeLimit in bytes or I can implement a small library like https://www.npmjs.com/package/bytes

I ultimately will defer to @manast and other core Bull contributors, but I would just accept a Number of bytes. If users want to use a third party library so that they can declare it as a string that's fine, but then they can bring their own preferred lib.

PR #1378, opened 2019-07-09
feat: added sizeLimit option when creating new jobs. issue #1357

joshterrill:
This is my first PR for this repo, I thought this issue would be a good one to get started on. This resolves issue #1357 and adds an optional parameter for sending through a `sizeLimit` parameter on `opts` when creating a job. If the data you're sending through is larger than the `sizeLimit` set, an exception is thrown.

A couple of notes:

1. I added a dependency called `bytes` which is a small dependency which allows a user to type a  "human readable" `sizeLimit` such as `1b`, `1mb`, `1gb`, etc. instead of having to convert everything to bytes themselves. I also added a utility function so that a user can either set the limit as a number in bytes, or the human readable format. If you don't want this other dependency there, let me know and I can make it so that it'll just take the number in bytes that a user types in.

2. I implemented a utility function that I saw on this [stack overflow post](https://stackoverflow.com/a/23318053/1347170) that describes how to count the size of strings, which also include non-ascii characters. Just to be safe, and to support other character encodings, I use this method to calculate the size.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -53,10 +53,6 @@ function setDefaultOpts(opts) {
   _opts.attempts = parseInt(_opts.attempts);
   _opts.backoff = backoffs.normalize(_opts.backoff);
 
-  if (_opts.sizeLimit) {
-    _opts.sizeLimit = utils.returnLimitAsBytes(_opts.sizeLimit);
-  }
-
   return _opts;
 }
 
@@ -142,13 +138,7 @@ Job.prototype.toData = function() {
     this.opts.sizeLimit &&
     utils.lengthInUtf8Bytes(json.data) > this.opts.sizeLimit
   ) {
-    throw new Error(
-      'The size of job ' +
-        (this.opts.jobId || this.name || Job.DEFAULT_JOB_NAME) +
-        ' exceeds the limit ' +
-        this.opts.sizeLimit +
-        ' bytes'
-    );
+    throw new Error(`The size of job ${this.opts.jobId || this.name || Job.DEFAULT_JOB_NAME} exceeds the limit ${this.opts.sizeLimit} bytes`);
   }
 
   json.opts = JSON.stringify(json.opts);
--- a/lib/job.js
+++ b/lib/job.js
@@ -53,6 +53,10 @@ function setDefaultOpts(opts) {
   _opts.attempts = parseInt(_opts.attempts);
   _opts.backoff = backoffs.normalize(_opts.backoff);
 
+  if (_opts.sizeLimit) {
+    _opts.sizeLimit = utils.returnLimitAsBytes(_opts.sizeLimit);
+  }
+
   return _opts;
 }
 
@@ -133,6 +137,20 @@ Job.prototype.toData = function() {
   const json = this.toJSON();
 
   json.data = JSON.stringify(json.data);
+
+  if (
+    this.opts.sizeLimit &&
+    utils.lengthInUtf8Bytes(json.data) > this.opts.sizeLimit
+  ) {
+    throw new Error(
+      'The size of job ' +
+        (this.opts.jobId || this.name || Job.DEFAULT_JOB_NAME) +
+        ' exceeds the limit ' +
+        this.opts.sizeLimit +
+        ' bytes'
+    );
+  }
+
   json.opts = JSON.stringify(json.opts);
   json.stacktrace = JSON.stringify(json.stacktrace);
   json.failedReason = JSON.stringify(json.failedReason);
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,4 +1,5 @@
 'use strict';
+const bytes = require('bytes');
 const errorObject = { value: null };
 function tryCatch(fn, ctx, args) {
   try {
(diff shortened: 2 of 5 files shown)


lib/job.js
```diff
@@ -53,6 +53,10 @@ function setDefaultOpts(opts) {
   _opts.attempts = parseInt(_opts.attempts);
   _opts.backoff = backoffs.normalize(_opts.backoff);
 
+  if (_opts.sizeLimit) {
```
manast: I think it is better to keep this option as a Queue option, so that all jobs added to that particular instance of a queue will need to honor the size.

lib/job.js
```diff
@@ -133,6 +137,20 @@ Job.prototype.toData = function() {
   const json = this.toJSON();
 
   json.data = JSON.stringify(json.data);
+
+  if (
+    this.opts.sizeLimit &&
+    utils.lengthInUtf8Bytes(json.data) > this.opts.sizeLimit
+  ) {
+    throw new Error(
+      'The size of job ' +
```
manast: use es6 string interpolation

lib/utils.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+const bytes = require('bytes');
```
manast: although it is nice to be able to add with human readable sizes, I think this can be used by the user on top of Bull, just to keep the dependencies as low as possible.

lib/utils.js
```diff
@@ -43,7 +44,29 @@ function isRedisReady(client) {
   });
 }
 
+/**
+ * Checks the size of string for ascii/non-ascii characters
+ * (Reference: https://stackoverflow.com/a/23318053/1347170)
+ * @param {string} str
+ */
+function lengthInUtf8Bytes(str) {
+  const m = encodeURIComponent(str).match(/%[89ABab]/g);
```
manast: is this better/faster than using ```Buffer.byteLength``` ?
https://stackoverflow.com/questions/9864662/how-to-get-the-string-length-in-bytes-in-nodejs

lib/utils.js
joshterrill: Sounds good. So I'll remove this dependency and if the user wants to add something like that to pass it in, they can.

lib/utils.js
joshterrill: I'm not sure, I'll do some tests.

lib/job.js
joshterrill: I was going to use string interpolation, but I saw several other places such as 

```javascript
throw new Error('Could not remove job ' + job.id);
```

Which didn't, so I wanted to be consistent. Should I update all of the other places as well, or do you want me to just update this one place?

lib/job.js
joshterrill: Sounds good, I didn't realize that was how it works, I'll make the change.

joshterrill:
@manast I made a few of the changes now. I'll work on getting the last change in soon. Thanks for the feedback.

package-lock.json
```diff
@@ -121,6 +121,14 @@
       "dev": true,
       "requires": {
         "semver": "5.6.0"
+      },
```
gabegorelick: @manast This is another example of why I think `package-lock.json` should not be checked in. Unless I'm mistaken, there's no reason it should be updated as part of this PR, and yet `npm` loves to make changes to it.

lib/job.js
```diff
@@ -133,6 +133,14 @@ Job.prototype.toData = function() {
   const json = this.toJSON();
 
   json.data = JSON.stringify(json.data);
+
+  if (
+    this.opts.sizeLimit &&
+    utils.lengthInUtf8Bytes(json.data) > this.opts.sizeLimit
+  ) {
+    throw new Error(`The size of job ${this.opts.jobId || this.name || Job.DEFAULT_JOB_NAME} exceeds the limit ${this.opts.sizeLimit} bytes`);
```
gabegorelick: Since toData can throw now, are there any other callers of `toData` that need to handle this?

gabegorelick: Also, `toData` seems like a weird place to check this. Wouldn't `addJob` be a better place?

lib/job.js
manast: yes, that is because we just upgraded the minimum required nodejs version

package-lock.json
manast: yes, for libraries it should not be checked-in, I will remove it.

lib/job.js
manast: but even if it is called from other places, isn't it good that it throws if it exceeds the maxium size?

gabegorelick: I don't think so. If I want to serialize a job to JSON, I don't want that to throw if it's too big. I only care about it being too big when I add the job to the queue.

joshterrill: I can make this change, thanks for the feedback.

package-lock.json
joshterrill: Sounds good. Since the lib ended up getting removed anyway, I can revert the change and commit that with my next commit so that `package-lock.json` remains unchanged for this PR.

Issue #733, opened 2017-10-16
Priority doesn't work ?


jbeaudoin11:
Hi !

I'm having a hard time to understand how priorities work with the queue. I'm making a multimedia processing server, basically compression and resize stuff. So I have 3 queue : image, audio and video, but i'm focusing my test only on the image one for now.

The image queue has 2 types of job (for now), ResizeImage (priority: 1, high priority) and CompressImage (priority: 10, process "after" ResizeImage). I'm doing batch jobs so for each image those 2 jobs are added to the queue. I changed the order of how jobs are added to the queue and now it's by priority so for this case, every ResizeImage jobs are added before all CompressImage jobs.

So what i'm expecting to happen is that most of ResizeImage jobs are done first and after CompressImage jobs should be starting. But it just doesn't work like that at all :

```
ResizeImage 1629
ResizeImage 1628
ResizeImage 1630
ResizeImage 1631
ResizeImage 1635
ResizeImage 1634
ResizeImage 1633
ResizeImage 1632
CompressImage 1679
CompressImage 1678
CompressImage 1677
CompressImage 1676
CompressImage 1674
CompressImage 1675
CompressImage 1673
CompressImage 1672
CompressImage 1671
CompressImage 1670
CompressImage 1669
CompressImage 1668
CompressImage 1667
CompressImage 1666
CompressImage 1665
CompressImage 1664
CompressImage 1663
CompressImage 1662
CompressImage 1661
CompressImage 1660
CompressImage 1659
CompressImage 1658
CompressImage 1657
CompressImage 1656
ResizeImage 1653
ResizeImage 1652
CompressImage 1655
ResizeImage 1651
ResizeImage 1650
ResizeImage 1649
ResizeImage 1648
ResizeImage 1647
CompressImage 1654
ResizeImage 1646
ResizeImage 1645
ResizeImage 1643
ResizeImage 1644
ResizeImage 1642
ResizeImage 1641
ResizeImage 1640
ResizeImage 1639
ResizeImage 1638
ResizeImage 1637
ResizeImage 1636
``` 
This is base on the queue.on("completed", ..) event and with 4 workers. So i don't know how reliable that log is.

I don't think it matters, but that is a small sample size for fast testing, we will get prob 1000+ images by batches.. Also my nodejs server is on a VM with vagrant/virtualbox on ubuntu 16.04.

manast:
[...]

PR #1405, opened 2019-07-31
fix(clean-priority): remove job from priority set on clean

bdentino:
First contribution, so apologies if I've missed anything or stepped on any project standards - happy to make updates to the PR as necessary.

I noticed some strange behavior in one of my applications where jobs were not being processed according to priority. I could reproduce reliably in production, but not in a new staging environment so I dug into the repo and checked some of the keys in Redis. Long story short, I found out that whenever I run the `cleanJobsInSet` script on the waiting jobs it causes the priority set to go wildly out of sync with the jobs list because these jobs aren't being removed, and I believe this was causing new jobs to get inserted into the jobs list at strange indexes. I have a script running in production that periodically cleans up the different states, and once I re-synced by resetting the priority key and removing all jobs the problem was gone. 

This fix simply updates the `cleanJobsInSet` script to remove each job from the priority set if present. 

I don't know when the clean feature was introduced, but I wonder if this could have been causing the OPs issues in #733?


manast:
thanks for the patch, It looks great. However I think you may have also added the ```addMany```  feature in the same PR, it is better to keep them as 2 different.

bdentino:
You are absolutely correct - my mistake! I've removed those commits from this branch. 

On that note, would you consider accepting a PR for the `addMany` feature? I have some scripts that schedule 100k+ jobs and it has significantly improved the performance of those scripts, but it doesn't seem to be an issue any other users are running into.

luastoned:
I encountered the same bug, when will this fix be released?

bdentino:
ping @manast - any interest in merging this still?

manast:
@bdentino looks good, very sorry that I missed the reply to my comment :/

Issue #408, opened 2016-12-07
Can jobs be added in bulk?


brandonmp:
I have a process which generates 100k+ jobs for my queue, and adding them 1-at-a-time is a lot of i/o overhead. this snip:

```
 for (let i = 0; i < 10000; i++) {
        await testQ.add({ taskID: i, taskMeta: "this is metadata 1" }, { removeOnComplete: true })
    }
```

took several minutes to complete with a `redis` server on `redislabs.com`

Is there a way to construct the jobs locally and add to `redis` in bulk?

manast:
There is no way to do this with bull, but should not be very difficult to implement as an addBulk operation, where an array of jobs is sent to redis in one atomic operation, that would increase the performance in at least one order of magnitude:
https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L73

n3trino:
@manast any plan to implement this? I think it would be very helpful addition.

stonecold123:
@manast @n3trino 
Any news? I'd love to have this feature.
Currently my workaround is using https://github.com/palanik/forloop to make the loop non blocking. Performance isn't bad. like this:

````
forloop(1, 20000000, 10, (i => {
          queueItemUpdate.add(i, {delay: 3000, removeOnComplete: true, removeOnFail: true});
     })
);
````

**Comment by manast** [comment]
you do not need that library in this case, since ```add``` is already an asynchronous function. On the other hand you may want to store the resulting promises to know when you are actually done adding jobs, something like this:
```js
var adds = [];
for(var i=0; i < NUM_JOBS;i++){
  adds.push(queueUpdate.add(i, {delay: 3000, removeOnComplete: true, removeOnFail: true});
}

Promise.all(adds).then(function(){
  console.log('done');
});
```

This issue is about having a bulk method so that we can save roundtrips to redis.

jrr:
Another reason for this is to be able to transactionally enqueue a bunch of jobs.

Igor-Lopes:
Hello, are there any updates on this issue ?

Igor-Lopes:
> There is no way to do this with bull, but should not be very difficult to implement as an addBulk operation, where an array of jobs is sent to redis in one atomic operation, that would increase the performance in at least one order of magnitude:
> https://github.com/OptimalBits/bull/blob/master/lib/scripts.js#L73

[...]

PR #1419, opened 2019-08-06
Support for bulk adding jobs

holm:
This is a sketch for adding support for posting multiple jobs at a time to the queue. I have added basic tests, but could add a few more to validate defaulting of options and names.

I wanted to check first that you think this is the right approach. This seems simpler than trying to modify the lua script, and should be the same performance.

Fixes https://github.com/OptimalBits/bull/issues/408


manast:
Good job!

Igor-Lopes:
@manast , when can we have it published in npm?

PR #1422, opened 2019-08-10
feat: add immediately option for every repeatable job

### LINKED ISSUE #1239: [Question] How to run a repeatable job immediate?
Agenda has a option to skip the immediate run:
https://github.com/agenda/agenda#manually-working-with-a-job

`options.skipImmediate`: `true` | `false` (default) Setting this true will skip the immediate run. The first run will occur only in configured interval.

In Bull, a repeatable job will skip the immediate run by default. How to not skip the immediate run?

HugoPoi:
Try to resolve #1239 

* `immediately` not implemented with `cron`
* my changes in the `createNextJob` function are not pretty, maybe add a `delete opts.immediately` should be better because with the present code it will always be saved in redis ?
* My unit test doesn't cover lot of cases


Eywek:
Any news on this?

HugoPoi:
I need someone to review it. And it might be useful to have #1498 with this.

lib/repeatable.js
```diff
@@ -58,7 +58,9 @@ module.exports = function(Queue) {
               _.defaultsDeep(
                 {
                   repeat: {
-                    count: currentCount
+                    count: currentCount,
+                    // Do NOT save immediately flag to true
```
manast: why is this needed?

HugoPoi: I re-read my code, and like I suggest, I need to replace this by a `delete opts.repeat.immediately;` after the getNextMillis call.

lib/repeatable.js
```diff
@@ -31,6 +31,7 @@ module.exports = function(Queue) {
     now = prevMillis < now ? now : prevMillis;
 
     const nextMillis = getNextMillis(now, repeat);
+    delete repeat.immediately;
```
manast: I am not sure about this. Why do we need to delete it?

HugoPoi: Because it will be re-used by the next job and that's not what I want. When the first job have been completed `nextJobFromJobData` is called but we need to plan the next job to the next interval not immediately.

manast: ok, but we should not mutate properties from the input object, that can potentially create side effects on the callers code

HugoPoi: OK so my first solution was better ? If we don't want to mutate so we need to add a separate flag on the prototype of `nextRepeatableJob` and all over the code ?

manast: I do not remember the previous solution :/.

HugoPoi: Just the one with saving immediately flag in the job object, before my refactor commit.

oskarstrasburger:
Any news @manast ?

faller:
Any news?

manast:
I would merge this but my last comment needs to be addressed!

HugoPoi:
Ok so I need to refactor, passing the flag directly in `opts.immediately` because it will need to be handle only at the first iteration and not saved in Redis. And no mutation, so no delete, no assignment with `=`.

joaozitopolo:
Same problem here. I need to skip the first execution.
If the object "job" on process have a execution count, or a firstRun flag, we can do the skip in the process function.

santiq:
@HugoPoi hey man I know it's been a year but could you do those changes? I'll be happy to help.

HugoPoi:
I have rebase the change on develop, an immediately flag will be saved in redis is it bad ? I think not.
If we don't want to save the `immediately` flag in db, need a change to the `deepDefault` to omit the prop.

manast:
If somebody wants to fix this PR it is just a matter of backporting this PR from BullMQ: https://github.com/taskforcesh/bullmq/pull/536

PR #1428, opened 2019-08-15
getStalledJobs

NassiHarel:


manast:
the commit history for this PR is too polluted, also we are following https://www.conventionalcommits.org/en/v1.0.0-beta.4/ now.
Finally, we upgrade the package.json before the release (in this case it would be a minor update since it is a new feature not a fix).

manast:
what I mean is that you do not need the upgrade the package version.

NassiHarel:
Yes you are right, it's ours CI

PR #1431, opened 2019-08-19
Support for get stalled jobs

NassiHarel:
probably there is a better lua way to do it...


lib/commands/getStalledJobs-7.lua
```diff
@@ -0,0 +1,65 @@
...
+          -- Do not Move the job back to the wait queue, so it wouldn't be picked up by a waiting worker.
+          table.insert(stalled, jobId)
+        end
+      end
+    end
+  end
+end
+
+-- Mark potentially stalled jobs
+local active = rcall('LRANGE', KEYS[3], 0, -1)
```
manast: since this function only should return stalled jobs, it should not update the stalled set.

lib/commands/getStalledJobs-7.lua
```diff
@@ -0,0 +1,65 @@
...
+  local MAX_STALLED_JOB_COUNT = tonumber(ARGV[1])
+
+  -- Remove from active list
+  for i, jobId in ipairs(stalling) do
+    local jobKey = ARGV[2] .. jobId
+
+    -- Check that the lock is also missing, then we can handle this job as really stalled.
+    if(rcall("EXISTS", jobKey .. ":lock") == 0) then
+      --  Remove from the active queue.
+      local removed = rcall("LREM", KEYS[3], 1, jobId)
```
manast: we should not remove jobs from active queue if we just want to return stalled jobs.

NassiHarel:
@manast since I'm not familiar with LUA, I need to learn it first :relaxed:
 Thanks.

ccollie:
If **manast**'s comments are the extent of changes required, I can cleanup the PR, since I have a bit of experience dealing with LUA scripting in redis.

stansv:
@ccollie any help would be appreciated, thank you

ccollie:
Having a look. Will update in a day or so.

NassiHarel:
@ccollie any news with this PR?

ccollie:
@NassiHarel sorry about that, things are a bit hectic here. I took a look and I stopped partly because it seems like the code is mixing different concerns and I wasn't sure how to untangle it. If you want to chat let's take it offline, but keep in mind that the latest version of Bull is progressing well and it may be worth it to concentrate the effort in the new repository.

manast:
gonna close this since nobody available to take care of it, reopen if somebody can work on it.

PR #1440, opened 2019-08-24
new interface job.syncData 

okamuuu:
I use `OptimalBits/bull`  very useful library. My currency project required to store some data in the processing job queue. So I often write codes following.

```
job.data.foo = "bar"
job.update(job.data)
```
But I prefer following.

```
job.data.foo = "bar"
job.syncData()
```

As a matter of fact, the function is not essential, but I'm glad to consider this PR.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -35,6 +35,7 @@
   - [Job#log](#joblog)
   - [Job#getState](#jobgetstate)
   - [Job#update](#jobupdate)
+  - [Job#asyncData](#jobasyncdata)
   - [Job#remove](#jobremove)
   - [Job#retry](#jobretry)
   - [Job#discard](#jobdiscard)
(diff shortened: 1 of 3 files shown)


stansv:
Can you please explain your case a bit? I believe the job data should be immutable once you push job into a queue. If you need to store some intermediate job results inside job processing callback, do not mutate job data, use progress() instead. If you really need to update job data, and you are sure no workers will take it to process in the meantime, there is no problem to call update().

manast:
I appreciate the effort for this PR but this is just sugar for current library, and I think it is better to keep it as clean as possible from redundant code.

okamuuu:
Originally, I expected the following behavior to update function.  I wanted each way are the same result.

```
// way 1
job.data.title = 'a very different title now'
job.update().then(() => {})

// way 2
job.update({
  title: 'a very different title now'
}).then(() => {})
```
Unfortunately, the job.update() function save undefined into redis-server. I was confused by this behavior. So I tried to create this PR.

Anyway, I agree with the lines  `is better to keep it as clean as possible from redundant code.`

Thank you for considering:)

PR #1451, opened 2019-09-02
Adds bull-board to list of UIs

joaomilho:


joaomilho:
How can coverage decrease!?

ngouy:
should we also add that to the bull-mq list ?

image
image

PR #1461, opened 2019-09-09
Fix-[Sandbox] If the child process is killed, remove it from the pool.

### LINKED ISSUE #1460: When the OS closes a child process, causes an ERR_IPC_CHANNEL_CLOSED exception, causing the master process to fail.
## Description

If the OS (or anyone else) closes the child process, instead of checking its `signalCode`, the child is released in the pool. This causes an error that leads, when trying to send that child another message, to fail, causing an `uncaughtException` that closes the main process.

Related error and detais: [here](https://sentry.valiu.co/share/issue/e19395ad587244e78bcff78eb61b0d2f/)

Create a project using these two files, that depends on Bull@3.10.0

`index.js`
```javascript
'use strict';

const { resolve } = require('path');
const Bull = require('bull');

async function main() {
    const queue = new Bull('myqueue', {
        redis: 'redis://localhost:6379',
        settings: {
            lockDuration: 100000, // Key expiration time for job locks.
            stalledInterval: 100000, // How often check for stalled jobs (use 0 for never checking).
            maxStalledCount: 10, // Max amount of times a stalled job will be re-processed.
            guardInterval: 1000, // Poll interval for delayed jobs and added jobs.
            retryProcessDelay: 5000, // delay before processing next job in case of internal error.
            drainDelay: 20
        }
    });

    try {
        queue.process(resolve(__dirname, './process/index.js'));
        await queue.add({ hello: 'world' });
        await queue.add({ hello: 'world' });
    } catch (error) {
        throw error;
    }
}

main();
```

`process/index.js`
```javascript
'use strict';

const setTimeoutPromise = timeout => new Promise(resolve => setTimeout(() => resolve(), timeout)); 

module.exports = async function () {
    try {
        console.log('opening /my/little/pony.txt');
        await setTimeoutPromise(100000);
        console.log('/my/little/pony.txt opened');
    } catch (error) {
        console.error(error);
    }
};
```

Open the main process. Run:

```bash
node index.js
```

Kill the child process. Run:

```bash
ps -ax | grep master.js | awk '{print $1}' | xargs -I % kill -9 %
```

## Bull version

3.10.0

## Additional information

N/A

pandres95:
Closes #1460

### DIFF
--- a/lib/process/sandbox.js
+++ b/lib/process/sandbox.js
@@ -52,7 +52,7 @@ module.exports = function(processFile, childPool) {
         child.removeListener('message', msgHandler);
         child.removeListener('exit', exitHandler);
 
-        if (child.exitCode !== null) {
+        if (child.exitCode !== null || /SIG.*/.test(child.signalCode)) {
           childPool.remove(child);
         } else {
           childPool.release(child);


manast:
looks good, thanks.

PR #1468, opened 2019-09-12
Improve documentation for getJobs()

### LINKED ISSUE #1454: getJobs() start and end options are applied per job state
Don't know if this is intended behaviour, but I noticed that when using `getJobs()` with multiple states, the optional `start` and `end` parameters to skip/limit the results seem applied **per job state**. So, given this:

`getJobs(['active', 'completed'], 0, 1)`

and having, say 10 active and 10 completed jobs in the queue, the call yields 4 jobs: The first 2 jobs with state `active`, and the first two jobs with state `completed`. I'd have expected to just get two results.

If this is indeed intended, I suggest updating the reference manual to point this out, because it took me some time to figure out what's going on there. 😄

soulchild:
Clarify use of `start` and `end` options. Fixes #1454

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -414,7 +414,7 @@ getJobs(types: string[], start?: number, end?: number, asc?: boolean): Promise<J
 
 Returns a promise that will return an array of job instances of the given types. Optional parameters for range and ordering are provided. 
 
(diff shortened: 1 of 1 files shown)


manast:
thanks!

PR #1475, opened 2019-09-17
Add getPuased() function to getters

fbukevin:
I notice that **bull** still provides function `Queue.pause()` to get tasks into state **paused**. But when I trying to look up tasks in a queue with **bull-arena**, I got an error: 
```
WRONGTYPE Operation against a key holding the wrong kind of value
```

I check out [this PR](https://github.com/bee-queue/arena/pull/133) in *bull-arena*. The author modified in `src/server/views/dashboard/index.js`. However, after I modified, the new error occurred:
```
UnhandledPromiseRejectionWarning: TypeError: queue[(("get" + (intermediate value)) + "")] is not a function
```
**bull-arena** cannot map getPaused to function in **bull**.

### DIFF
--- a/lib/getters.js
+++ b/lib/getters.js
@@ -115,6 +115,10 @@ module.exports = function(Queue) {
     return this.getJobs('failed', start, end, false);
   };
 
+  Queue.prototype.getPaused = function(start, end) {
+    return this.getJobs('paused', start, end, false);
+  };
+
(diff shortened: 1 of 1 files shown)


manast:
but the thing is that there are no paused state for jobs in Bull, you can either pause a worker or you can pause the whole queue. So this is probably an issue in Arena.

manast:
closing since it works as designed

PR #1489, opened 2019-09-27
:sparkles: support custom async backoff

akirilyuk:
This PR introduces a support for async backoff strategies. Sometimes we need to make calls to some external apis etc to determine if a job should be restarted or not, for example checking if the file was already processed etc.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -237,7 +237,6 @@ Job.prototype.discard = function() {
  */
 Job.prototype.moveToFailed = function(err, ignoreLock) {
   this.failedReason = err.message;
-  // comment for travis build
   return new Promise(async (resolve, reject) => {
     let command;
     const multi = this.queue.client.multi();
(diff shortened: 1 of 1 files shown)


akirilyuk:
Hi tested it locally and the tests passs with node 8.16.1. actually don't know whats the issue is here :(

manast:
thanks. Gonna need a test before this can be merged. Btw, I am surprised this does not break any existing code, if the backoffs.calculate does not return a promise this should lead to a crash, or maybe I am missing something :/.

stansv:
As I know if you try to use await for any non-promise value, it would be wrapped with Promise.resolve(). Also, if you awaiting for a chain of promises (i.e. Promise resolves to next Promise and so on), the engine will add a loop to resolve until value returned/error is thrown. The single drawback is that some another code can be executed in between, once await is reached, and mutate the state, even if return value is available right away.

akirilyuk:
@manast created a new PR #1490

PR #1490, opened 2019-09-29
feat: support async custom backoffs

akirilyuk:


akirilyuk:
@manast What do you think should happen if a promise is rejected? Return -1? Or should it do nothing and the user will be responsible to have proper error handling?

akirilyuk:
@manast any update on this one?

akirilyuk:
@manast ?

PR #1498, opened 2019-10-08
feat: exactRecurrence to repeat job regardless of the interval of every

mabudari:
Adding `exactRecurrence` flag to repeatable job configuration, giving Bull the capability of running recurring jobs referenced to the time the job was added (hours, minutes, seconds).


oskarstrasburger:
It have to be added asap. I don't want to run tasks only in full hours.

manast:
sorry for late response I will try to merge this ASAP.

oskarstrasburger:
@manast up

PR #1533, opened 2019-11-01
test: fix mocha options

gabegorelick:
Having `test/test_*` in `mocha.opts` meant that it was not possible to run an individual test in isolation.

This may have been caused by the recent mocha version upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   "scripts": {
     "lint": "eslint lib test",
     "pretest": "npm run lint",
-    "test": "NODE_ENV=test mocha --exit",
+    "test": "NODE_ENV=test mocha 'test/test_*'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
   "scripts": {
     "lint": "eslint lib test",
     "pretest": "npm run lint",
-    "test": "NODE_ENV=test mocha --exit",
+    "test": "NODE_ENV=test mocha 'test/test_*'",
```
gabegorelick: Anyone running `npm test`, including CI, will still run all tests. But you can now also run `mocha test/path_to_file.js` to run a single test.

test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 --reporter spec
 --timeout 5000
-test/test_*
+--exit
```
gabegorelick: `--exit` is an option we always want, even if not running via `npm test`.

PR #1534, opened 2019-11-01
fix: queue.pause(true, true) doesn't pause queue

### LINKED ISSUE #1532: Passing doNotWaitActive=true to queue.pause means queue won't actually be paused
##  Description

`queue.pause` has an option called `doNotWaitActive`. I believe the intention is that when this is set to `true`, `pause` will not wait for any active jobs to finish before resolving. But I would still expect no new jobs to be processed. 

Internally, `doNotWaitActive=true` means `pause` will not call `whenCurrentJobsFinished`. But `whenCurrentJobsFinished` is where we stop polling for new jobs: https://github.com/OptimalBits/bull/blob/146bc8ed4341ec3e54b299c22e60dc3267d8b6d1/lib/queue.js#L1176-L1178

Thus, without calling  `whenCurrentJobsFinished`, the queue isn't really paused: a job added after `queue.pause(true, true)` will be processed.

```js
queue.process(job => { throw new Error('Job should not be processed'); });

// I don't want to process any more jobs,
// but I want to let any active jobs finish without waiting for them
await queue.pause(true, true);

await queue.add('foo', {});
```

## Bull version

3.11.0

gabegorelick:
Fixes #1532

### DIFF
--- a/test/test_pause.js
+++ b/test/test_pause.js
@@ -15,343 +15,390 @@ describe('.pause', () => {
     return client.flushdb();
   });
 
-  afterEach(() => {
+  afterEach(async () => {
     sinon.restore();
+    await utils.cleanupQueues();
+    await client.flushdb();
     return client.quit();
   });
 
-  it('should pause a queue until resumed', () => {
-    let ispaused = false,
-      counter = 2;
+  describe('globally', () => {
+    it('should pause a queue until resumed', () => {
+      let ispaused = false,
+        counter = 2;
 
-    return utils.newQueue().then(queue => {
-      const resultPromise = new Promise(resolve => {
-        queue.process((job, jobDone) => {
-          expect(ispaused).to.be.eql(false);
-          expect(job.data.foo).to.be.equal('paused');
-          jobDone();
-          counter--;
-          if (counter === 0) {
-            resolve(queue.close());
-          }
-        });
-      });
-
-      return Promise.all([
-        queue
-          .pause()
-          .then(() => {
-            ispaused = true;
-            return queue.add({ foo: 'paused' });
-          })
-          .then(() => {
-            return queue.add({ foo: 'paused' });
-          })
-          .then(() => {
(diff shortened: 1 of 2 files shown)


test/test_pause.js
```diff
@@ -20,338 +20,379 @@ describe('.pause', () => {
     return client.quit();
   });
 
-  it('should pause a queue until resumed', () => {
```
gabegorelick: Most of the changes in this file are from organizing the tests into "global" and "local" sub-suites. Otherwise, the existing tests have not been modified. 

I can back that out if it's not desired.

test/test_pause.js
```diff
@@ -20,338 +20,379 @@ describe('.pause', () => {
...
+          });
         });
       });
     });
-  });
 
-  it('should not processed delayed jobs', function(done) {
-    this.timeout(5000);
-    const queue = new Queue('pause-test');
+    describe('with doNotWaitActive=true', () => {
```
gabegorelick: These tests are new.

test/test_pause.js
```diff
@@ -20,338 +20,379 @@ describe('.pause', () => {
...
+        const jobProcessor = sinon.stub();
+        queue.process(jobProcessor);
+
+        await delay(1000);
+        await queue.pause(true, true);
+
+        // Add another job. This should not be processed
+        await queue.add({});
+
+        await delay(1000);
```
gabegorelick: Let me know if there's a better way to do this other than wait an arbitrary amount of time to see if a job got processed.

gabegorelick:
@manast Tests are passing now. Would appreciate if you could take a look at this.

gabegorelick:
Thank you very much @manast. I appreciate the quick turnaround.

Any plans to publish a new version of bull?

manast:
@gabegorelick can you please ping me with an email here manast@taskforce.sh ? thanks.

PR #1536, opened 2019-11-02
Emulate job.progress function for sandboxed processors

douglascayers:
The [Job.progress](https://github.com/OptimalBits/bull/blob/develop/lib/job.js#L121) function acts as a getter and setter depending on the argument value passed to it.

In the context of a sandboxed job processor (forked process), the job instance given to the processor is actually a [wrapped JSON message](https://github.com/OptimalBits/bull/blob/develop/lib/process/master.js#L102). The `progress` function on the wrapped job object only behaves as an asynchronous setter.

This pull request modifies the `progress` function on the wrapped job object to behave as both a getter and a setter.

### DIFF
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -99,14 +99,41 @@ process.on('uncaughtException', err => {
   throw err;
 });
 
+/**
+ * Enhance the given job argument with some functions
+ * that can be called from the sandboxed job processor.
+ *
+ * Note, the `job` argument is a JSON deserialized message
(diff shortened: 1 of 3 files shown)


manast:
Great job!. @stansv this PR would be great on BullMQ too :).

stansv:
@manast I'll port this tomorrow

douglascayers:
Happy to help! Thanks for the wonderful project and for accepting my PR :)

PR #1538, opened 2019-11-04
chore: lint JS files in root directory

gabegorelick:
Not a huge deal since these files are trivial, but important if they ever grow.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,1 +1,3 @@
+'use strict';
+
 module.exports = { extends: ['@commitlint/config-conventional'] };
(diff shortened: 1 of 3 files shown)


manast:
thanks. Not sure though what is the difference between what we have to day and this change, do you mind to clarify?

gabegorelick:
> Not sure though what is the difference between what we have to day and this change, do you mind to clarify?

@manast Today, `npm run lint` only lints .js files in the `lib` and `test` directory. But there are a few JS files in the root directory. `*.js` is not recursive, so it only globs the 2 JS files in the root directory and not anything in sub-directories.

manast:
ah ok, got it, I was a bit blind earlier today. thanks :).

PR #1542, opened 2019-11-06
chore: cleanup whenCurrentJobsFinished

gabegorelick:
Cleanup and document whenCurrentJobsFinished. And since it's now official API, add some tests for it.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1174,49 +1174,20 @@ Queue.prototype.clean = function(grace, type, limit) {
  * @returns {Promise}
  */
 Queue.prototype.whenCurrentJobsFinished = function() {
-  return new Promise((resolve, reject) => {
-    if (!this.bclientInitialized) {
-      // bclient not yet initialized, so no jobs to wait for
-      return resolve();
-    }
+  if (!this.bclientInitialized) {
+    // bclient not yet initialized, so no jobs to wait for
+    return Promise.resolve();
+  }
 
-    //
-    // Force reconnection of blocking connection to abort blocking redis call immediately.
-    //
-    const forcedReconnection = redisClientDisconnect(this.bclient).then(() => {
-      return this.bclient.connect();
-    });
+  //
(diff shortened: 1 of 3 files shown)


lib/queue.js
```diff
@@ -1174,49 +1174,20 @@ Queue.prototype.clean = function(grace, type, limit) {
  * @returns {Promise}
  */
 Queue.prototype.whenCurrentJobsFinished = function() {
-  return new Promise((resolve, reject) => {
```
gabegorelick: Removing the unnecessary `new Promise`.

lib/queue.js
```diff
@@ -1174,49 +1174,20 @@ Queue.prototype.clean = function(grace, type, limit) {
...
-        return forcedReconnection;
-      })
-      .then(resolve, reject);
-
-    /*
-    this.bclient.disconnect();
-    this.bclient.once('end', function(){
-      console.error('ENDED!');
-      setTimeout(function(){
-        this.bclient.connect();
```
gabegorelick: Removed all this commented code.

lib/queue.js
```diff
@@ -1174,49 +1174,20 @@ Queue.prototype.clean = function(grace, type, limit) {
...
-      });
-      stream.on('error', function(err){
-        console.error('errir', err);
-        this.bclient.connect();
-      });
-      this.bclient.connect();
-    }
-    */
-    //this.bclient.connect();
+  return Promise.all([this.processing[0]]).then(() => {
```
holm: After upgrade to latest version of bull we are seeing that on shutdown, it doesn't properly wait for all jobs to complete. 

It seems to me that this is likely the change that caused the problem. Was it intentional to change this to only wait for the first job?

gabegorelick: No, that seems like a dumb mistake on my part. Thanks for catching it @holm.

gabegorelick: Fix: https://github.com/OptimalBits/bull/pull/1586

holm: Thanks for the quick fix!

PR #1552, opened 2019-11-08
fix(pause): don't initialize bclient if not waiting for jobs to finish

gabegorelick:
This is the same change as #1347 but for `pause(true, true)`.

Fixes https://github.com/OptimalBits/bull/issues/1553

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -779,6 +779,11 @@ Queue.prototype.pause = function(isLocal, doNotWaitActive) {
           });
         }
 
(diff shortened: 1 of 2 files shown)


gabegorelick:
Thanks @manast! It would be nice to release this and https://github.com/OptimalBits/bull/pull/1542 as 3.12.1, since they're minor fixes to stuff released in 3.12.0.

gabegorelick:
Forgot to mention that 3.12.1 was released with this change. Thanks @manast!

PR #1558, opened 2019-11-13
fix some type in reference doc regard status names

abdolrhman:
move to fail: status from active to failed
move to complete status active to completed


manast:
The documentation is correct already. It could probably be even more clear, but it is correct nonetheless.

PR #1563, opened 2019-11-16
fix(options): merge redis options with parsed url

klouvas:
With this improvement we can pass Redis TLS options and use url parameter at the same time.
Example:
```javascript
const tls = { key: "", cert:"", ca:"" };
const redis_url = "redis://localhost:6379";
const queue = new Queue(name, redis_url, { redis: { tls } });
```


manast:
Closing since this is already fixed.

PR #1595, opened 2019-12-21
Update REFERENCE.md

Akintunde102:
Simple spelling Correction


manast:
thanks for the fix, however we need the commits to be in conventional commits format:
https://www.conventionalcommits.org/en/v1.0.0/
So for example:
docs: correct grammar

PR #1605, opened 2020-01-03
fix: Remove the lazy client error handler on close

lavarsicious:
The lazy client error event handler is never unregistered when the queue is closed. This becomes more problematic in a scenario in which `options.createClient` is specified.


gabegorelick:
@lavarsicious Are you able to add a test for this?

PR #1607, opened 2020-01-09
Handling Number type check for Job Progress

ganeshcse2991:
Adding Number type check to the Job progress field.


smaznet:
You should do a commit to change documents

Many of users like me need to set object for progresses

ganeshcse2991:
Yeah I kind of misunderstood it. Sorry for the PR

Issue #1231, opened 2019-03-01
Calling Promote on Job does not check if Queue is Ready


chocof:
## Description
When we create delayed jobs in a queue from a producer application and try to promote them from a consumer application then we get an error.

## Scenario

The first script produces some delayed jobs
```javascript 
const Queue = require('bull');
const testQueue = new Queue('testqueue', 'redis://127.0.0.1:6379');

testQueue.process((job, done) => {
  console.log('Executing Job', job.data);
  done();
});

const init = async () => {
  await testQueue.add({ test: 'test1' }, {
    delay: 1000000,
    attempts: 2,
    timeout: 10,
    removeOnComplete: true,
    removeOnFail: true
  });
};

init();
```
This script reads and promotes them to be executed
```javascript
const Queue = require('bull');
const testQueue = new Queue('testqueue', 'redis://127.0.0.1:6379');

testQueue.process((job, done) => {
  console.log('Executing Job', job.data);
  done()
});

const init = async () => {
  // assume that there is another place where we initialise
  // some delayed jobs and there is at least one of them in
  // the results of getDelayed
  const jobs = await testQueue.getDelayed();
  for (let i = 0; i < jobs.length; i++) await jobs[i].promote();
}
init();
```

## Result
```bash
(node:13481) UnhandledPromiseRejectionWarning: TypeError: queue.client.promote is not a function
    at Object.promote (/Users/ftsokos/mainProj/bull/lib/scripts.js:342:25)
    at Job.promote (/Users/ftsokos/mainProj/bull/lib/job.js:281:20)
    at init (/Users/ftsokos/mainProj/code_samples/bull.sample.js:11:17)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:13481) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:13481) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## Cause 
After looking into the code I believe that the problem here is that the queue has not yet
been properly initialised when we are trying to access the promote function.

To fix this issue I added the following code to job.promote function lib/job.js:280
```javascript
Job.prototype.promote = function() {
  const queue = this.queue;
  const jobId = this.id;
[...]

PR #1619, opened 2020-01-19
fix: call Queue.isReady() before job promote is called

chocof:
The issue is described here: https://github.com/OptimalBits/bull/issues/1231

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -304,12 +304,13 @@ Job.prototype.moveToDelayed = function(timestamp, ignoreLock) {
 Job.prototype.promote = function() {
   const queue = this.queue;
   const jobId = this.id;
-
-  return scripts.promote(queue, jobId).then(result => {
-    if (result === -1) {
-      throw new Error('Job ' + jobId + ' is not in a delayed state');
-    }
-  });
+  return queue.isReady().then(() =>
+    scripts.promote(queue, jobId).then(result => {
+      if (result === -1) {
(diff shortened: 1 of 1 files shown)


gohiei:
I test it at my local machine, and it works well.

lib/job.js
```diff
@@ -304,12 +304,13 @@ Job.prototype.moveToDelayed = function(timestamp, ignoreLock) {
...
   const queue = this.queue;
   const jobId = this.id;
-
-  return scripts.promote(queue, jobId).then(result => {
-    if (result === -1) {
-      throw new Error('Job ' + jobId + ' is not in a delayed state');
-    }
-  });
+  return queue.isReady().then(() =>
+    scripts.promote(queue, jobId).then(result => {
```
manast: a return is needed here so that the promise returned by promote is propagated correctly.

chocof: Hello @manast ,  do you mean the following part of the code
`... .then(() => scripts.promote() ...)`
if yes then in es6 the syntax `() => value` is identical to `() => { return value }` so it is practically like writing 
`return queue.isReady().then(() => { return scripts.promote(queue, jobId) }).then() ... `
If this is not what you meant please correct me

chocof: however if you would like to use the previous syntax `() => {return ...}` (which as i can see is used in this file) then i can edit my pr

manast: no sorry, the code is correct, I did not notice the arrow function.

PR #1631, opened 2020-01-31
feat: add "preventParsingData" option of job to prevent data parsing

### LINKED ISSUE #1630: Feature request: add option to prevent parsing of data
### Issue type: Enhancements, Feature request

The `JSON.parse()` method accepts an optional the second param - **reviver** function, that can be used to perform a transformation on the resulting object before it is returned ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)).

Will be nice to have possibility to use my own **reviver** function during parsing data of job. For example, it'll be useful when we use Date in data of job and would like to parse it exactly to JS Date, not to string.

pakhuta:
With this improvement we can specify a job option `{preventParsingData: true}` to preventing parsing of job data.
Close #1630

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -556,8 +556,10 @@ Job.prototype._saveAttempt = function(multi, err) {
 };
 
 Job.fromJSON = function(queue, json, jobId) {
-  const data = JSON.parse(json.data || '{}');
   const opts = JSON.parse(json.opts || '{}');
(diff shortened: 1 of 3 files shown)


pakhuta:
Hi @manast,
Could you tell when this PR will be released?

PR #1646, opened 2020-02-11
fix: add timeout fallback

wenq1:
Small enhancement to allow for custom message passed to TimeoutError

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -197,7 +197,7 @@ const emailQueue = new Queue('email');
 emailQueue.process('sendEmail', 25, sendEmail);
 ```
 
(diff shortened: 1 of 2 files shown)


REFERENCE.md
```diff
@@ -260,6 +260,7 @@ interface JobOpts {
 
   lifo: boolean; // if true, adds the job to the right of the queue instead of the left (default false)
   timeout: number; // The number of milliseconds after which the job should be fail with a timeout error [optional]
+  timeoutFallback: string | Error; // Fallback passed to p-timeout when timeout is specified [optional]
```
manast: I think it can be a string or a Function according to p-timeout documentation, right?

Issue #1304, opened 2019-05-02
Possible infinite loop if BRPOPLPUSH fails ?


smennesson:
## Description

Hello today we faced a problem that seemed to be an infinite loop on Bull library.
Our service is hosted on Heroku with the Redis addon and today we reached the memory quota of the Redis DB.
What happened is that we had an enormous log stack saying this:
```
BRPOPLPUSH { ReplyError: OOM command not allowed when used memory > 'maxmemory'. 
    at parseError (/app/node_modules/ioredis/node_modules/redis-parser/lib/parser.js:179:12) 
    at parseType (/app/node_modules/ioredis/node_modules/redis-parser/lib/parser.js:302:14) 
  command: 
   { name: 'brpoplpush', 
     args: 
      [ 'bull:<name-of-our-job>:wait', 
        'bull:<name-of-our-job>:active', 
        '5' ] } } 
```

The log stack took up to 1gb in a few minutes until we fix the quota issue.

By looking a little bit to the code in `lib/queue.js`, it seems that the error on BRPOPLPUSH is ignored in `Queue.prototype.getNextJob`. So I guess that what happened is that the loop searching for new jobs to process was infinitely popping the error.

I don't have enough knowledge about how Bull internally works to propose a fix, but I think this is something that should be handled, maybe by detecting when there is several errors on BRPOPLPUSH and add a waiting duration when this happens to frequently.

## Bull version
v3.7.0 

(just seen that 3.8 is out ; it doesn't seem that it would be fixed in this version by reading the changelog)

balexandre:
exactly same issue on the same circumstances: Heroku, Redis, BRPOPLPUSH :(

[![enter image description here][1]][1]

  [1]: https://i.stack.imgur.com/Cn3xS.png

smennesson:
@balexandre actually reaching the Redis limit is not the real problem here. There could be a multiple reason causing your Redis db to be full and there is a few tools to show you what keys are taking the most memory space. So you can either clean a little bit or upgrade your plan.

Problem is that the Bull library is in an infinite loop when this occurs, generating a huge volume of logs, and ignoring the error, so we have no way to be warned this is occurring.

kand617:
I have the same issue..... 

Is there some way to catch this error? 
And log a different message? 

I am not even sure where it's thrown from

mmelvin0:
[...]

PR #1674, opened 2020-03-11
fix: only swallow connection errors #1304

tnolet:
Please reference the issue #1304

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -1091,8 +1091,8 @@ Queue.prototype.getNextJob = function() {
           }
         },
         err => {
-          // Swallow error
-          if (err.message !== 'Connection is closed.') {
+          // Swallow error only for connection closed. See GH issue #1304
+          if (err.message === 'Connection is closed.') {
             console.error('BRPOPLPUSH', err);
           }
         }


tnolet:
This seems to only fail on Node 12

manast:
ok, so we are still swallowing the error (i,.e. not throwing it up), it is just that we not write the error on the console.

tnolet:
@manast I don't really know what you mean then. 

> ok, so we are still swallowing the error (i,.e. not throwing it up), it is just that we not write the error on the console.

So we do want to catch the "connection closed" error, but just rethrow it? i.e.
```
if (err.message === 'Connection is closed.') {
  throw(err)
}
```

Or do we want to ignore the "connection closed" error and rethrow all others?

```
if (err.message !== 'Connection is closed.') {
  throw(err)
}
```

in both cases, we log nothing.

tnolet:
@manast would love to move ahead with this. We are still at risk blowing up our logging infrastructure if we have an hick up

tnolet:
@manast It has been one month since I last pinged this PR. Can you let me know if there is any chance we this will be picked up by on of the maintainers?

manast:
I am not sure what to say, this PR does nothing particularly useful, just changes when to console.log the error, it is not addressing #1304 as far as I can see.

tnolet:
@manast Ok, clear. But it seems we are going in circles here a bit. What do we need to change to address #1304 effectively. As mentioned in the original issue:

###
@manast what is your suggestion for moving this ticket forward? I have the feeling we are going in circles a bit. I would be happy to remove/close my PR in favour of a better solution.

The reason I'm pushing this is that it is still a risk for many folks using logging solutions with hard limits.

As mentioned, I'm ready to help or have my team also look at it.
###

manast:
I think that we need to add a delay, exponential best but fixed will also help, where in case of this "unexpected errors" we wait a couple of seconds. The connection is closed should not be handled as unexpected though since this is the error you get when closing the queue and you do not want to delay the closing more than necessary.

lib/queue.js
```diff
@@ -1091,8 +1091,8 @@ Queue.prototype.getNextJob = function() {
           }
         },
         err => {
-          // Swallow error
-          if (err.message !== 'Connection is closed.') {
+          // Swallow error only for connection closed. See GH issue #1304
+          if (err.message === 'Connection is closed.') {
             console.error('BRPOPLPUSH', err);
           }
```
dobesv, line 1097:
```suggestion
          // Ignore 'Connection is closed.' error if closing the queue
          if (!(err.message === 'Connection is closed.' && this.closing)) {
            console.error('BRPOPLPUSH', err);
            throw err;
          }
```

tnolet: @dobesv this is interesting. How about the delay mentioned by @manast? Do we need that, or not? It seems to be part of your suggestion here https://github.com/OptimalBits/bull/issues/1746

dobesv: Yeah, after I proposed this I also thought of a way to sort of throttle attempts if we got an error.  It's hard to say exactly what the best policy is here, we're a bit in the dark.  But for a resilient system we could try to assume that redis is temporarily down or a graceful shutdown was initiated but we haven't been told yet (e.g. closing was not set *yet* but might be very soon).

It seems like maybe some kind of `emit('error')` or something like that to allow the library user to detect this condition might also be useful.

tnolet:
This was fixed by https://github.com/OptimalBits/bull/pull/1974
Thanks to @manast for clearing this up!

PR #1692, opened 2020-04-06
Error initializing Lua scripts fix

Abdullahzafar400:
changing **__dirname** to **path.dirname(require.resolve("bull")) + '/lib/commands'**  because webpack has different behaviour and its generate error **Error: Error initializing Lua scripts\n at**

### DIFF
--- a/lib/commands/index.js
+++ b/lib/commands/index.js
@@ -29,7 +29,7 @@ module.exports = (function() {
 
   return function(client) {
     return utils.isRedisReady(client).then(() => {
-      scripts = scripts || loadScripts(__dirname);
+      scripts = scripts || loadScripts(path.dirname(require.resolve("bull")) + '/lib/commands');
 
       return scripts.then(_scripts => {
         return _scripts.forEach(command => {


mikezucc:
I believe this is incorrect as well. If you are using webpack to build your bundle before packaging in a docker image, this will reference an absolute path only valid on the building machine.

mikezucc:
Ok for anyone running into this into the future with a webpack build to docker build flow: define bull as an external module and npm install in your Dockerfile such that bull isn't built into the module and instead is able to reference valid paths.

Here is a simple webpack.config.js we came up with to solve for this problem (note you develop with webpack build as well instead of plain node)
```javascript
const pkg = require('./package.json');

module.exports = {
  mode: process.env.NODE_ENV,
  entry: ['@babel/polyfill', './src/index.js'],
  target: 'node',
  externals: {
    'bull': 'commonjs2 bull',
  },
  module: {
    . . .
  }
};```

i-ce-fish:
My webpack config , copy lua script to output root path, it solve my problem

`const CopyWebpackPlugin = require('copy-webpack-plugin');  `

`plugins: [`
 `   new CopyWebpackPlugin({`
  `    patterns: [{`
   `       from: "node_modules/bull/lib/commands",`
    `      to: ""`
     `  }]`
`})]`

PR #1695, opened 2020-04-09
Fix getJobs type description

Mizumaki:
I fixed the DefinitelyTypes also.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43772

I think the `types` can be only `JobStatus`.
Am I right?


Mizumaki:
Seems [the PR in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43772) is successfully merged 🎉

Issue #1664, opened 2020-03-02
UnhandledPromiseRejection with on connection error. 


jan-osch:
I am getting an `UnhandledPromiseRejectionWarning`. What 

```javascript
const Queue = require('bull');

const q = new Queue('test', 'redis://localhost:8009'); // no redis instance here

q.on('error', e => console.error('Error event', e.message));

q.process('task', 1, (job) => {
  console.log('ok');
  return Promise.resolve(job.id);
})
 .catch(error => console.error('Failed to start consumer', error.message));
```

I am getting this output:
```
Error event connect ECONNREFUSED 127.0.0.1:8009
Error event Error initializing Lua scripts
Failed to start processing connect ECONNREFUSED 127.0.0.1:8009
(node:12173) UnhandledPromiseRejectionWarning: Error: connect ECONNREFUSED 127.0.0.1:8009
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1137:16)
(node:12173) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:12173) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Error event connect ECONNREFUSED 127.0.0.1:8009
Error event connect ECONNREFUSED 127.0.0.1:8009
```

How can I handle the `UnhandledPromiseRejectionWarning` error ?

cfey:
Hi @jan-osch,
I got the same error
I fixed it by removing
`global.Promise = require('bluebird');`
This were declared somewhere in my project, without this global Promise override everything works as expected.

Let me know if this fixed your issue :)

Cheers

jan-osch:
@cfey 
Hey  - thanks for checking that out. I don't override promises anywhere in the project. The sample code I provided is a single isolated file that I use to reproduce the issue.

I can confirm that the issue still exists.
Node version: v12.16.0
Bull 3.13.0

LamSkills:
I still get this error but I find a (maybe ?) cleaner solution to fix that annoying warning by using Promise.

### Solution

```js
queue.isReady()
.then(() => {
[...]

PR #1697, opened 2020-04-12
fix: update yarn packages to latest minor

bvallelunga:
Ran `yarn upgrade` to update all packages to the latest minor version. This is done so we can include a fix by https://github.com/luin/ioredis/pull/1095 for Bluebird customers. I believe there have been a couple issues like #1664 it could fix.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1725,9 +1725,9 @@ inquirer@^6.2.2:
     through "^2.3.6"
 
 ioredis@^4.14.1:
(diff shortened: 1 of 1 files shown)


bvallelunga:
@manast would love your review if you have the time. Appreciate the look.

manast:
thanks for the PR. In order for this to have any effect it is needed to update the package.json itself, the lock file is not used by dependents only for development of bull itself.

Issue #1134, opened 2018-11-18
'waiting' event is not emitted for priority job


btd:
Consider such basic code:
```js
"use strict";

const Queue = require("bull");

const q = new Queue("test");

q.on("error", function(error) {
  console.log("error");
})

  .on("waiting", function(jobId) {
    console.log("waiting");
  })

  .on("active", function(job, jobPromise) {
    console.log("active");
  })

  .on("stalled", function(job) {
    console.log("stalled");
  })

  .on("progress", function(job, progress) {
    console.log("progress");
  })

  .on("completed", function(job, result) {
    console.log("completed");
  })

  .on("failed", function(job, err) {
    console.log("failed");
  })

  .on("paused", function() {
    console.log("paused");
  })

  .on("resumed", function(job) {
    console.log("resumed");
  })

  .on("cleaned", function(jobs, type) {
    console.log("cleaned");
  })

  .on("drained", function() {
    console.log("drained");
  })

  .on("removed", function(job) {
    console.log("removed");
  });

q.process("*", 1, () => {
  console.log("EXECUTING");
});

q.add({});
q.add({}, { priority: 100 });
```

As output you will see:
```
waiting
EXECUTING
active
completed
EXECUTING
active
completed
drained
```
## Bull version
```
$ npm ls bull
ingo-app-tasks@4.5.0 /dir
└── bull@3.4.8
```

btd:
I think missing second call for publish in addJob lua script like in this line https://github.com/OptimalBits/bull/blob/develop/lib/commands/addJob-6.lua#L84

abdatta:
Any updates on this bug? I'm facing issue with this :(

PR #1710, opened 2020-04-26
fix: emit waiting event when adding a priority job (#1134)

seriallos:
Ensure a waiting event is sent for priority jobs and adds a test to verify this behavior.


manast:
good catch, thanks!

PR #1720, opened 2020-05-05
feat: pass clientName to createClient function

tebriel:
Overrides the options.redis settings for clientName for creating a new
client

When specifying `createClient` as part of the `Bull.QueueOptions` there is no way to know the `Queue`'s name (from `queue.clientName`).

When calling `queue.process`, [`this.setHandler`](https://github.com/OptimalBits/bull/blob/develop/lib/queue.js#L621) is invoked which calls [`setWorkerName`](https://github.com/OptimalBits/bull/blob/develop/lib/worker.js#L12-L21) however this only does so on the `client`, and not on the `eclient` (`subscriber`) nor on the `bclient`.

I opted to hand the `clientName` via `connectionName` in the redis options (overriding the defaults passed in to the `QueueOptions` to allow the `createClient` function to determine what to do with these options.

Additionally I slightly tweaked the example to note that this value is present and to better showcase how to use the `createClient` function.


test/test_queue.js
```diff
@@ -1915,7 +1920,10 @@ describe('Queue', () => {
       queue.add({});
       queue.add({});
 
-      queue.on('completed', _.after(2, () => done()));
+      queue.on(
```
tebriel, line 1923: running `npm run prettier` formatted this line, I can back out the change if you like.

PR #1748, opened 2020-05-23
Add a note about redis disconnect not being called when you use creat…

dobesv:
…eClient


manast:
Thanks for the contribution. Since we have improved the docs since this PR. I will close it since it seems superfluous.

PR #1753, opened 2020-05-28
Grouped rate limiting

aramix:
This commit adds a `key` property to the `limiter` and allows you to define a key from the `data` object passed to `Queue#add` method so that when doing rate limiting it will group and limit the entries by that key.
`key` property should be a string and can also define a nested key

### DIFF
--- a/lib/commands/moveToActive-8.lua
+++ b/lib/commands/moveToActive-8.lua
@@ -55,6 +55,9 @@ if jobId then
     -- local jobCounter = tonumber(rcall("GET", rateLimiterKey))
     if(ARGV[9]) then
       local handle = string.match(jobId, "[^:]+$");
+      if(handle == nil) then
+        return
+      end
       rateLimiterKey = KEYS[6] .. ":" .. handle;
     end
     local jobCounter = tonumber(rcall("INCR", rateLimiterKey));
--- a/lib/commands/moveToActive-8.lua
+++ b/lib/commands/moveToActive-8.lua
@@ -29,6 +29,7 @@
       ARGV[6] optional jobs per time unit (rate limiter)
       ARGV[7] optional time unit (rate limiter)
       ARGV[8] optional do not do anything with job if rate limit hit
+      ARGV[9] optional specifies the key to limit by
 ]]
 
 local jobId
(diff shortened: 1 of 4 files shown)


lib/queue.js
```diff
@@ -705,6 +707,9 @@ Queue.prototype.add = function(name, data, opts) {
   opts = _.cloneDeep(opts || {});
   _.defaults(opts, this.defaultJobOptions);
 
+  if (this.limiter && this.limiter.key) {
+    opts.jobId = Date.now() + ":" + _.get(data, this.limiter.key);
```
manast, line 711: Instead of Date.now() a uuid should be used, otherwise we risk to loose jobs that are added within the same millisecond

lib/commands/moveToActive-8.lua
```diff
@@ -50,17 +51,22 @@ if jobId then
 
   if(maxJobs) then
     local rateLimiterKey = KEYS[6];
+
     -- local jobCounter = tonumber(rcall("GET", rateLimiterKey))
-    local jobCounter = tonumber(rcall("INCR", rateLimiterKey))
+    if(ARGV[9]) then
+      local handle = string.match(jobId, "[^:]+$");
```
manast, line 57: handle could be nil if jobId for some reason does not comply with the naming convention. I propose we check for nil and in that case no rateLimiting is performed at all.

lib/queue.js
manast: also, the user may have specified already a custom jobId, in that case we should base the new jobId on the one specified by the user instead of uuid.

lib/commands/moveToActive-8.lua
aramix: @manast thanks for the code review and useful comments. I have updated the code to cover those cases.

manast-apsis:
I am thinking, what would be the pros/cons of instead of using the jobId to pass the rate limit key, we add another option to the job options, i.e. ```rateLimitKey``` and use that key directly as ARGV[9] in the script. So you will do something like:
```ts
queue.addJob(jobData, { rateLimitKey: 'foobar' });
```

?

suchakraborty:
This would be a great feature for my use case.

manast:
I reworked this PR into this one_ https://github.com/OptimalBits/bull/pull/1804
Thanks for your initial code and ideas!

PR #1770, opened 2020-06-16
docs(REFERENCE.md): fix interface code block

n1ru4l:


manast:
closing this because even though the original code block is wrong, the proposed change leaves the code with unbalanced curly braces

PR #1780, opened 2020-06-29
fix: error when .lua scripts missing in builded bundle

zhelnov:


manast:
Thanks for the PR. I would love to merge this PR if you could just fix the conflict :).

zhelnov:
@manast just fixed.

PR #1788, opened 2020-07-03
fix: forbid usage of addBulk for repeatable jobs

### LINKED ISSUE #1731: addBulk does not work with repeatable jobs
## Description

`addBulk` function does not work properly for `repeatable` jobs, the `nextRepeatableJob` is not invoked for repeatable jobs.

#### (An easy to reproduce test case will dramatically decrease the resolution time.)

The following code :

```ts
imgQueue.addBulk([{ data: { img }, opts: {repeatable: { every: 5_000 }}}])
```

is different than : 

```ts
imgQueue.add({ img }, {repeatable: { every: 5_000 }})
```

In the `add` code we have this behaviour once the job is created : 

```ts
if (opts.repeat) {
  return this.isReady().then(() => {
    return this.nextRepeatableJob(name, data, opts, true);
  });
} 
```

It is missing in the `addBulk` function.

## Solutions

I'm OK to fix it. Here are two possible solutions : 

* forbid addBulk for repeatable jobs **it could sounds weird to bulk add repeatable jobs, but it is kind of breaking**
* call `nextRepeatableJob` for each repeatable jobs of the batch once created. **Bad for perfs but still better than multiple adds**

which one do you prefer ?

## Bull version
3.14

## Additional information

Thx for your lib :)

jtassin:
Closes #1731

Forbid in code to use `addBulk` on repeatable jobs.

See : https://github.com/OptimalBits/bull/issues/1731 for details

Once merged, I will also patch the types in [@types/bull](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/bull)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - feat: add isPaused fixes #1274
 - fix: emit waiting event when adding a priority job (#1134)
+- fix: forbid usage of addBulk for repeatable jobs (#1731)
```
manast, line 7: 3.15 has already been released so this change should go on the next release

manast:
thanks for the patch. Unfortunately I made some changes recently on addBulk which are conflicting with your PR.

jtassin:
> thanks for the patch. Unfortunately I made some changes recently on addBulk which are conflicting with your PR.

Damn OK, I will rebase it

alexfu:
Any update on this? I ended up spending a large amount of time debugging my app, thinking I was doing something wrong since the parameters for both `add` and `addBulk` are the exact same. I'd hate the see new users of `bull` fall into the same trap.

I also think updating the docs to explicitly state that `addBulk` does not work with `repeat` and it will throw an exception is worth including in this PR.

manast:
Currently the tests are failing, as soon as they pass I can merge it.

Issue #1747, opened 2020-05-23
Queue close does not close redis connection when using custom createClient function


dobesv:
## Description

When creating queues using the `createClient` function to connect to redis, the queues are not closed when the queue is closed.

## Minimal, Working Test code to reproduce the issue.

```
const Queue = require('bull');
const Redis = require('ioredis');
const Bluebird = require('bluebird');

const main = async () => {
  const clients = [];
  const createClient = (type, redisOpts) => {
    const redisOptions = {
      ...redisOpts,
      connectionName: ['bull', type].join('-'),
    };
    const client = new Redis(redisOptions);
    clients.push(client);
    console.log('start redis', redisOptions.connectionName);
    client.on('connect', () =>
      console.log('redis connected', redisOptions.connectionName),
    );
    client.on('end', () =>
      console.log('end redis', redisOptions.connectionName),
    );
    return client;
  };
  const q1 = new Queue('1', { createClient });
  const q2 = new Queue('2', { createClient });
  const q3 = new Queue('3', { createClient });
  q1.process(1, job => console.log('q1', job.data)).catch(err =>
    console.error(err),
  );
  q2.process(2, job => console.log('q2', job.data)).catch(err =>
    console.error(err),
  );
  q3.process(3, job => console.log('q3', job.data)).catch(err =>
    console.error(err),
  );
  console.log('processors ready');
  await Bluebird.delay(1000);
  await Promise.all([q1.add({ q: 1 }), q2.add({ q: 2 }), q3.add({ q: 3 })]);
  await Bluebird.delay(1000);
  await q1.close();
  console.log('closed q1');
  await q2.close();
  console.log('closed q2');
  await q3.close();
  console.log('closed q3');
  Object.values(clients).forEach(client => {
    console.log(client.options.connectionName, client.status);
  });
  console.log('end of main');
};

main();

```

Output:

```
$ NODE_OPTIONS=--max-old-space-size=100 node ./reproduce2.js 
start redis bull-client
start redis bull-client
start redis bull-client
processors ready
redis connected bull-client
redis connected bull-client
redis connected bull-client
start redis bull-subscriber
start redis bull-subscriber
start redis bull-subscriber
redis connected bull-subscriber
redis connected bull-subscriber
redis connected bull-subscriber
start redis bull-bclient
start redis bull-bclient
redis connected bull-bclient
start redis bull-bclient
redis connected bull-bclient
redis connected bull-bclient
q1 { q: 1 }
[...]

PR #1790, opened 2020-07-03
Add some notes to Reusing Redis Connections

dobesv:
Help avoid pitfalls with closing reused connections and keyPrefix.

Fix for https://github.com/OptimalBits/bull/issues/1747

### DIFF
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -70,7 +70,8 @@ A standard queue requires **3 connections** to the Redis server. In some situati
 
 Notes:
 - bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880), so you should return a new connection each time this is called.
-- client and subscriber connections can be shared and will not be closed when the queue is closed.  When you are shutting down the process, first close the queues, then the shared connections (if they are shared).  If you are not sharing client and subscriber connections but using `createClient` to do some custom connection logic, you will need to keep a list of all the connections you created so you can manually close them later when the queue shuts down
+- client and subscriber connections can be shared and will not be closed when the queue is closed.  When you are shutting down the process, first close the queues, then the shared connections (if they are shared).
+- if you are not sharing connections but still using `createClient` to do some custom connection logic, you may still need to keep a list of all the connections you created so you can manually close them later when the queue shuts down, if you need a graceful shutdown for your process
 - do not set a `keyPrefix` on the connection you create, use bull's built-in prefix feature if you need a key prefix
 
 ```js
--- a/PATTERNS.md
+++ b/PATTERNS.md
@@ -69,7 +69,7 @@ Reusing Redis Connections
 A standard queue requires **3 connections** to the Redis server. In some situations you might want to re-use connections—for example on Heroku where the connection count is restricted. You can do this with the `createClient` option in the `Queue` constructor.
 
 Notes:
-- bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880), so you should return a new connect each time this is called. These connections will be disconnected/reconnected by bull automatically as part of queue operations and will be disconnected when the queue is closed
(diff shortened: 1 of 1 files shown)


manast:
thanks for the contribution, appreciated.

PATTERNS.md
```diff
@@ -66,7 +66,12 @@ The most robust and scalable way to accomplish this is by combining the standard
 Reusing Redis Connections
 -------------------------
 
-A standard queue requires **3 connections** to the Redis server. In some situations you might want to re-use connections—for example on Heroku where the connection count is restricted. You can do this with the `createClient` option in the `Queue` constructor (note: bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880)):
+A standard queue requires **3 connections** to the Redis server. In some situations you might want to re-use connections—for example on Heroku where the connection count is restricted. You can do this with the `createClient` option in the `Queue` constructor.
+
+Notes:
+- bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880), so you should return a new connect each time this is called. These connections will be disconnected/reconnected by bull automatically as part of queue operations and will be disconnected when the queue is closed
```
manast, line 72: A note here, the bclient is not disconnected when the queue closes. See it like this, all connections created outside of Bull are never closed or destroyed by Bull and need to be disconnected manually after closing Bull.

PATTERNS.md
```diff
@@ -66,7 +66,12 @@ The most robust and scalable way to accomplish this is by combining the standard
...
+- client and subscriber connections can be shared and will not be closed when the queue is closed.  When you are shutting down the process, first close the queues, then the shared connections (if they are shared).  If you are not sharing client and subscriber connections but using `createClient` to do some custom connection logic, you will need to keep a list of all the connections you created so you can manually close them later when the queue shuts down
+- do not set a `keyPrefix` on the connection you create, use bull's built-in prefix feature if you need a key prefix
```
manast, line 74: the built-in prefix is the queue name indeed, maybe some users will get confused if we mention a prefix feature :/

PATTERNS.md
dobesv, line 72:
```suggestion
- bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880), so you should return a new connection each time this is called.
```

PATTERNS.md
```diff
@@ -66,7 +66,12 @@ The most robust and scalable way to accomplish this is by combining the standard
...
+A standard queue requires **3 connections** to the Redis server. In some situations you might want to re-use connections—for example on Heroku where the connection count is restricted. You can do this with the `createClient` option in the `Queue` constructor.
+
+Notes:
+- bclient connections [cannot be re-used](https://github.com/OptimalBits/bull/issues/880), so you should return a new connection each time this is called.
+- client and subscriber connections can be shared and will not be closed when the queue is closed.  When you are shutting down the process, first close the queues, then the shared connections (if they are shared).  If you are not sharing client and subscriber connections but using `createClient` to do some custom connection logic, you will need to keep a list of all the connections you created so you can manually close them later when the queue shuts down
```
dobesv, line 73:
```suggestion
- client and subscriber connections can be shared and will not be closed when the queue is closed.  When you are shutting down the process, first close the queues, then the shared connections (if they are shared).
- if you are not sharing connections but still using `createClient` to do some custom connection logic, you may still need to keep a list of all the connections you created so you can manually close them later when the queue shuts down, if you need a graceful shutdown for your process
```

PATTERNS.md
dobesv: I ran into HUGE issues using a `keyPrefix` in my redis connections, so I was hoping by specifically mentioning NOT to set a keyPrefix here it might help people avoid that.

PATTERNS.md
dobesv: Ah right, I was confused because it does kind of ... disconnect and reconnect all the time.  But after the queue is closed, the connection is left in a "connected" state, so it still has to be closed in a separate step.

PR #1794, opened 2020-07-08
docs: added @breejs

niftylettuce:
I created @breejs as a simpler alternative to Bull.  Just in case someone doesn't need something as advanced as Bull, it might be helpful to share.

Bull can even be used inside of Bree itself, in one of the workers.  For example, you could have a Bree worker that spawns a Bull queue!

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Since there are a few job queue solutions, here is a table comparing them:
 | UI              | ✓             |   ✓   |     |   ✓    |
 | Optimized for   | Jobs / Messages | Jobs | Messages | Jobs |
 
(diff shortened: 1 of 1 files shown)


manansharma1991:
The creator of bree is just sharing false information on his own repo. I created a issue for the same, but he decided to delete it (not close). By deleting maybe he doesn't want others to see what I was talking about.

All his claims about Bull not being production ready are false and I will share the screenshot of that issue to explain why.

CleanShot 2021-11-13 at 09 22 11@2x

manast:
@manansharma1991 thanks for the information, but it is not worth to bring attention to this. Let's keep focus in improving Bull rather than waste time on FUD spreaders.

PR #1796, opened 2020-07-09
Throw an error if concurrency value is invalid

theoephraim:
Had a configuration error on my side and was passing through `"123,"` as a concurrency setting for a queue. This caused `concurrency--` to be `NaN` which stopped the queue from processing any jobs. So I've added a quick check to throw an error.

Also in the process while trying to get things running, I was having problems because spread syntax is not supported under node v8.3 - which was making the linter complain.

I also tried to add a test, but there wasn't really an obvious spot to do it. Generally more docs around getting the code up and running and contributing would be helpful.

Cheers and thanks for writing and maintaining this package!


lib/queue.js
```diff
@@ -858,6 +858,10 @@ Queue.prototype.isPaused = function(isLocal) {
 };
 
 Queue.prototype.run = function(concurrency) {
+  // Check if `concurrency--` will be NaN, which is falsy, so jobs never start processing
+  if (_.isNaN(concurrency - 1)) {
```
manast, line 862: I think it is better to use the more straightforward isFinite: https://lodash.com/docs/4.17.15#isFinite

PR #1802, opened 2020-07-15
feat(sandbox): kill child workers gracefully

GabrielCastro:
Properly handle SIGTERM in the child workers by waiting for the current
job to exit, then exiting cleanly.

Queue.close now also kills the workers and waits for them to close


GabrielCastro:
@manast This is the same change as https://github.com/taskforcesh/bullmq/pull/243 would it be possible to back-port it here too

PR #1804, opened 2020-07-16
feat(rate-limiter): add grouping support

manast:


iflairgit:
If we are providing groupKey to the queue options how can we group the bunch of jobs instead just like https://docs.bullmq.io/bullmq-pro/groups.

manast:
We have moved away from this implementation of groups as it does have several issues that can affect performance quite a lot. Instead we recommend the much more capable Pro version https://bullmq.io/#bullmq-pro
If you still want to use groupKeys in Bull, it is as simple as specifying a certain groupKey and that will be used for an independent rate-limit for just that key.

PR #1818, opened 2020-08-13
feat: make pause forward compatible with bullmq

manast:


lib/queue.js
```diff
@@ -90,20 +90,15 @@ const Queue = function Queue(name, url, opts) {
...
+    opts = {
+      ...opts,
+      redis: { ...redisOptsFromUrl(url), ...(opts || {}).redis }
+    };
   } else {
-    opts = url;
+    opts = url || {};
   }
 
-  opts = _.cloneDeep(opts || {});
```
marshalys, line 104: delete this code may cause some problem，becase `delete opts.redis.keyPrefix;` after this code.

marshalys: code example:
```
var Queue = require('bull');
var option = { redis: { host: '127.0.0.1', port: '6379', password: '', db: 1, keyPrefix: 'test' }};
var videoQueue = new Queue('video transcoding', {...option}); // prefix: test
var imageQueue = new Queue('image transcoding', {...option}); // prefix: bull
```

manast: yes, you are right, I overlooked it.

PR #1819, opened 2020-08-13
Add Monitoro to list of UIs

AbhilashJN:


AbhilashJN:
@manast  Are there any specific requirements for a UI to be featured here?

PR #1835, opened 2020-08-27
Update return value of getRepeatableJobs in docs

cincodenada:
It does not, as indicated, return an array of Job instances, but of configuration data (as noted in the text).

I'm not sure what the best way to notate this is (perhaps just `Promise<Object[]>`?), but `Job[]` is not correct.


manast:
thanks for the correction!.

PR #1836, opened 2020-08-27
Fix wording around "require" in named jobs

cincodenada:
I'm not actually 100% confident that this is the intended meaning, but the original sentence doesn't parse correctly, and this is my best guess. Please make sure the revised statement is factually correct as well as syntactically :)


manast:
thanks!

PR #1846, opened 2020-09-03
fix(sandbox): clear dangling timeout

epoberezhny:
Cleared dangling timeout which prevents node process from exiting


manast:
very nice, thanks!.

PR #1868, opened 2020-09-22
feat(job): implement Job#retriedOn

### LINKED ISSUE #1867: Question/feature request: Get time since a job was added to the queue OR manually retried?
Hi!

In our setup we have some metrics and alerting around how long jobs spend in the queue before being processed so we can continuously fine-tune the number of workers. There's one scenario where we're having a bit of trouble coming up with the right number: A job that fails and some hours gets retried manually. In that case `processedOn` will be reset to `null` when the job gets retried, and `timestamp` still reflects the time where the job was first added to the queue.

At that point, is there any way to find out how long time the job has spent in the queue since it was manually retried? If the answer is no, I would be happy to open a PR that adds another timestamp to the job so it's possible. Just wanted to check first :)

papandreou:
Add a `retriedOn` property to jobs so "waiting time in queue" metrics can be based on that when the job has been retried.

Use case described in https://github.com/OptimalBits/bull/issues/1867

Fixes #1867

### DIFF
--- a/lib/commands/reprocessJob-4.lua
+++ b/lib/commands/reprocessJob-4.lua
@@ -9,6 +9,8 @@
 
     ARGV[1] job.id,
     ARGV[2] (job.opts.lifo ? 'R' : 'L') + 'PUSH'
+    ARGV[3] token
+    ARGV[4] timestamp
 
   Output:
     1 means the operation was a success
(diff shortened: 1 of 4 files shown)


papandreou:
Hey @manast, do you have a sec to review this? 🤗

manast:
This is a fine PR and I am sorry it took so long to merge, thanks for the good job!

papandreou:
Thanks a lot 🤗

hckhanh:
Is there any documentation for this feat @papandreou

papandreou:
I went to look in the docs, but they didn't mention the much more important `Job.processedOn` either, only the methods. Would probably be good to document all these public properties, yeah.

PR #1872, opened 2020-09-26
Remove ";" from event reference

ianjennings:
These `;` prevents the code examples from being valid code.


manast:
semicolon is optional in javascript, however we use "prettier" so the standard is to actually use semicolon after every sentence.

manast:
https://prettier.io/

Issue #1875, opened 2020-09-30
Issues with rate limiter


SrividhyaB:
## Description
Rate limits don't get applied properly when jobs are added to the queue at a rate greater than the rate limit of the queue continuously.
I think there is some problem with the delay with which the jobs are pushed to the delayed queue while there are many jobs also getting added to the queue.

If I run a script first to add 200 jobs to the queue with a rate limit of 50 jobs/min and then start a script to process these jobs, rate limiter works fine. But while it is processings jobs in the queue, if I add a lot more jobs, the rate limiter doesn't work as expected.

Problems:

1. Rate limit is not obeyed when there is a consistent inflow of jobs to the queue at a rate higher than the limit. It usually processes jobs lesser than the limit configured.
2. Sometimes the jobs are processed out of order.

Trial 1: Reproduces out of order 

Ran a script to add jobs to a queue.
**Rate limit is 2 requests/ per min.**

Output: 

Min:seconds => id of the job scheduled

```

41:12 => 1
41:12 => 2
41:12 => 3
41:12 => 4
42:50 => 5
42:50 => 6 
42:50 => 7
42:50 => 8
43:50 => 9
43:50 => 10
43:50 => 11
43:50 => 12
43:51 => 13
43:51 => 14
43:51 => 15
43:51 => 16
43:52 => 17
43:52 => 18
43:52 => 19
43:52 => 20
44:34 => 21
44:34 => 22
44:34 => 23
44:34 => 24
46:27 => 25
46:27 => 26
46:27 => 27
46:27 => 28
51:29 => 29
51:29 => 30
51:29 => 31
51:29 => 32
```
Ran another script to process jobs from the queue

Min:seconds => id of the job processed

```
41:12 => 1
41:12 => 2
42:12 => 3
42:45 => 4
43:12 => 5
43:45 => 6
44:18 => 7 // NOTE: job with id 8 doesnt get picked.
45:18 => 9
45:51 => 10 
46:24 => 11
47:30 => 13
48:3 => 14
48:36 => 15
48:54 => 8
49:36 => 16
49:42 => 17
50:36 => 12
50:48 => 19
51:36 => 18
52:6 => 20
52:36 => 21
53:0 => 22
```

Trail 2: Reproduces rate limit not being obeyed

**Rate limit is 5 requests/ per min.**

Min:seconds => id of the job scheduled

```
23:34 => 1
23:34 => 2
23:34 => 3
23:34 => 4
23:34 => 5
[...]

PR #1879, opened 2020-10-03
[Rate limiter] Increment jobCounter only when a job is processed. Do not count when it is rate limited

SrividhyaB:
Fixes https://github.com/OptimalBits/bull/issues/1875


manast:
Great job!. I will replace the commit comment so that it adheres to the conventional commit standard (https://www.conventionalcommits.org/en/v1.0.0/)

SrividhyaB:
> Great job!. I will replace the commit comment so that it adheres to the conventional commit standard (https://www.conventionalcommits.org/en/v1.0.0/)

Thank you!

PR #1884, opened 2020-10-10
Fix opts recomposition order

zebulonj:
Default options, if set, were overriding passed options.


manast:
fixed, already, sorry for the confusion.

PR #1905, opened 2020-11-14
fix(queue): allow inline options to override jobDefaultOptions

### LINKED ISSUE #1904: queue.add options gets overriden by defaultJobOptions
I have a code like the following and I expected the options passed to `queue.add` to have priority over queue's `defaultJobOptions`. However, as for now, if you have an option on `defaultJobOptions` it will always override anything you have in your `add` call. This was very confusing and it took me some time to figure out why it wasn't working.

```js
const queue = new Queue('test', {
  defaultJobOptions: {delay: 10000 } 
});
queue.add('job', {}, { delay: 1000 });
```

otaviosoares:
Allows options passed to `queue.add` to override queue's jobDefaultOptions.

fixes #1904

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -687,7 +687,7 @@ Queue.prototype.add = function(name, data, opts) {
     data = name;
     name = Job.DEFAULT_JOB_NAME;
   }
-  opts = { ...opts, ...this.defaultJobOptions };
+  opts = { ...this.defaultJobOptions, ...opts };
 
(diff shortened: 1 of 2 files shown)


manast:
I missed this PR and I created a new equivalent one... only reason I will not merge this one instead is because the one I wrote used await/async and I want to modernize the codebase everytime I have a chance. Thanks for the PR, and I am sorry that I overlooked it.

PR #1926, opened 2020-12-04
Clarify that the limiter max is the number of jobs, not their 1/1000ths.

ryanprince:
The thousands separator confused me for a little while --- in the United States, a `.` usually indicates a decimal separator. This changes the explanation so that its references to the values are identical to the values in the code example.

This should be the correct interpretation of what's written, since it's the interpretation that the test 'should obey the rate limit' in test_rate_limiter.js verifies.


manast:
thanks for the fix!

Issue #758, opened 2017-10-25
Error: Sending custom commands in pipeline is not supported in Cluster mode when use Cluster Mode


zetflo:
When i use bull in redis cluster mode.

Configuration options
```js
const opts = {
    createClient: function(type){
       return new redis.Cluster([
            ...
        ]);
    },
    prefix: '{}'
};
```

Create job it's worked but when i process job with queue.process, 
I encounter the following error

```
Error: Sending custom commands in pipeline is not supported in Cluster mode.
    at Pipeline.exec (\node_modules\ioredis\lib\pipeline.js:249:19)
    at Pipeline.pipeline.exec (\node_modules\ioredis\lib\transaction.js:34:26)
    at \node_modules\bull\lib\job.js:210:18
```

Apparently the pipelines are not compatible with cluster mode of redis.

Can you confirm this problem?
How to use bull with a cluster redis without encountering this problem?

manast:
ok, seems like it is a limitation of ioredis at least. I did not now about it: https://github.com/luin/ioredis/issues/536
This requires some rewrite in order to make it work in cluster mode unfortunately.

ajwootto:
@manast we are starting to run into this. Do you know what features specifically require pipelines? Is there some subset of features we can still use without them?

btzsoft:
@manast I'm running in the same issue when calling `job.moveToFailed(err)`. it's possible to do something else here? it's strange for me, because I use `job.moveToCompleted()` without any problems. thx!

manast:
@btzsoft that is because ioredis does not support multi commands using custom lua scripts, and only ```moveToFailed```  currently uses that. In order to solve this issue the method should be rewritten to not use multi, which is a bit to work to do...

btzsoft:
@manast oh no, do you have some plans in future to change this? if no can you give pls some hints where to take a look, maybe I will make a fork and will changes this, bcs it's really a big blocker for us. thank you for your time.

manast:
you can check the implementation of moveToFailed, it is quite inconvenient to not being able to use multi in this case. The other alternative is to convince the authors of ioredis to fix this issue:
https://github.com/luin/ioredis/issues/536

btzsoft:
@manast just to better understand, why `moveToFailed` use multi command? and why doesn't work as `moveToFinished`. thx!

manast:
[...]

PR #1933, opened 2020-12-12
fix: resolve repeatable test suite breakage with ioredis 4.19.0+

Rua-Yuki:
This PR resolves failures in the [repeatable job](https://github.com/OptimalBits/bull/blob/dbc08f23e93147c5c6b2b04fdad5a0c92d746654/test/test_repeat.js#L16) test suite brought about by upgrading [`ioredis`](https://github.com/luin/ioredis) to 4.19.0+, as mentioned in the comments for https://github.com/OptimalBits/bull/commit/6fe29283e85a64c07586fe8463de065ae7480513 and https://github.com/OptimalBits/bull/issues/758.

The precise change in `ioredis` which yields breakage is the introduction of a new script cleanup helper interval, intended to periodically purge a dictionary used for tracking scripts loaded for the purpose of pipelining. See [here](https://github.com/luin/ioredis/blob/d94f97d6950035818a666c08447a9d5e0ef5f8c7/lib/redis/index.ts#L299-L301), or observe the excerpt below.

```ts
      this._addedScriptHashesCleanInterval = setInterval(() => {
        this._addedScriptHashes = {};
      }, this.options.maxScriptsCachingTime);
```

Nothing is intrinsically wrong with what `ioredis` has done here. We only see breakage within the Bull tests due to a chance interaction with the new cleanup interval, brought about by extremely taxing usage of [`sinon`](https://github.com/sinonjs/sinon).

During the test setup phase, `sinon` is used to spoof timing functions, applying a base time equivalent to the Unix epoch.
https://github.com/OptimalBits/bull/blob/dbc08f23e93147c5c6b2b04fdad5a0c92d746654/test/test_repeat.js#L21

Within some tests, the same `sinon` clock is advanced by massive amounts at once (over 47 years).
https://github.com/OptimalBits/bull/blob/dbc08f23e93147c5c6b2b04fdad5a0c92d746654/test/test_repeat.js#L200-L201

These `clock.tick(...)` calls result in a punishing number of executions (around 25 million) of the aforementioned interval. This all takes quite a while, leading to timeouts in the end, and ultimately being the root of test failures.

The solution for this thankfully appears to be a simple one: merely swapping out `clock.tick` with `clock.setSystemTime`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "cron-parser": "^2.13.0",
     "debuglog": "^1.0.0",
     "get-port": "^5.1.1",
-    "ioredis": "4.18.0",
+    "ioredis": "^4.19.2",
     "lodash": "^4.17.19",
     "p-timeout": "^3.2.0",
     "promise.prototype.finally": "^3.1.2",
--- a/test/test_repeat.js
+++ b/test/test_repeat.js
@@ -198,7 +198,7 @@ describe('repeat', () => {
     this.timeout(20000);
     const _this = this;
     const date = new Date('2017-02-07 9:24:00');
-    this.clock.tick(date.getTime());
+    this.clock.setSystemTime(date);
     const nextTick = 2 * ONE_SECOND + 500;
 
     queue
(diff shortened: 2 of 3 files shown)


manast:
Great!, thanks for the fix!.

PR #1938, opened 2020-12-15
fix: promote jobs to the right "list" when paused

### LINKED ISSUE #1935: Promote does not respect paused queues.
We should port the fix from here: https://github.com/taskforcesh/bullmq/pull/338

ersoma:
Fixes #1935

See https://github.com/taskforcesh/bullmq/pull/338 for more details


manast:
wow, great, thanks for backporting!.

PR #1940, opened 2020-12-16
feat: add getPrioritizedJobCounts

### LINKED ISSUE #1939: Per-priority metrics
Hey! I would like to make a PR implementing a way to query job counts per priority. Any recommendations or ideas on the best way to proceed?

Right now I'm thinking of a `getPrioritizedJobCounts` that returns a `Record<Priority, Count>` object.

wyozi:
Adds a method to get prioritized job counts. Currently the function requires passing all the priorities you want to query for, since priorities can be pretty much anything.

Fixes #1939

### DIFF
--- a/lib/getters.js
+++ b/lib/getters.js
@@ -70,6 +70,28 @@ module.exports = function(Queue) {
     });
   };
 
+  /**
+   * Returns waiting job counts for each of the given priorities.
+   * By nature this only counts jobs currently in waiting state.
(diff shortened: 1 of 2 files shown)


manast:
Sorry for late response. This PR could be merged if 1) the "priorities" array was optional (or maybe not needed at all), and 2) a piece of documentation in REFERENCE.md

PR #1948, opened 2021-01-08
async version of process.send for master progress and log

bakkerthehacker:
Rather than use an empty promise and ignoring the error from `process.send()`, the async version is used and returned directly for both master progress and logs.  This allows the function calling progress/logs to receive the rejection rather than having it emitted somewhere else.


manast:
thanks!

PR #1994, opened 2021-03-05
fix(queue): avoid null stall event when other worker completes job #1933

alexburke-peakon:


manast:
this one is tricky. So this is the case where the job get stalled but processed so quickly by some worker that the job is not there anymore... of course it could also be argued that the in the proposed solution the job was indeed stalled but we never emitted the event :/. gotta need to sleep about this.

alexburke-peakon:
> this one is tricky. So this is the case where the job get stalled but processed so quickly by some worker that the job is not there anymore... of course it could also be argued that the in the proposed solution the job was indeed stalled but we never emitted the event :/. gotta need to sleep about this.

Yeah, I think that's it - the job was stalled but before the event can be emitted another worker completes it. As you mention, there is an argument that the job stalled but no event comes out with this change ..as things stand though, you still don't get any useful information if the code tries to emit this event because it comes back as null.

But please do take your time to decide how is best to handle it and then just let me know - though I'm not sure I can be much help in terms of a PR if you'd prefer to address this on the lua side.

PR #2040, opened 2021-05-07
Stalled job failed

svet93:
The issue is that today when a job stalls more than the allowed number of times, it gets moved to failed, but the event doesn't get published. We noticed that because we were running a job that would start other child jobs and use the `.finished()` function in order to know when the work is done. However, if a job stalled, it never ended in the `global:failed` handler so the `finished` function never resolved.

If a job has stalled more than the allowable limit, fail and publish failure


lib/commands/moveStalledJobsToWait-7.lua
```diff
@@ -62,6 +62,7 @@ if(#stalling > 0) then
         if(stalledCount > MAX_STALLED_JOB_COUNT) then
           rcall("ZADD", KEYS[4], ARGV[3], jobId)
           rcall("HSET", jobKey, "failedReason", "job stalled more than allowable limit")
+          rcall("PUBLISH", KEYS[4], jobId)
```
manast, line 65: I think it would be better to create a json object in lua, its pretty easy (untested):;
```lua
rcall("PUBLISH", KEYS[4],  "{\"jobId\":\"" .. jobId .. "\", \"val\": \"job stalled more than maxStalledCount\"}")
```
That will also eliminate the special case handling you added in queue.js.

manast (changes requested):

PR #2084, opened 2021-07-01
fix: deep clone options (#2083)

### LINKED ISSUE #2083: Unexpected deletion of the keyPrefix option
## Description

In latest code, the instantiation of queue will deleted `keyPrefix` property of the options object which we passed to the second function parameter:

```js
if (_.isString(url)) {
  // ...
} else {
  opts = url || {};
}

// ...

delete opts.redis.keyPrefix;
```

It causes that we will loss the `keyPrefix` if we use one options instance to initialize multiple queues.

I think we should use a deep cloned options object to make sure the passed one will not be changed.

I wrote a test to reproduce the issue:
https://github.com/nekocode/bull/commit/06feb3176c85891b7180b72ea7c20d3a4514c089

```js
it('should not change the options object', async () => {
  const originalOptions = { redis: { keyPrefix: 'myQ' } };
  const options = _.cloneDeep(originalOptions);

  let queue = new Queue('q', 'redis://127.0.0.1', options);
  expect(_.isEqual(options, originalOptions)).to.be.true;
  await queue.close();

  queue = new Queue('q', options);
  expect(_.isEqual(options, originalOptions)).to.be.true;
  await queue.close();
});
```

## Bull version
3.22.9 (latest)

nekocode:
Fixed #2083

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -90,12 +90,13 @@ const Queue = function Queue(name, url, opts) {
   }
 
   if (_.isString(url)) {
+    const clonedOpts = _.cloneDeep(opts || {});
     opts = {
-      ...opts,
-      redis: { ...redisOptsFromUrl(url), ...(opts || {}).redis }
+      ...clonedOpts,
+      redis: { ...redisOptsFromUrl(url), ...clonedOpts.redis }
     };
   } else {
-    opts = url || {};
+    opts = _.cloneDeep(url || {});
   }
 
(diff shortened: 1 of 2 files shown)


nekocode:
Could you help review this PR? @manast

manast:
looks good, thanks for the fix!

nekocode:
I'm glad it helped 😄

PR #2085, opened 2021-07-06
Adding support to ESM sandboxed jobs

mcolmant:
Hello,

Thanks for all the work you did on bull.

As NodeJS 10 is now dead since April, I propose to correctly support ESM sandboxed jobs.
It will be a big improvement for us as we are moving to ESM modules.

Let me know if you need something else.


Sparksx:
Nice work @mcolmant

manast:
Thanks for the nice PR. Just wondering if this change will break existing deployments that are still using node 10. If so it will be a breaking change and we need to release a major version. I guess the tests fails in node 10 and thats why you removed them from the CI script?

mcolmant:
Hello @manast, thanks for your quick answer.
I pushed a new commit to fix few errors I made.

Yes, you're right, the import statements are not supported in node 10.
Moreover, this version is now ended as quoted from [NodeJS](https://nodejs.org/en/about/releases/).

You're right, it should be a major release if it's accepted.

manast:
Gonna delay the merge of this one for a while  because there are still some more stuff that I want to incorporate into 3.x.

mcolmant:
Ok, I understand. I will keep this fork updated until it can be released and will use it for now.
Could I propose the same changes for bullmq?

manast:
> Ok, I understand. I will keep this fork updated until it can be released and will use it for now.
> Could I propose the same changes for bullmq?

Sure. Since it is typescript maybe it will not be a breaking change, not sure.

Issue #1861, opened 2020-09-16
Using Bull within ES6 module


gielow:
## Description
I have a certain repo configured as a module (type: "module" in package.json).
The application works all fine, imports, etc...

While trying to initialize/add the workers I have the following error:

`Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /some_path/processor-xxxxx.js`

## Additional information

I setup my processors:
queue.process(<number of workers>, </some_path/processor-xxxx.js>);

Is there anything wrong with my project setup?

PR #2086, opened 2021-07-06
fix: allow cjs extension for external processors (#1861)

fredriklindberg:
By allowing .cjs extensions it's possible to use Bull within an ES module with an external processor.

The obvious restriction is that the external processor would have to be a CommonJS file instead of ESM,
but existing ES modules within the same code base can be used inside the CJS module through import().


fredriklindberg:
Already done better in 2085 :)

PR #2093, opened 2021-07-14
docs: remove examples folder since it's outdated

DavideViolante:
Remove examples folder because it's:
- outdated by 3 years
- misleading
- not referenced anywhere


manast:
Thanks for the cleanup.

PR #2095, opened 2021-07-14
docs: prefer const + linting

DavideViolante:


manast:
Thanks for the nice PR!

PR #2105, opened 2021-07-16
chore(deps): upgrade husky to v7 and add 2 hooks + drop Node v10 support

DavideViolante:
- Upgrade to latest husky version (used `npx husky-init && npm i`)
- Add 2 hooks: `pre-commit` (linting and prettier) and `commit-msg` as before


.husky/commit-msg
```diff
@@ -0,0 +1 @@
+npx --no-install commitlint --edit $1
```
DavideViolante, line 1: From: https://typicode.github.io/husky/#/?id=husky_git_params-ie-commitlint-

DavideViolante:
Just noted that Node v10 is not supported by latest husky, so I dropped support for it in this PR.
Note that Node v10 is not officially supported anymore, so the change would be ok to bring along.

manast:
I think we are going to need a new branch for breaking changes that we can merge when we have a handful of them, there is another PR with a breaking change that was submitted recently (also due to node 10 not being supported)

Issue #1113, opened 2018-11-06
Named processors and per processor concurrency


alolis:
Hello,

From the [docs](https://github.com/OptimalBits/bull/blob/master/REFERENCE.md#queueprocess):

```
/***
 * For each named processor, concurrency stacks up, so any of these three process functions
 * can run with a concurrency of 125. To avoid this behaviour you need to create an own queue
 * for each process function.
 */
const loadBalancerQueue = new Queue('loadbalancer')
loadBalancerQueue.process('requestProfile', 100, requestProfile)
loadBalancerQueue.process('sendEmail', 25, sendEmail)
loadBalancerQueue.process('sendInvitation', 0, sendInvite)
```

It seems that this has been a major drawback, for my use case at least, and I was wondering why is there such a limitation? If it's just a matter of code refactoring then perhaps you can make it more clear to me and what needs to be changed in order for the concurrency per named processor to work properly so I can try to submit a PR.

My use case is one queue per user and each user queue has multiple named processors. However, I need the concurrency for each named processor to be respected instead of stacking up. 

Thank you for your time

manast:
I am not sure there can be defined a consistent behaviour by having a concurrency setting per named job, since you still have only one queue, lets say you have 2 named jobs, name A with concurrency 1, and B with concurrency 5, if you then in the queue you have something like this: BAAAAA. Would it be reasonable that B is not processed until A has been processed one by one (since concurrency is 1 for A), even if there is capacity left for processing B much earlier?

alolis:
Hm, I understand what you are saying but, if i had A with concurrency 1 and B with concurrency 5, I would expect if i added BAAAAA to run the B and 1 A in "parallel"...or maybe at least it would be up to the caller to decide if they should stack up or not. Maybe to achieve what I am saying an internal queue (aka a simple array or [queue](https://www.npmjs.com/package/queue)) can be used and `getNextJob`, check if the concurrency for the specific named processor is filled, and push it back to the queue if yes, or process it if not.

There are definitely cases where stacking up is not wanted. For example in my case where i set 2 for A and 2 for B, I do not want to run more than 2 named processors of type A in order to avoid hitting the rate limit of an API which my named processor is using underneath  but with stacking up, if i add 4 A jobs, then they all run and i am hitting the rate limit.

manast:
The behaviour you are requesting is to basically have two queues, one for A and one for B. Having internal queues in the worker is very complex if you want to achieve a consistent and robust system...

[...]

Issue #1098, opened 2018-10-23
Error: Could not remove job


alolis:
## Description

I am experiencing the following error whenever I try to remove a job:

```
2018-10-23T14:08:46.619Z - debug: Error: Could not remove job 1
    at /myproject/node_modules/bull/lib/job.js:392:15
From previous event:
    at /myproject/node_modules/bull/lib/job.js:388:42
    at runCallback (timers.js:637:20)
    at tryOnImmediate (timers.js:610:5)
    at processImmediate [as _immediateCallback] (timers.js:582:5)
```

## Minimal, Working Test code to reproduce the issue.

I am using the sandboxed environment and the issue is reproducible even with a simple processor like so:

```
// processor.js
module.exports = function myProcessor(job, done) {
  setTimeout(() => {
    done();
  }, 60000);
}

// normal code
import Queue from 'bull';

const queue = new Queue('myQueue');

queue.process('/path/to/my/processor.js');
await queue.add({foo: 'bar'}, {uuid: 1});

const job = await queue.getJob(1);
await job.remove(); // error here
```

## Bull version

I am using **Node.js v6.9.1** with **bull v3.4.8** and **redis-server v3.0.7**

Also by looking at [this](https://github.com/OptimalBits/bull/blob/master/lib/commands/removeJob-8.lua#L2) I understand that I can remove jobs while in progress. Is this true? What is the correct way to force remove a job if indeed true?

Thank you for your time.

manast:
you cannot remove a job white is is being processed, a job is in the active state and locked until it either: completes, fails or stales, although it should never stale if using a sandbox processor in practice.

alolis:
@manast , thanks for getting back to me. Isn't there some workaround I can do without destroying the library? I find it very strange that nobody has requested this functionality!

alolis:
UPDATE:

I tried to do the following which indeed removes the job from Redis but does not kill the processor process.

```
async remove() {
    await this.releaseLock();
    return super.remove();
 }
```

I didn't like it anyway because I do not want to mess with the locks.

The only way to do this probably is to signal via Redis pub/sub events and `throw new Error('removed')` in the processor to exit as it would normally do.....unless there is something else I could use which I am all ears!

manast:
We have some discussion regarding aborting jobs here: https://github.com/OptimalBits/bull/issues/114
[...]

PR #2117, opened 2021-07-26
feat: support job.discard function in sandboxed processors

alolis:
* Implement Job.prototype.isDiscarded

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -419,6 +419,10 @@ Job.prototype.isStuck = function() {
   });
 };
 
+Job.prototype.isDiscarded = function() {
+  return this._discarded;
+};
+
 Job.prototype.getState = function() {
   const fns = [
     { fn: 'isCompleted', state: 'completed' },
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -183,5 +183,13 @@ function wrapJob(job) {
     });
     job.data = data;
   };
+  /*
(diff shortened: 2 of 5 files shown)


manast:
I am going to merge this because I understand this feature is useful to you despite I am not very happy with the "discard" functionality to begin with since it is a bit hackish in the sense that it is just a local property of the job and not persisted which can lead to surprising behaviours.

alolis:
@manast, thanks for the speedy handling of the PR! 

I understand your concern. It did cross my mind as well but then I thought that the people who use this library are engineers and they should know better to not mess with things they do not fully understand yet :) 

My intentions for this PR will be more clear as soon as I reply to the comments [here](https://github.com/OptimalBits/bull/issues/1098#issuecomment-776918687) and [here](https://github.com/OptimalBits/bull/issues/1113#issuecomment-851642543).

If you do feel that this PR needs to be reverted however in case you've changed your mind, feel free to do so of course!

alolis:
@manast, [this](https://www.alexanderlolis.com/riding-the-bull), explains why I needed the function.

I hope you approve this article. There are a few people in issues that have been asking me about my approach so I thought to just write up everything.

manast:
@alolis that was a very informative article, always helpful to understand what pain points are there for bull users so that I can try to mitigate them with new features.

alolis:
Cheers @manast! I am glad you liked it

Issue #2120, opened 2021-07-30
Queue is missing redis username support from url


StefanCenusa:
## Description

Queue doesn't parse the username out of the redis url. Of course, username could be passed with redis options, but I think it would be nice to parse/get all the information from the url string if possible.

```
function redisOptsFromUrl(urlString) {
  const redisOpts = {};
  try {
    const redisUrl = url.parse(urlString);
    redisOpts.port = redisUrl.port || 6379;
    redisOpts.host = redisUrl.hostname;
    redisOpts.db = redisUrl.pathname ? redisUrl.pathname.split('/')[1] : 0;
    if (redisUrl.auth) {
      redisOpts.password = redisUrl.auth.split(':')[1];
    }
  } catch (e) {
    throw new Error(e.message);
  }
  return redisOpts;
}
```
https://github.com/OptimalBits/bull/blob/develop/lib/queue.js#L305

A simple solution would be:

```
if (redisUrl.auth) {
    const columnIndex = redisUrl.auth.indexOf(':');
    redisOpts.password = redisUrl.auth.slice(columnIndex + 1);
    if (columnIndex > 0) {
        redisOpts.user = redisUrl.auth.slice(0, columnIndex);
    }
}
```

Should I open a PR for this change?

## Bull version
3.27.0

manast:
Support for users is a new feature in redis so that is why it has not been implemented in Bull yet, you are welcome to submit a PR.

StefanCenusa:
@manast Thanks for your quick reply. I've submitted a PR for this change.

PR #2123, opened 2021-08-04
feat(queue): handle redis url containing username

StefanCenusa:
Implements #2120


manast:
Good job, thanks!

StefanCenusa:
@manast Actually, there is a typo. Since bull is using ioredis it's `username`, not `user`. 
https://github.com/luin/ioredis/blob/master/lib/redis/RedisOptions.ts#L18

PR #2152, opened 2021-09-02
fix(connection): fail only if redis connection does not recover

gmauricio:
On client initialization, Bull waits for the Redis client connection to be 'ready' before trying to send commands to it.
In order to do so, it keeps a promise that only resolves after the Redis client emits a 'ready' event.
It also listens for 'error' events, if an error comes, initialization fails.

However, having an error reported from Redis client, should not fail initialization immediately, because Redis client can retry the connection. Initialization should not fail on the first error it sees, but instead wait for the client to give up on reconnecting. That's why here I'm changing from listening to 'error's and instead listen to an 'end' event, which signals that the client won't try to reconnect anymore or that it failed to establish a connection to the server. See [Connection Events](https://github.com/luin/ioredis#connection-events).

For context, what I'm trying to fix is situations where a queue is instantiated before Redis server is ready, causing the instance to stay in a broken state (_initializing promise rejected and unrecoverable), even when the client is able to connect later, thanks to `ioredis` "auto-reconnect" feature.

I didn't remove the error listener entirely, it is still useful to keep track of seen errors, so when the 'end' occurs we have some information on why the connection failed, by exposing that last seen error when rejecting initialization promise.

Note: for testing, I passed a `retryStrategy` that always returns false, so no retries are made. By default `ioredis` will retry forever. This means that bull users should pay extra care on connection error handling and probably play with `retryStrategy` option, for example to limit the number of retries. This will change expected behavior, since before this, calling `isReady` on a queue where the connection could not be immediately established resulted in a error, but now, by default, it won't crash but hold indefinetely until a connection can be established, unless number of retries are limited.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -19,17 +19,25 @@ function isRedisReady(client) {
       resolve();
     } else {
       function handleReady() {
+        client.removeListener('end', handleEnd);
         client.removeListener('error', handleError);
         resolve();
       }
 
+      let lastError;
       function handleError(err) {
+        lastError = err;
+      }
+
+      function handleEnd() {
         client.removeListener('ready', handleReady);
-        reject(err);
+        client.removeListener('error', handleError);
+        reject(lastError);
       }
 
       client.once('ready', handleReady);
-      client.once('error', handleError);
+      client.on('error', handleError);
+      client.once('end', handleEnd);
     }
   });
 }
--- a/test/test_connection.js
+++ b/test/test_connection.js
@@ -123,11 +123,12 @@ describe('connection', () => {
       });
   });
 
-  it('should fail if redis connection fails', done => {
+  it('should fail if redis connection fails and does not reconnect', done => {
     queue = utils.buildQueue('connection fail', {
       redis: {
         host: 'localhost',
-        port: 1234
+        port: 1234,
+        retryStrategy: () => false
       }
     });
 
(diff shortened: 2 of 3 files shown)


lib/utils.js
```diff
@@ -19,17 +19,25 @@ function isRedisReady(client) {
...
+      let lastError;
       function handleError(err) {
+        lastError = err;
+      }
+
+      function handleEnd() {
         client.removeListener('ready', handleReady);
-        reject(err);
+        client.removeListener('error', handleError);
+        reject(lastError);
```
manast, line 35: What if the queue was closed manually?,wouldn't it also reject this promise? Maybe it should only reject if lastError is not undefined, otherwise just resolve.

manast (changes requested):
Looks good, please check my comment!

lib/utils.js
gmauricio: Makes sense. I'll check and let you know.

gmauricio: I'm thinking it should reject still, even when the connection is manually closed.
Considering this is a very particular case where the 'end' listener will be called if the connection never got to a 'ready' state in the first place.
If connection ends, then pending initialization should still fail I think, since the connection was never established. Otherwise, anything waiting on this, after resolved, will try to send commands to Redis on a closed connection.
What do you think ?

manast: yeah, I guess it is safer to reject to avoid other corner cases, although some people will complain that they get errors when closing too fast in other corner cases :)

manast (approved):

PR #2184, opened 2021-10-23
feat: emit event on job lock extend failure

bewithjonam:
Existing: When a job fails to extend its lock in redis, there is no way to debug or do find it out.

Change: When a job fails to extend its lock in redis, a new event `lockExtendingFailed` will be emitted.

Reason: In our use-case, each job processes data from a socket continuously and sends signals based on it when the jobs conditions are met. It is a long job(runs for days). We have multiple workers processing jobs. As already clearly mentioned in the Readme, if lock extension fails, it will lead to double processing of jobs.

Using this event, if extending lock fails, I will disconnect the socket so that it will stop duplicate processing.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -955,7 +955,7 @@ A queue emits also some useful events:
   // workers that crash or pause the event loop.
 })
 
-.on('lockExtendingFailed', function (job) {
+.on('lock-extension-failed', function (job) {
   // A job failed to extend lock. This will be useful to debug redis 
   // connection issues and jobs getting restarted because workers
   // are not able to extend locks.
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -955,6 +955,12 @@ A queue emits also some useful events:
   // workers that crash or pause the event loop.
 })
 
+.on('lockExtendingFailed', function (job) {
+  // A job failed to extend lock. This will be useful to debug redis 
+  // connection issues and jobs getting restarted because workers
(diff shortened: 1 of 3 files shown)


bewithjonam:
Renamed it. 👍

REFERENCE.md
```diff
@@ -955,6 +955,12 @@ A queue emits also some useful events:
   // workers that crash or pause the event loop.
 })
 
+.on('lock-extension-failed', function (job) {
```
manast, line 958: I noticed that there is another argument, "err" that is missing in this documentation.

test/test_events.js
```diff
@@ -211,6 +211,26 @@ describe('events', () => {
...
   });
 
+  it('should emit an event if a job fails to extend lock', done => {
+    const LOCK_RENEW_TIME = 1;
+    queue = utils.buildQueue('queue fails to extend lock', {
+      settings: {
+        lockRenewTime: LOCK_RENEW_TIME
+      }
+    });
+    queue.once('lock-extension-failed', lockingFailedJob => {
```
manast, line 221: it would be great to also include the err object in this test and check that the error matches the expected error (like the expected error has the same error message).

manast (changes requested):

bewithjonam:
Not sure why the test is failing. Please let me know if I can do anything else to get this merged and released. 

Very much appreciate your help/work.

bewithjonam:
👍 👍 👍

Issue #2203, opened 2021-11-05
jobPromise.cancel() gets clobbered for jobs with timeouts and promises that aren't native Promises


cincodenada:
## Description

There are a couple situations where calling `jobPromise.cancel()` in the `active` event callback doesn't work:
 - If the job returns a promise that is not a native Promise (e.g. if using Bluebird or another cancelable promise library)
 -  If the job is given a timeout

## Minimal, Working Test code to reproduce the issue.

```js
const Queue = require('bull')
const delay = require('delay')
const Bluebird = require('bluebird')
Bluebird.config({ cancellation: true })

// A simple demo processor that runs for `jobData.delay` milliseconds
// It provides a cancelable promise, and stops processing early if canceled
async function tickProgress (job, state) {
  for (let i = 0; i < 10; i++) {
    if (state.canceled) {
      console.log('Job canceled!')
      return
    }
    job.progress(i * 10)
    await delay(job.data.delay / 10)
  }
  job.progress(100)
}

function bluebirdProcessor (job) {
  return new Bluebird((resolve, reject, onCancel) => {
    const state = { canceled: false }
    onCancel(() => { state.canceled = true })
    tickProgress(job, state).then(resolve).catch(reject)
  })
}

function nativeProcessor (job) {
  const state = { canceled: false }
  const promise = tickProgress(job, state)
  promise.cancel = function () { state.canceled = true }
  return promise
}

// replace with nativeProcessor to un-break it
const processor = bluebirdProcessor

;(async () => {
  // Demonstrating that cancel works fine without the queue...
  console.log('Running job processor outside of queue...')
  const jobPromise = processor({
    data: { delay: 5000 },
    progress: progress => console.log('Progress:', progress)
  })
  // Wait a bit...
  await delay(2000)
  // Cancel
  jobPromise.cancel()
  // Wait for it to see the cancellation
  await delay(500)

  // Cancel does not work when wrapped in a queue job
  const queue = new Queue('test')
  queue.process(processor)
  queue.on('error', async (error) => {
    console.log('Queue error:', error)
  })
  queue.on('progress', function (job, progress) {
    console.log('Progress at', progress)
  })
  queue.on('completed', async (job, result) => {
    queue.close()
  })

  // When a job starts, wait 3.5 seconds and then attempt to cancel it
  queue.on('active', async (job, jobPromise) => {
    console.log('Job started:', jobPromise)
    await delay(2000)
    console.log('Attempting to cancel...')
[...]

PR #2204, opened 2021-11-05
fix(queue): fix cancelling jobs with timeouts or non-native Promises

cincodenada:
Proposed fix for #2203

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -670,7 +670,12 @@ Queue.prototype.setHandler = function(name, handler) {
     } else {
       this.handlers[name] = function() {
         try {
-          return Promise.resolve(handler.apply(null, arguments));
+          const maybeThenable = handler.apply(null, arguments)
+          if(maybeThenable.then) {
+            return maybeThenable
+          } else {
+            return Promise.resolve(maybeThenable)
+          }
         } catch (err) {
           return Promise.reject(err);
         }
@@ -1116,7 +1121,12 @@ Queue.prototype.processJob = function(job, notFetch = false) {
     let jobPromise = handler(job);
 
     if (timeoutMs) {
+      let cancelHandler
+      if(jobPromise.cancel) {
+        cancelHandler = jobPromise.cancel.bind(jobPromise)
+      }
       jobPromise = pTimeout(jobPromise, timeoutMs);
+      jobPromise.cancel = cancelHandler
     }
 
     // Local event with jobPromise so that we can cancel job.
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -1062,6 +1062,59 @@ describe('Queue', () => {
       });
     });
 
+    describe('canceling jobs that return cancelable promises', () => {
+      beforeEach(() => {
+        queue.process('native', (job) => {
+          job.progress('not canceled')
+          const promise = delay(20)
+          promise.cancel = function() { job.progress('canceled') }
+          return promise
+        });
+        queue.process('promiseLike', (job) => {
+          job.progress('not canceled')
+          const promise = delay(20)
+          return {
+            then: promise.then,
+            catch: promise.catch,
+            cancel: function() { job.progress('canceled') }
+          }
+        })
+
+        queue.on('active', async (job, jobPromise) => {
+          await delay(10)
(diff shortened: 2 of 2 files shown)


lib/queue.js
```diff
@@ -670,7 +670,12 @@ Queue.prototype.setHandler = function(name, handler) {
     } else {
       this.handlers[name] = function() {
         try {
-          return Promise.resolve(handler.apply(null, arguments));
+          const maybeThenable = handler.apply(null, arguments)
```
manast, line 673: Why is this change needed? According to the Promise specification, resolving a promise just chains that promise to the new promise, so there should not be any need to differentiate the two. This works the same as the "await" keyword, you can apply it to a value or to a promise, the result will always be the resolved value.

lib/queue.js
```diff
@@ -1116,7 +1121,12 @@ Queue.prototype.processJob = function(job, notFetch = false) {
     let jobPromise = handler(job);
 
     if (timeoutMs) {
+      let cancelHandler
```
manast, line 1124: Could you explain why this change solves the issue?

manast (commented):
Thanks for the PR.
I wrote some comments.

In the end the problem is that native Promises do not support cancelation, so the current implementation more a hack than anything else. However when using Bluebird cancellation should kind of work.

lib/queue.js
cincodenada: This is needed because, as you point out, native Promises do not support cancellation. That makes this a bit tricky. In the original code, if `handler` returns a Bluebird promise with a `cancel()` method, `Promise.resolve()` will return a native Promise that resolves to whatever the Bluebird promise did - but since it is now a native Promise and those don't support cancellation, the `cancel()` method doesn't get carried along and is lost, along with any other features that Bluebird promises have.

The reason it works for native Promises with a cancel method added is that in the special case where you call `Promise.resolve()` with a native Promise, instead of creating a new native Promise that resolves to what that promise resolves to, Node simply returns the original promise, so the added-on cancel method remains intact.

This change makes it so we do that for any promise-like object (that is, an object with a `then` method, hence "thenable", which I've seen promise-like objects referred to elsewhere). If the return value from the handler looks like a promise, then we return that value without wrapping it in a native promise, which preserves any `cancel()` methods. If it doesn't look like a promise, we still wrap it in a `Promise.resolve()` as before.

I can add a comment here with a brief note about why we're doing this manually, which might be helpful.

cincodenada: That got a bit long, so I'll try to sum up briefly: resolving a promise does chain that promise to a new promise, but it does so by creating a new promise that resolves to the same value as the old promise - so while the resolved value gets carried over, anything that is not supported in native promises (such as cancellation) does not.

lib/queue.js
cincodenada: `pTimeout`, while it cancels the promise that it wraps, doesn't support canceling the resulting promise, and doesn't give any access to the wrapped promise once created.

So, _before_ we wrap the promise, if there is a `cancel` handler on the promise returned from the job handler, this will retain a reference to it that is bound to the context of the original promise. I then attach that method to the promise that `pTimeout()` returns as a cancel handler, so that the end-user can call `cancel()` on the pTimeout-wrapped promise and it will behave exactly as if they called the cancel method on the original promise.

As I write this, I realize that this could (and probably will) result in the `cancel()` method getting called twice, since `pTimeout()` will eventually timeout and call it again.

I see a couple ways to resolve that: we could use pTimeout's `options.customTimers` to pass in a wrapper for setTimeout() that lets us cancel the timer in `cancelHandler()`, or have `cancelHandler` instead be a function that deletes itself from the pTimeout promise when it is called, so that `pTimeout` can't call it a second time. The first seems cleaner, I think.

lib/queue.js
manast: I understand. I think the best would be to refactor this as a special promiseResolve function that works the same for both native and non-native promises, then it will be easier to understand its intent as well.

lib/queue.js
manast: ok, so same comment as above, find a good descriptive name and refactor out to function so that the intent is clear and the code easier to follow.

PR #2237, opened 2021-12-16
fix(queue): enabled queues to share childPool instance

jfontaine-lifion:
My team use bull queues to create dynamic queues and noticed that we for every instance of a queue a childPool instance is create. This change allows multiple bull queues to share the same bull queue instance.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -217,7 +217,7 @@ const Queue = function Queue(name, url, opts) {
     retryProcessDelay: 5000,
     drainDelay: 5,
     backoffStrategies: {},
-    isSharedChildPool: false
+    isSharedChildPool: false,
   });
 
   this.settings.lockRenewTime =
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -71,6 +71,7 @@ const MINIMUM_REDIS_VERSION = '2.8.18';
       guardInterval?: number = 5000,
       retryProcessDelay?: number = 5000,
       drainDelay?: number = 5
+      isSharedChildPool?: boolean = false
     }
   }
 
@@ -215,7 +216,8 @@ const Queue = function Queue(name, url, opts) {
     guardInterval: 5000,
     retryProcessDelay: 5000,
     drainDelay: 5,
-    backoffStrategies: {}
+    backoffStrategies: {},
+    isSharedChildPool: false,
   });
 
   this.settings.lockRenewTime =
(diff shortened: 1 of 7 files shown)


mitdave88:
@jfontaine-lifion  This is awesome. Might save memory. Provided, not all queues process simultaneously.
@manast

lib/scripts.js
```diff
@@ -326,17 +326,13 @@ const scripts = {
...
-    duration,
-  ) {
+  extendLock(queue, jobId, duration) {
     return queue.client.extendLock([
       queue.toKey(jobId) + ':lock',
       queue.keys.stalled,
       queue.token,
       duration,
-      jobId,
+      jobId
```
manast, line 335: This is not following the current format standard, we use trailing commas (as the default in "prettier"): https://stackoverflow.com/questions/61370583/trailing-comma-after-last-line-in-object

test/test_queue.js
```diff
@@ -2837,11 +2848,14 @@ describe('Queue', () => {
         });
     });
 
-    it('should clean the number of jobs requested even if first jobs timestamp doesn\'t match', async () => {
+    it("should clean the number of jobs requested even if first jobs timestamp doesn't match", async () => {
```
manast, line 2851: We should use single quotes instead of double.

test/test_queue.js
```diff
@@ -2121,6 +2121,17 @@ describe('Queue', () => {
       queue.on('failed', cb);
       queue.on('error', done);
     });
+
+    it('should share child pool across all different queues created', async () => {
```
manast, line 2125: It would be great to also test the opposite behavior, i.e. that the pools are different if we are not specifying the isSharedChildPool.

test/test_queue.js
```diff
@@ -2853,7 +2867,7 @@ describe('Queue', () => {
       expect(len).to.be.eql(2);
     });
 
-    it('shouldn\'t clean anything if all jobs are in grace period', async () => {
+    it("shouldn't clean anything if all jobs are in grace period", async () => {
```
manast, line 2870: Use single quotes

lib/queue.js
```diff
@@ -215,7 +216,8 @@ const Queue = function Queue(name, url, opts) {
     guardInterval: 5000,
     retryProcessDelay: 5000,
     drainDelay: 5,
-    backoffStrategies: {}
+    backoffStrategies: {},
+    isSharedChildPool: false
```
manast, line 220: use trailing commas

manast (commented):
Thanks. This could be useful for other users. Please check my comments.

jfontaine-lifion:
> Thanks. This could be useful for other users. Please check my comments.

@manast  I applied the feedback and added some unit tests

PR #2263, opened 2022-01-20
docs: Update REFERENCE.md

stefanwille:
Clarify connection between backoff and attempts in JobOptions.


manast (approved):
Thanks for the clarification.

PR #2271, opened 2022-01-25
fix(queue): enable queues to init childPool when initialized

jfontaine-lifion:


manast:
Thanks for the PR, could you please elaborate what is the use for this feature?

jfontaine-lifion:
hi @manast We noticed that the childPool instance was lazy loaded when called when `queue.process` function is called with a string. We wanted a way for the childPool code to be available when the queue is created.

manast:
@jfontaine-lifion  I see, but for which purpose? The pool is only needed when using sandboxed processors.

jfontaine-lifion:
> @jfontaine-lifion I see, but for which purpose? The pool is only needed when using sandboxed processors.

@manast  Is there a way to init the sandbox process code once the code is enabled or we are committing to use it

Issue #854, opened 2018-02-05
Set the failed/completed jobs automatically removed after a period time


Tin-Nguyen:
Hi all,

Do we have an option in Bull allows to remove the failed/completed jobs in queue after a period time (1 day/week/month)? I think it's necessary because it can reduce the number items in Redis storage.

Thanks in advanced.
Tin

mscharley:
For what it's worth, we just run `queue.clean(3600 * 1000, "completed")` once a minute and this means we only have an hours worth of completed tasks at any point in time.

Tin-Nguyen:
@mscharley  thanks. But I'm wondering why we don't have a function to let us be able to set a permanent process for the queue to remove the jobs after a period time? Could we consider to have an improvement on this?

mscharley:
Sure, and I 👍 'd this issue too. Just letting people know of workarounds in the meantime.

aleccool213:
We store our job details elsewhere so this is just duplicated data being stored. This is a much wanted feature from me as well :)

sriharshakappala:
@mscharley I wanted to clear jobs from failed state and I tried `queue.clean(3600 * 1000, "failed")` and it didn't work. What's the right way?

axos88:
Without this feature the amount of data in this increases to infinity, unless one of the workarounds is implemented of course.

It would be really great if it was possible to retain statistics (number of jobs processed), without having to keep the actual job details around.

addisonElliott:
There is a `removeOnCompleted` & `removeOnFailed` option for each job that will allow removing once the job is complete or failed. This should work in doing what needs to be done.

As for retaining statistics for jobs without keeping the jobs around, I think that's relatively easy to do  in userland by tracking the 'completed' and 'failed' events on the queue. Just keep increment some value in redis.

axos88:
The `removeOnCompleted` and `removeOnFailed` would remove the information immediately. I'm looking for a solution that would keep that around for an amount of time, so that they can be reviewed for say 7 days, and then scrapped so they don't clog up the redis storage.

The problem with implementing the statistics in userland is that there is already statistics provided by bull, and that info would become incorrect when deleting (old) jobs.

manast:
@axos88 it does not remove it immediately, please check the documentation: https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueadd

axos88:
@manast you can keep around "X" number of items, but cannot set a time limit for them.
[...]

PR #2275, opened 2022-01-26
feat: add support for removeOn based on time

### LINKED ISSUE #2265: [Feature] removeOnComplete/Failed based on time
It is often the case that you want to keep completed jobs but just up until a given date instead of a fixed number of them.  Since the completed/finished jobs are currently stored in a zset sorted by timestamp, removing older jobs is quite straightforward.

We would possibly also want to be able to mix both total max with "up to a date", so that we can get both behaviours at the same time if wanted

manast:
Fixes #2265 and #854

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -290,14 +290,39 @@ interface JobOpts {
   // jobId is unique. If you attempt to add a job with an id that
   // already exists, it will not be added (see caveat below about repeatable jobs).
 
-  removeOnComplete: boolean | number; // If true, removes the job when it successfully
+  removeOnComplete: boolean | number | KeepJobs; // If true, removes the job when it successfully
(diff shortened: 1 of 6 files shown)


jasperkennis:
Hi @manast , I'm super excited for this update! One question, does @types/bull also require to be updated? It currently doesn't define the new option yet, see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/911cf7b01956b6555c39ac2cb8b6eee168d3274d/types/bull/index.d.ts#L425).

manast:
I do not maintain the typings sorry...

jasperkennis:
I understand, as a workaround `// @ts-ignore` can be used. I guess `@types/bull` will update eventually. At any rate, super thanks for this update.

PR #2309, opened 2022-02-24
feat: have Queue#clean consult job.{finishedOn,processedOn}

papandreou:
Background: In our bull installation we're cleaning out failed jobs after 24 hours by periodically running `await queue.clean(24 * 60 * 60 * 1000, 'failed');`. Yesterday we had a big workload and ran into some fairly troublesome jobs that had to be tried multiple times, but still failed in the end. This morning I was expecting that they were still around so I could make some tweaks and retry them, but I found that they had been cleaned out despite having been attempted within the last 24 hours.

I looked into it and found that `Queue#clean` always compares [uses the job's `timestamp` when determining whether it falls within the grace period](https://github.com/OptimalBits/bull/blob/develop/lib/commands/cleanJobsInSet-3.lua#L63). That feels undesirable in a situation like the above. Seems fairly easy to also have it look at when the job was last processed and when it finished, so I went ahead and implemented that.

LMK what you think 🤗

### DIFF
--- a/lib/commands/cleanJobsInSet-3.lua
+++ b/lib/commands/cleanJobsInSet-3.lua
@@ -60,7 +60,20 @@ while ((limit <= 0 or deletedCount < limit) and next(jobIds, nil) ~= nil) do
 
     local jobKey = jobKeyPrefix .. jobId
     if (rcall("EXISTS", jobKey .. ":lock") == 0) then
-      jobTS = rcall("HGET", jobKey, "timestamp")
+      -- Find the right timestamp of the job to compare to maxTimestamp:
+      -- * finishedOn says when the job was completed, but it isn't set unless the job has actually completed
+      -- * processedOn represents when the job was last attempted, but it doesn't get populated until the job is first tried
+      -- * timestamp is the original job submission time
+      -- Fetch all three of these (in that order) and use the first one that is set so that we'll leave jobs that have been active within the grace period:
+      for _, ts in ipairs(rcall("HMGET", jobKey, "finishedOn", "processedOn", "timestamp")) do
+        if (ts) then
+          jobTS = ts
+          break
+        end
+      end
+      if (not jobTS) then
+        jobTS = rcall("HGET", jobKey, "timestamp")
+      end
       if (not jobTS or jobTS < maxTimestamp) then
         if isList then
           rcall("LREM", setKey, 0, jobId)
--- a/lib/commands/cleanJobsInSet-3.lua
+++ b/lib/commands/cleanJobsInSet-3.lua
@@ -71,9 +71,6 @@ while ((limit <= 0 or deletedCount < limit) and next(jobIds, nil) ~= nil) do
           break
         end
       end
-      if (not jobTS) then
-        jobTS = rcall("HGET", jobKey, "timestamp")
-      end
       if (not jobTS or jobTS < maxTimestamp) then
         if isList then
           rcall("LREM", setKey, 0, jobId)
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -2876,10 +2876,12 @@ describe('Queue', () => {
       queue.on(
         'completed',
         _.after(2, () => {
-          queue.clean(0).then(jobs => {
-            expect(jobs.length).to.be.eql(2);
-            done();
-          }, done);
+          delay(200).then(() => {
+            queue.clean(0).then(jobs => {
+              expect(jobs.length).to.be.eql(2);
+              done();
+            }, done);
+          });
         })
       );
     });
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -2911,6 +2911,29 @@ describe('Queue', () => {
         });
     });
 
+    it('should leave a job that was queued before but processed within the grace period', done => {
+      queue.process((job, jobDone) => {
+        jobDone();
+      });
+      queue.add({ some: 'data' });
+      queue.add({ some: 'data' }, { delay: 100 });
+      delay(200)
+        .then(() => {
+          // At this point both jobs have executed, but the delayed one was only processed about 100 milliseconds ago:
+          return queue.clean(150);
+        })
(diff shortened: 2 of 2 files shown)


manast:
The question is why are you using such a complicated setup instead of just setting the "removeOnFail" which supports both a maximum amount of jobs to keep as well as an age: https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueadd
```ts
  removeOnFail: boolean | number | KeepJobs; // If true, removes the job when it fails after all attempts. A number specified the amount of jobs to keep, see KeepJobs if using that interface instead.
  // Default behavior is to keep the job in the failed set.
```

manast:
Regarding your fix, I think it would be much easier if considering that the SCORE in the failed ZSET represents the timestamp when the job failed.

papandreou:
> The question is why are you using such a complicated setup instead of just setting the "removeOnFail" which supports both a maximum amount of jobs to keep as well as an age

Hmm, that's a good question! I wasn't aware of that option. It looks like that kinda does what I want, since `age` is compared to `finishedOn`. It seems like I could remove my current cleanup of `failed` and just specify `removeOnFail: {age: 24 * 60 * 60 }` whenever I queue a job.

Side-comment: It seems a bit weird for the `count` option to specified _per_ _job_ is a bit strange? Shouldn't that be a queue-wide setting? 🤔 

> Regarding your fix, I think it would be much easier if considering that the SCORE in the failed ZSET represents the timestamp when the job failed.

Hmm, yeah, that's a good point. That would only work when `isList` is false, though?

papandreou:
> > Regarding your fix, I think it would be much easier if considering that the SCORE in the failed ZSET represents the timestamp when the job failed.
> 
> Hmm, yeah, that's a good point. That would only work when `isList` is false, though?

I tried doing that only for the ZSET case:

```diff
diff --git a/lib/commands/cleanJobsInSet-3.lua b/lib/commands/cleanJobsInSet-3.lua
index 8301059..fc3063b 100644
--- a/lib/commands/cleanJobsInSet-3.lua
+++ b/lib/commands/cleanJobsInSet-3.lua
@@ -65,11 +65,16 @@ while ((limit <= 0 or deletedCount < limit) and next(jobIds, nil) ~= nil) do
       -- * processedOn represents when the job was last attempted, but it doesn't get populated until the job is first tried
       -- * timestamp is the original job submission time
       -- Fetch all three of these (in that order) and use the first one that is set so that we'll leave jobs that have been active within the grace period:
-      for _, ts in ipairs(rcall("HMGET", jobKey, "finishedOn", "processedOn", "timestamp")) do
-        if (ts) then
-          jobTS = ts
-          break
+     if isList then
+        for _, ts in ipairs(rcall("HMGET", jobKey, "finishedOn", "processedOn", "timestamp")) do
+          if (ts) then
+            jobTS = ts
+            break
+          end
         end
+      else
+        jobTS = rcall("ZSCORE", setKey, jobId)
+        redis.log(redis.LOG_WARNING, "job score", jobTS, "vs maxTimestamp", maxTimestamp)
       end
       if (not jobTS or jobTS < maxTimestamp) then
         if isList then
```

But the redis log indicates that the score is in a different range than the timestamp:

```
1:M 24 Feb 2022 17:16:36.799 # job score 6740881414803458 vs maxTimestamp 1645722996765
```

manast:
> Side-comment: It seems a bit weird for the `count` option to specified _per_ _job_ is a bit strange? Shouldn't that be a queue-wide setting? 🤔

Because there are no queue settings, but you can define it once using "defaultJobOptions" on the queue instance.

manast:
> But the redis log indicates that the score is in a different range than the timestamp:

That's strange considering it is the same timestamp as in finishedOn: 
https://github.com/OptimalBits/bull/blob/develop/lib/commands/moveToFinished-8.lua#L63-L64

papandreou:
> > But the redis log indicates that the score is in a different range than the timestamp:
> 
> That's strange considering it is the same timestamp as in finishedOn: https://github.com/OptimalBits/bull/blob/develop/lib/commands/moveToFinished-8.lua#L63-L64

Ah, the test that failed and caused that log output is this one: https://github.com/OptimalBits/bull/blob/02ad90de6851ba7db117ae6d05f859d5cb5b819e/test/test_queue.js#L2620-L2635

Looks like the `timestamp` is being fiddled with for the `delayed` queue: https://github.com/OptimalBits/bull/blob/02ad90de6851ba7db117ae6d05f859d5cb5b819e/lib/scripts.js#L247-L249

Maybe it's better to avoid using the `ZSCORE`?

manast:
The timestamp in the delay ZSET is different yes.

papandreou:
> The timestamp in the delay ZSET is different yes.

Do you still think using the score is simpler then?

I'm giving `removeOnFail` a go in my app, seems to work great so far! However, I still think this is a good change, as it's a less surprising behavior for `Queue#clean`.

papandreou:
I can't really figure out why the `Queue Cleaner should clean two jobs from the queue` test is failing on CI consistently. It works for me locally, also with node.js 10 🤔

papandreou:
I'm using redis 6.0.5 locally, not sure which version gets picked up by CI?

papandreou:
Ping @manast? 🤗

manast:
@papandreou do I have any actions to do here?

manast:
I mean, it is failing in one of the tests related to your changes...

papandreou:
> @papandreou do I have any actions to do here?
> I mean, it is failing in one of the tests related to your changes...

I was asking for help with fixing that. It passes for me locally, also with node.js 10, so I'm a bit lost.

papandreou:
@manast, I think I figured it out with a bit of trial-and-error on CI 😇 

At first I thought I needed a maintainer to trigger a CI build, which made me doubt that I could figure it out, but seems like that's not the case 🍀

manast:
Good job!

PR #2329, opened 2022-03-21
fix: unlock job when moving it to delayed

papandreou:
Hi! For various reasons I'm using a rather high `lockDuration` setting in my app, and I noticed that jobs that fail an attempt and have to be moved to `delayed` remain locked. This means that I cannot easily remove such a job from our small homegrown queue control app, as it isn't the lock-taker.

When I tried to figure out what was going on, I noticed this comment: https://github.com/OptimalBits/bull/blob/61a87a8509bfc9227775084839544c444d69994d/lib/job.js#L309

... However, it seems like the job isn't actually being unlocked by `moveToDelayed-3`, so I tried fixing that 😇

### DIFF
--- a/lib/commands/moveToDelayed-3.lua
+++ b/lib/commands/moveToDelayed-3.lua
@@ -1,7 +1,7 @@
 --[[
   Moves job from active to delayed set.
 
-  Input: 
+  Input:
     KEYS[1] active key
     KEYS[2] delayed key
     KEYS[3] job key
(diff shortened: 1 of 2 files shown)


papandreou:
WDYT @manast? :hugs:

manast:
Sorry for the late reply. Yes, you are right, this was working mostly out of luck, thanks for the fix.

PR #2340, opened 2022-04-14
fix(stalled-jobs): move stalled jobs to wait in batches

plus-:
This is a backport of this fix from BullMQ https://github.com/taskforcesh/bullmq/pull/625

We currently encounter the following error in production:

> ERR Error running script (call to f_ff9c18634832b0b4115a19b4de5f4788a7cfbd4e): @user_script:81: user_script:81: too many results to unpack


manast:
Thanks for the backport and sorry for the delay merging it.

Issue #924, opened 2018-04-30
Separate processor with ES module syntax not working


AoEmaster:
## Description

Currently you can't use a processor coded with ES module syntax (export keyword)
It works by importing the job into index.mjs, but I want to separate processes.

## Test code to reproduce

**index.mjs**

```
let queue = new Queue('queue');
queue.process('/processor.mjs');
```

**processor.mjs**

```
export default async (job) => {
return 'job completed;
}
```

**result**
```
internal/modules/cjs/loader.js:712
    throw new ERR_REQUIRE_ESM(filename);
    ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /processor.mjs
```

## Bull version
latest

DaHoopster:
Would love to have ES6 module works for separate processes

yedhukrishnan:
Any update on this? I am still not able to use a processor with ES6 module.

wxs77577:
The reason is use `require` to import processor files, hope some guy to fix this, thanks a lot.

Sparksx:
up

0x6c23:
+1

PR #2341, opened 2022-04-15
fix(processor): support ESM processor

imathews:
Hi! I was attempting to use sandboxed processes in an ESM project, and was running into issues as described in #924 . This change seems to resolve the issue for me, and I believe should work for both ESM and CJS projects.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -2,7 +2,7 @@ env:
   node: true
 
 parserOptions:
-  ecmaVersion: 2018
+  ecmaVersion: 2020
 
 extends:
   - eslint:recommended
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -47,6 +47,8 @@ process.on('message', async msg => {
   switch (msg.cmd) {
     case 'init':
       try {
+        // Dynamic imports are only supported in node >= 14.
+        // By using dynamic imports, we can use both ESM and CJS processors.
         const NODE_MAJOR_VERSION = process.versions.node.split('.')[0];
         if (NODE_MAJOR_VERSION < 14) {
           processor = require(msg.value)
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -47,7 +47,12 @@ process.on('message', async msg => {
   switch (msg.cmd) {
     case 'init':
       try {
-        processor = await import(msg.value);
+        const NODE_MAJOR_VERSION = process.versions.node.split('.')[0];
+        if (NODE_MAJOR_VERSION < 14) {
+          processor = require(msg.value)
+        }else {
+          processor = await import(msg.value);
+        }
       } catch (err) {
         status = 'Errored';
         err.message = `Error loading process file ${msg.value}. ${err.message}`;
--- a/lib/process/master.js
+++ b/lib/process/master.js
@@ -43,11 +43,11 @@ async function waitForCurrentJobAndExit() {
 process.on('SIGTERM', waitForCurrentJobAndExit);
 process.on('SIGINT', waitForCurrentJobAndExit);
 
-process.on('message', msg => {
+process.on('message', async msg => {
   switch (msg.cmd) {
     case 'init':
       try {
-        processor = require(msg.value);
+        processor = await import(msg.value);
       } catch (err) {
         status = 'Errored';
         err.message = `Error loading process file ${msg.value}. ${err.message}`;
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "chai": "^4.2.0",
     "coveralls": "^3.1.0",
     "delay": "^4.3.0",
-    "eslint": "^7.4.0",
+    "eslint": "^8.17.0",
     "eslint-plugin-mocha": "^7.0.1",
     "eslint-plugin-node": "^8.0.1",
     "expect.js": "^0.3.1",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "chai": "^4.2.0",
     "coveralls": "^3.1.0",
     "delay": "^4.3.0",
-    "eslint": "^8.17.0",
+    "eslint": "^7.4.0",
     "eslint-plugin-mocha": "^7.0.1",
     "eslint-plugin-node": "^8.0.1",
     "expect.js": "^0.3.1",


manast:
Thanks for the PR, unfortunately, it does not seem to work with the current tests...

imathews:
Hi @manast - it looks like this was failing eslint tests because the eslint `parserOptions` didn't support dynamic imports. If we bump eslint parserOptions to `ecmaVersion: 2020` it looks like those tests run fine.

I also realized that dynamic imports aren't supported in node < 14 (and I believe this version of Bull supports back to v10), so I added a check to fallback to the previous `require` syntax for old versions of node. Hopefully this seems reasonable - with [more and more libraries](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) pushing ESM in the node ecosystem, it's becoming increasingly necessary for us to be able to write our processors in ESM.

JiiHu:
@imathews Any news on this? I would love to use this in production 😄

VladimirMikulic:
Hi @manast. I started migrating a codebase from CJS to ESM and realized that ESM isn't supported in Bull.
This change by @imathews is 100% backwards compatible + allows projects to transition to ESM.

[Dynamic imports](https://nodejs.org/api/esm.html#interoperability-with-commonjs) started being available with Node.js 13 (14 for LTS).
It is worth noting that dynamic import is an **expression** that can import **both CJS and ESM code**.

So if Node.js version is less than 13, the `if` semaphore will use require to load worker.
On the other hand, if Node.js version is greater than or equal to 13, we know that dynamic import expression is available and it will load just fine both CJS and ESM.

This is not a breaking change. 
Legacy projects will still be able to upgrade BullMQ to benefit from latest features and fixes while new projects can be started in ESM. 

Let me know if this explanation makes sense and if you need anything more.
I'd really appreciate if you could merge these few magic lines :)

manast:
We have tried several times with such updates and always ends up breaking for some users. So in theory it is not a breaking change, as many times before, but in practice, it will be. I have merged and reverted many such changes before.

manast:
Note, I wish NodeJS worked better in this regard, I am frustrated as everybody else.

manast:
This is the latest attempt that I had to revert (in BullMQ but the code is basically the same): https://github.com/taskforcesh/bullmq/pull/1384

imathews:
@manast I think the notable difference in this PR (compared to https://github.com/taskforcesh/bullmq/pull/1384) is that it performs a node version check before using any import semantics. So I can't imagine this would have any effect for users on previous versions of Node, and on more recent versions this code should be fully cross-compatible between ESM and CJS.

VladimirMikulic:
Yes, you're right @imathews. `await import` is an **expression** so having that line in an environment that doesn't support dynamic imports is fine - as longs as the statement is not reached. (the reason for if check)

The ESM issue mentioned in this https://github.com/taskforcesh/bullmq/issues/917 wasn't actually caused by https://github.com/taskforcesh/bullmq/pull/1384. The root cause was improper addition of `exports` field in package.json that this PR reverted https://github.com/taskforcesh/bullmq/pull/1388

manast:
Maybe you are right, but, I am not sure it is so simple. Depending on your package settings, Node could complain also, for example, if you are not using the "type: module" and try to use import in some version of NodeJS... so in order to be sure it would be needed to test all the combinations of the supported node versions and the package.json settings that affect the import behaviur.

VladimirMikulic:
@manast `await import` is an expression that can be used in CJS context. (i.e. without `"type": "module" being set)

VladimirMikulic:
https://nodejs.org/api/esm.html#import-expressions
"[Dynamic import()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) is supported in both CommonJS and ES modules. In CommonJS modules it can be used to load ES modules."

VladimirMikulic:
For anyone also browsing the thread and who also has to migrate the codebase to ESM, you can use this as a workaround:

In your package.json (for NPM; if you're on yarn 1.x check out [patch-package](https://www.npmjs.com/package/patch-package))
```json
  "scripts": {
    "postinstall": "sed -i 's/require(processorFile)/await import(processorFile)/g' ./node_modules/bullmq/dist/cjs/classes/child-processor.js"
  }
```

Now you can use ESM in your sandboxed worker files:
```js
import doSomeWork from "./myfile.js";

export default (job) => {
  console.log("Job received", job);
  doSomeWork(job);
}
```

jonaskello:
Also migrating a codebase from CJS to ESM and hit this problem. However I'm hesitant to do the monkey patch with sed :-). Could this be re-opened please? If breaking change is the only concern then could this just be put behind some kind of flag to fully keep old behavior as default? Like check an ENV variable BULL_ESM=true?

manast:
I don't think a flag is sufficient as Node itself can complain in some scenarios, but maybe I am wrong. The problem here is that this has bitten us so many times in the past I am kind of skeptical it can be solve without ruining it for others.

VladimirMikulic:
@manast I covered the first failed rollout in https://github.com/OptimalBits/bull/pull/2341#issuecomment-1340171274.
If you have some more examples from history, I'd be happy to dive in and find what went wrong.
It's also worth mentioning that we've been my `sed` workaround for almost three months now in production without any issues.

Riveer01:
> 

Hello,
I am trying to find the path you mentioned below, but did not find **bullmq**, since I am using bull 4.10.4, is there anything missing? Any help is highly appreciated.
` ./node_modules/bullmq/dist/cjs/classes/child-processor.js`

  "scripts": {
    "postinstall": "sed -i 's/require(processorFile)/await import(processorFile)/g' ./node_modules/bullmq/dist/cjs/classes/child-processor.js"
  }

VladimirMikulic:
@Riveer01 I use `bullmq` so for me the path includes `bullmq`, for you it's `bull`.

Riveer01:
@VladimirMikulic  , Thanks for your kind help.
I checked the related path, seems there is no file named child-processor.js in bull, could you please tell which files are the counterparts in bull? 
Thanks so much again.

Riveer01:
Hello @VladimirMikulic  , 
I have modified the master.js file, but there are still errors as below:
**Error loading process file /mnt/sfs/project/service/processor.js. Cannot use import statement outside a module, job:[object Object]**
processor.js is exactly the same as yours:
```
import doSomeWork from "./myfile.js";

export default (job) => {
  console.log("Job received", job);
  doSomeWork(job);
}
```
myfile.js:
```
const fs = require('fs');

export default function doSomeWork () {
  fs.writeFileSync('/mnt/sfs/project/service/222.txt', '555');
}
```
The bull version is 4.10.4, and need your further help, thanks a lot.

Riveer01:
@VladimirMikulic , 
Today I have switched to bullmq, but still got below error. BTW, I have modified the files you metioned, and also used [esm](https://github.com/standard-things/esm#readme) in the project, will this cause problems regarding this issue?
q2
```
// new Worker
const videoWorker = new Worker(`VideoQueue`, path.join(__dirname, 'processor.js'), {connection: videoRedisConnection});
```
```
// processor.js
import { SandboxedJob } from 'bullmq';

export default (job) => {
  console.log("Job received", job);
  // await job.log("Start processing job");
}
```

VladimirMikulic:
@Riveer01 make sure to set `"type": "module"` in your package.json.

Riveer01:
@VladimirMikulic ,
I have tried to set `"type": "module"` in the package.json of my project, but there is still error. 
I have also tried to set `"type": "module"` in bullmq's package.json, still error.

Riveer01:
@VladimirMikulic 

I followed your way and recreated a small project, and this time it worked! Thank you for your help!

VladimirMikulic:
You're welcome!

Riveer01:
@VladimirMikulic 
By the way, just would like to ask one question: on a machine with multi-core CPUs, if I set concurrency=3,  does  bullmq allocate the jobs to only one CPU, or possibly to  different CPUs？Great thanks.

Riveer01:
@VladimirMikulic 
By the way, just would like to ask one question: on a machine with multi-core CPUs, if I set concurrency=3,  does  bullmq allocate the jobs to only one CPU, or possibly to  different CPUs？Great thanks.

Riveer01:
@VladimirMikulic 
By the way, just would like to ask one more question: using a sandboxed processor like the code below, will the processes be created on the multiple different CPUs (multi-core CPUs), or just on the same CPU?
```
const myQueue = new Worker(queueName, `${__dirname}/workers/my-worker.js`, {
  connection,
});
```

VladimirMikulic:
@Riveer01 if the tasks are CPU intensive, then it's the best to provide concurrency = number of CPU threads.

Riveer01:
@VladimirMikulic 
By the "CPU threads", do you mean the jobs will be processed on the same CPU core? 
If I have a machine with multi-core CPUs as C1, C2, ... , C8, and the concurrency is set to 3, for the job j1, j2, j3, will they be possibly allocated to the different CPU cores, or just on the same core such as C3? We are using pm2 along with nodejs.
Thanks very very much!

VladimirMikulic:
It's up to the kernel how it allocates CPU power, but generally concurrency should be equal to number of CPU threads (virtual CPUs) your machine has for distributed load.

Riveer01:
@VladimirMikulic Thanks a lot!

Riveer01:
@VladimirMikulic 

Just would like to know if I have the queue working on machine A, and the worker is deployed on machine B, and if there are jobs added to the queue, how to make the woker (on different machine B) to fetch the jobs from the queue and process them normally?

Thanks a lot.

VladimirMikulic:
@Riveer01 I'm not sure I understand your question. BullMQ automatically distributes tasks among connected workers.

manast:
@Riveer01 as long as the workers are connected to the same Redis instance you do not need to do anything, they will pick the jobs from the same queue.

Riveer01:
@manast Thanks a lot, I will try.

Riveer01:
@manast Thank you very much, my two workers successfully connected to the same redis, I wonder if they will get jobs equally?

manast:
@Riveer01 They will pick new jobs as they process the jobs, so how they get distributed among the workers depends on the kind of job they are doing, network latencies, Redis usage, and so on.

PR #2380, opened 2022-06-16
fix(worker): better disconnect when blocking connection

### LINKED ISSUE #2372: why is queue.close() too slow.. and why is it not empty?
`queue.close()` takes almost `0.5 secs` to complete and debugging further, I found that `line #1269` is being slow (inside `whenCurrentJobsFinished` which is being called by `.close()`).

https://github.com/OptimalBits/bull/blob/5649689531feef18956175e0db638592664279d3/lib/queue.js#L1265-L1270

I splitted the line to measure the time elapsed as follows:

```javascript
const t1 = new Date().getTime();
await Promise.all(this.processing);  
console.log("elapsed: ", (new Date().getTime() - t1)/1000);  // this reports almost 0.5 seconds!
return forcedReconnection;
```

Interestingly, my service does not add any job to the queue, yet `this.processing.length` is `4`. Why is it not  `0`?

I"m using `v3.22.0` but this code has not changed since then. Not sure if the other part of codebase makes this efficient in the recent version. I'm planning to use `bullmq` (`nestjs/bullmq`) instead. Just want to know if this slowness is known and acceptable?

manast:
Possibly fixes #2372

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -570,7 +570,11 @@ async function redisClientDisconnect(client) {
 }
 
 Queue.prototype.disconnect = async function() {
-  await Promise.all(this.clients.map(redisClientDisconnect));
+  await Promise.all(
+    this.clients.map(client =>
+      client.blocked ? client.disconnect() : redisClientDisconnect(client)
+    )
+  );
 };
 
 Queue.prototype.removeJobs = function(pattern) {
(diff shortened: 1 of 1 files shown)


lib/queue.js
```diff
@@ -570,7 +570,11 @@ async function redisClientDisconnect(client) {
 }
 
 Queue.prototype.disconnect = async function() {
-  await Promise.all(this.clients.map(redisClientDisconnect));
+  await Promise.all(
+    this.clients.map(client =>
+      client.blocked ? client.disconnect() : redisClientDisconnect(client)
```
snawaz, line 575: Not sure if that will [fix the issue][1]. I'll do experiment and see.

BTW, I guess this logic should move to `redisClientDisconnect()` itself, so that if that is called from anywhere else, then the logic still applies. I'm not 10% sure though whether that'd be a good idea.. 🤔 

[1]: https://github.com/OptimalBits/bull/issues/2372

manast: It is the opposite actually, it must only be used in this particular case.

Issue #416, opened 2016-12-19
-OOM command not allowed when used memory > 'maxmemory'. 


n3trino:
1|wover    | ReplyError: ERR Error running script (call to f_bed0f540aaab40cb5b63c076c00c75504e27424b): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.   
1|wover    |     at JavascriptReplyParser.returnError (/home/apiUser/wover-backend/node_modules/ioredis/lib/redis/parser.js:25:25)
1|wover    |     at JavascriptReplyParser.run (/home/apiUser/wover-backend/node_modules/ioredis/node_modules/redis-parser/lib/javascript.js:135:18)
1|wover    |     at JavascriptReplyParser.execute (/home/apiUser/wover-backend/node_modules/ioredis/node_modules/redis-parser/lib/javascript.js:112:10)
1|wover    |     at Socket.<anonymous> (/home/apiUser/wover-backend/node_modules/ioredis/lib/redis/event_handler.js:107:22)
1|wover    |     at emitOne (events.js:96:13)
1|wover    |     at Socket.emit (events.js:188:7)
1|wover    |     at readableAddChunk (_stream_readable.js:176:18)
1|wover    |     at Socket.Readable.push (_stream_readable.js:134:10)
1|wover    |     at TCP.onread (net.js:548:20)

manast:
I have no idea what this can be. But it seems like a limitation with redis. We should then document that maxmemory can not be used with bull.

jsarenik:
You may need to check out the `removeOnComplete` option. See [the documentation](https://github.com/OptimalBits/bull/blob/master/REFERENCE.md#queueadd) for more info.

If you implement your own `.on('failed', ...` then also `removeOnFail` should be set to `1`.

Otherwise redis is being filled with all the old jobs in the completed state.

jsarenik:
@manast The default `maxmemory-policy` is `noeviction` which just returns an error on write operations. That may be what @n3trino sees.

@n3trino could you please try to run redis with `maxmemory-policy allkeys-lru`? Of course it is not the solution, I just want to check if that error is caused by errors from redis because the memory is full and `noeviction` is set. Thank you.

rinogo:
Our production ElastiCache has run out of memory due to running with Bull and Redis defaults. It sounds like we either need to remove completed and failed jobs or change our configuration to use a non-default `maxmemory-policy` (probably `allkeys-lru`).

[...]

PR #2439, opened 2022-10-07
Add a mention to prevent redis from running out of memory

half-shell:
As mentionned in https://github.com/OptimalBits/bull/issues/416, this PR makes a small addition in `REFERENCE.md` to suggest using fallback `removeOnComplete` and `removeOnFailed` options when creating a new queue to avoid having a redis instance running out of memory because of Bull's job tracking.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -275,6 +275,9 @@ An optional name can be added, so that only process functions defined for that n
 **Note:**
 You need to define _processors_ for all the named jobs that you add to your queue or the queue will complain that you are missing a processor for the given job, unless you use the `*` as job name when defining the processor.
 
+**Note:**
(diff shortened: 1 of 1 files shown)


REFERENCE.md
```diff
@@ -275,6 +275,9 @@ An optional name can be added, so that only process functions defined for that n
 **Note:**
 You need to define _processors_ for all the named jobs that you add to your queue or the queue will complain that you are missing a processor for the given job, unless you use the `*` as job name when defining the processor.
 
+**Note:**
+Considering all jobs in a terminal state (`failed` or `completed`) are stored in redis, depending on the number of jobs running and your redis setup, you might want to setup a default maximum number of jobs kept, using the `removeOnComplete` and `removeOnFailed` options when creating a queue so redis does not end up running out of memory.
```
manast, line 279: Instead of "terminal" it is better to use "finished" as this is the term we use internally.

manast: redis -> Redis

manast:
Thanks. Please see my comments.

PR #2464, opened 2022-11-14
Add getName attribute to bull's getters

roych98:
Found it useful when I wanted to debug queues, thought it might be helpful to others.


manast:
Thanks. Since we also include typescript typings, we need a typing for this new method as well before we can merge:
https://github.com/OptimalBits/bull/blob/develop/index.d.ts

PR #2478, opened 2022-11-24
fix: error is not handled

andrewvasilchuk:
The inline arrow function is created, which is not returned nor executed. It prevents the proper error handling.


manast:
Interesting. I wonder where that arrow function comes from, it must be a mistake.

andrewvasilchuk:
@manast it was introduced in this [commit](https://github.com/OptimalBits/bull/commit/4e224e5533f729b9781b1db81e1875b1bd50afb0) when `then` was replaced with `async/await`.

PR #2501, opened 2022-12-22
fix: typescript Job type

JonatanColussi:
Hi, i would like to send my contribution.

I noticed that when the job arrives at the consumer, some types are changed to string, such as Date and ObjectId from MongoDb

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -19,6 +19,7 @@
 //                 Levi Bostian <https://github.com/levibostian>
 //                 Todd Dukart <https://github.com/tdukart>
 //                 Mix <https://github.com/mnixry>
+//                 Jonatan Colussi <https://github.com/JonatanColussi>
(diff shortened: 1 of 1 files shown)


manast:
I am not sure I understand what you mean, could you write some code example to clarify please?

JonatanColussi:
> I am not sure I understand what you mean, could you write some code example to clarify please?

For exemple:
I emit a event with:
`{
    date: new Date()
}` 

and I receive this data
`{
    date: '2022-12-22T19:18:52.262Z'
}` 

Dates and another non-primitive types are converted to string

manast:
Ok, and how is this PR preventing that?

JonatanColussi:
This PR only inform to TS the correct data type

manast:
Ok, so if you pass a data object like ```{ date: new Date() }``` TS will produce an error or what happens?

PR #2505, opened 2022-12-24
Processors that return false values returning an empty object

notsapinho:


manast:
Although the tests are passing, there must be a reason why an empty object was added as default 🤔

notsapinho:
Yeah, I was wondering thy that `{}` was there, removing it doesn't seem to affect the lib.

Issue #2521, opened 2023-01-19
Enqueue job when id is same and active is running, nothing in the waiting.


royalpinto:
## Description

I have a scenario to explain the need for it.

Let's say I have a task which takes about 5 minutes to finish (ex: updating the db records one by one). This task may be triggered multiple times, but at a time only one should be running and subsequent triggers must be ignored (as I don't want to duplicate the work). One easy solution I found is that I ensure the `jobId` is same and all subsequent requests are automatically ignored.
Now the tricky part here is that, in the middle of the task execution, the db state might have changed and ignoring all the subsequent pushes will miss this state change to be handled. So I am expecting one extra job to be enqueued and all remaining ones should be ignored.
That is: *when active count > 0 and waiting count = 0, `add` should be allowed with same jobId* (as if when job is executing, it's not in the queue, so add one).

Looking at the references, it doesn't seems to be this way. it considers even the active ones while ensuring the jobId uniqueness.

## Minimal, Working Test code to reproduce the issue.
```
const Queue = require('bull')
const queue = new Queue('eventsua', 'redis://127.0.0.1:6379')
queue.process(async (job) => {
  console.log('===>>>> Job started execution', job.jobId)
  await new Promise((resolve) => setTimeout(resolve, 60000))
  console.log('Job finished execution', job.jobId)
})

queue.on('active', job => {
  console.info(`Job(${job.id}) started`, job.data)
})

queue.on('completed', (job, results) => {
  const duration = job.finishedOn - job.processedOn
  console.info(`Job(${job.id}) finished in ${duration}ms`, results)
})

queue.on('error', err => {
  console.error(err)
})

queue.on('failed', (job, err) => {
  console.info(`Job(${job.id}) failed`, err)
})

queue.on('stalled', (job) => {
  console.info(`Job(${job.id}) stalled`)
})

const test = async () => {
  for (let index = 0; index < 10; index++) {
    console.log('Pushing', index)
    const job = await queue.add({ event: 'refresh' }, { jobId: 'refresh', removeOnComplete: true })
    console.log('Pushed', index, job.id)

    await new Promise((resolve) => setTimeout(resolve, 2000))

    console.log('Waiting count: ', await queue.count())
    console.log('Job Counts', await queue.getJobCounts())
  }
}

test().catch(console.error)
```

I am expecting the task to run two times as I am pushing some tasks during the middle execution of the first execution and still they are being ignored for jobId being same.

[...]

PR #2525, opened 2023-01-23
Buffer when jobId is present but it is active

royalpinto:
Fix https://github.com/OptimalBits/bull/issues/2521

### DIFF
--- a/lib/commands/addJob-6.lua
+++ b/lib/commands/addJob-6.lua
@@ -33,6 +33,7 @@
       ARGV[9]  priority
       ARGV[10] LIFO
       ARGV[11] token
+      ARGV[12] buffer
 ]]
 local jobId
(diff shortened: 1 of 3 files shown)


lib/commands/addJob-6.lua
```diff
@@ -47,7 +48,15 @@ else
   jobId = ARGV[2]
   jobIdKey = ARGV[1] .. jobId
   if rcall("EXISTS", jobIdKey) == 1 then
-    return jobId .. "" -- convert to string
+    if ARGV[12] == '1' then
```
manast, line 51: So basically what this is doing is creating a new jobId (with postfix buffer) if the current jobId is being used. So in practice is like supporting 2 jobs with the same jobId instead of one... I am not sure I like this solution to be honest.

PR #2535, opened 2023-02-02
Fix: don't reschedule delay timer if closing

hendrikmoree:
It's sometimes possible the close of bull does not correctly stop the update timer. In some circumstances the update time is stil active after the close call is finished. 

Our process hangs on stopping if the timer is active. Tried multiple times with this fix, and our process always stops


hendrikmoree:
Awesome! Thanks

PR #2546, opened 2023-02-08
Fix job retry paused

levysam:


manast:
Thanks for the attempt. I will try to fix it today. This needs to be fixed at .lua script level.

PR #2576, opened 2023-04-01
Add new UI library to README.md

alexbudure:
Created a compact new UI library for Bull 😊


manast:
Sorry, but we will not promote competing products that erode the sources of income used for developing Bull and BullMQ.

alexbudure:
@manast understandable! I checked the last UI that was added and didn't see any requirements (#1819) but it's been a while 😂

Issue #2235, opened 2021-12-15
Issues with delayed queue jobs processing


harshit0410:
#### Background
We have a rate limit configured on a bull queue (100 jobs per minute). We had a case where the number of incoming jobs exceeded the rate limit (maybe 200 jobs per minute) and the jobs started getting accumulated in the delayed queue as expected. After this event, the number of incoming jobs returned to normal (20 jobs per minute) which is far less than the rate limit we had configured.

#### Problems
We started seeing a delay in the processing of new incoming jobs and we also noticed that the delayed queue always had some delayed jobs in it even though we had limits configured far higher than that of incoming jobs + jobs in the delayed queue. This means that all the incoming jobs are getting processed with some delay which shouldn't be the case as limits are far higher.

On further investigation, we found a case in the bull codebase where the incoming jobs get delayed by some amount of time if there are some jobs in the delayed queue. Due to this, it becomes a cycle where incoming jobs are being delayed due to the existing jobs in the delayed queue.

#### Steps to reproduce

- Configure a bull queue with rate limiter config (150 jobs per minute).
- Run two scripts that add jobs to the queue at the rate of(60 jobs per minute, 120 jobs per minute).
- Observe the delayed queue getting filled as expected.
- Kill the script that adds jobs at (60 jobs per minute).
- Observe the size of the delayed queue reducing first and then getting stabilized roughly at the same rate as incoming jobs.
- The delayed queue never gets empty due to which incoming jobs are delayed as well no matter how much time we give the queue to process delayed jobs. 
- Increasing the bull queue rate limit config(max jobs) does not help.

## Bull version
V3.29.2

manast:
Thanks for the reporting, I will try to reproduce this case following your instructions.

grakhejapstmn:
is there any update on this? its still happening for us.

manast:
I could reproduce it but I haven't had time to dig more into it.

grakhejapstmn:
I believe the root cause lies somewhere in this script, 
https://github.com/OptimalBits/bull/blob/develop/lib/commands/moveToActive-8.lua#L64

We are using the size of the limited queue to determine how much delay to add to the incoming jobs. 
So if there's a sudden spike and we hit the rate limit once, since there are jobs in the waiting queue that will be processed after a given delay and the current rate is less than the limit it leads to a loop where all incoming jobs go the waiting queue, and since more jobs come in before the queue is empty all new jobs again go the waiting queue and the cycle repeats. 

[...]

Issue #1110, opened 2018-11-01
Heap out of memory error on large number of queued jobs


ajwootto:
Hi,

I currently have around 2 million tasks queued in Bull which all need to be processed. I'm trying to test the performance of a single worker against this queue of tasks, but shortly after I start the worker I get the following heap memory error:

```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

<--- Last few GCs --->

[13:0x2d0a5c0]    82667 ms: Mark-sweep 947.4 (1508.8) -> 947.3 (1510.8) MB, 877.4 / 0.0 ms  allocation failure GC in old space requested
[13:0x2d0a5c0]    83565 ms: Mark-sweep 947.3 (1510.8) -> 947.2 (1456.8) MB, 897.2 / 0.0 ms  last resort GC in old space requested
[13:0x2d0a5c0]    84474 ms: Mark-sweep 947.2 (1456.8) -> 947.2 (1436.3) MB, 909.3 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x142c70c25879 <JSObject>
    1: _settlePromise [/src/node_modules/bluebird/js/release/promise.js:~542] [pc=0xac62245a421](this=0x38989dfe5681 <Promise map = 0x3f8459124389>,promise=0x38989dfed891 <Promise map = 0x3f8459124389>,handler=0x1756049822d1 <undefined>,receiver=0x1756049822d1 <undefined>,value=0x173509168221 <Number 1.5411e+12>)
    2: _drainQueue(aka _drainQueue) [/src/node_modules/bluebird/js/release/async.js...

 1: node::Abort() [node /src/dist/app.js]
 2: 0x8cbf4c [node /src/dist/app.js]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node /src/dist/app.js]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node /src/dist/app.js]
 5: v8::internal::Factory::NewUninitializedFixedArray(int) [node /src/dist/app.js]
 6: 0xd801bc [node /src/dist/app.js]
 7: 0xd97a95 [node /src/dist/app.js]
[...]

PR #2580, opened 2023-04-14
fix: rate limiter improvement

andreas:
As described in https://github.com/OptimalBits/bull/issues/2235, the current rate limiter will "lag behind" on processing of jobs once the rate limit has been exceeded. This is because the current implementation will throttle jobs purely based whether there are any delayed jobs, and attempts to "space" out jobs according to the set `max` and `duration`. However, this strategy means that if new jobs come in while processing jobs that have been delayed due to rate limiting, these jobs will be rate limited also, even if the new jobs would not result in the rate limit being exceeded.

The diagram below tries to depict this scenario:

The proposed solution instead stacks up delayed jobs at the beginning of the next rate limiter "windows".

The current tests do not expose this behaviour, as this bug does not surface with `max = 1`. I've included tests with combinations of `max`, `duration` and job counts (`numJobs`), which fail with the current implementation, but pass with the proposed solution. They rely on the fact that processing `numJobs` for a specific `max` and `duration` should be processed in a duration from`(Math.ceil(numJobs/max) - 1) * duration)` to `Math.ceil(numJobs/max) * duration`.

### DIFF
--- a/lib/commands/moveToActive-8.lua
+++ b/lib/commands/moveToActive-8.lua
@@ -68,7 +68,7 @@ local rateLimit = function(jobId, maxJobs)
 
     if numLimitedJobs > 0 then
       -- Note, add some slack to compensate for drift.
-      delay = ((numLimitedJobs * ARGV[7] * 1.1) /  maxJobs) + tonumber(rcall("PTTL", rateLimiterKey))
+      delay = tonumber(rcall("PTTL", rateLimiterKey)) + ARGV[7] * math.floor(numLimitedJobs / maxJobs)
     end
   end
 
@@ -80,7 +80,7 @@ local rateLimit = function(jobId, maxJobs)
   if (delay == 0) and (jobCounter >= maxJobs) then
     -- Seems like there are no current rated limited jobs, but the jobCounter has exceeded the number of jobs for this unit of time so we need to rate limit this job.
     local exceedingJobs = jobCounter - maxJobs
-    delay = tonumber(rcall("PTTL", rateLimiterKey)) + ((exceedingJobs) * ARGV[7]) / maxJobs
+    delay = tonumber(rcall("PTTL", rateLimiterKey)) + ARGV[7] * math.floor(exceedingJobs / maxJobs)
   end
 
   if delay > 0 then
--- a/test/test_rate_limiter.js
+++ b/test/test_rate_limiter.js
@@ -201,6 +201,56 @@ describe('Rate limiter', () => {
     });
   });
 
+  [
+    { max: 1, numJobs: 3, duration: 100 },
+    { max: 2, numJobs: 11, duration: 100 },
+    { max: 3, numJobs: 16, duration: 100 },
+    { max: 10, numJobs: 51, duration: 100 },
+    { max: 100, numJobs: 501, duration: 500 }
+  ].forEach(({ max, numJobs, duration }) => {
+    it(`should obey the rate limit with max = ${max}, duration = ${duration}ms, and numJobs = ${numJobs}`, done => {
+      const startTime = new Date().getTime();
+      let completedJobs = 0;
+      let enqueuedJobs = 0;
+
+      const newQueue = utils.buildQueue(`test rate limiter with ${max}`, {
+        limiter: {
+          max,
+          duration
+        }
+      });
+
+      newQueue.on('failed', e => {
+        assert.fail(e);
+      });
+
+      newQueue.process(() => {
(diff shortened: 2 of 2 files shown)


manast:
I appreciate this PR, however, I am struggling to understand what is that it is solving. 🤔

andreas:
@manast, consider the following two qualities of a rate limiter:

1. A rate limiter should make full use of the specified rate limit, i.e. if there are more than `max` jobs enqueued, then `max` jobs should be processed every `duration` (assuming processing time is not a bottleneck).
2. A rate limiter should not delay the processing of jobs unnecessarily. A rate limited job should be processed as quickly as possible.

The current rate limiter do not fully have these qualities. Specifically, it does not make full use of the specified rate limit, and jobs are delayed unnecessarily. This is what's described in #2235, and we experience the same thing in production for our workload. Essentially jobs are moved into the delayed state, despite the queue processing jobs below the rate limit.

Below is a scenario that shows the problem (admittedly designed to make matters worst). It ignores the `1.1` drift compensation factor to the time calculations simpler.

1. Consider a rate limiter `max=3` and `duration=60*60*1000` (1 hour), where each job takes 1m, and jobs are being processed serially (concurrency 1). Let's call the current time `t0`.
2. `t0`: Enqueue 5 jobs. Let's call them `j1` through `j5`.
3. `t0 + 0h1m`: `j1` completes.
4. `t0 + 0h2m`: `j2` completes.
5. `t0 + 0h3m`: `j3` completes. The rate limit is now exceeded, so `j4` is delayed until `t0 + 1h0m` and `j5` is delayed until `t0 + 1h20m`.
7. `t0 + 1h1m`: `j4` completes. Now, enqueue job `j6`, which will be delayed until `t0 + 2h0m` because `j5` is delayed.

In conclusion, `j5` is processed at `t0 + 1h20m`, but could be processed at `t0 + 1h1m`, so processing is delayed 19 min more than necessary. Further, `j6` is processed at `t0 + 2h0m`, but could be processed at `t0 + 1h2m`, so processing is dealyed 58 min more than necessary 😨

manast:
Thanks for the explanation. Ok so the issue is due to an optimization performed several years ago to minimize jobs being delayed too little and ending in a loop where they are rate-limited, delayed, go back to wait, then rate limit again, and so on, which in high load scenarios would make things really bad. So your fix reverts to the old implementation... and I am not sure this is good. The main issue here is that due to features like group keys for rate-limited jobs, it is not possible to actually "pause" the queue when it is rate limited, which would be the optimal way to solve this issue, i.e. as soon as the queue is rate limited, no jobs are processed at all, not even moved to the delay set, instead keep them waiting until the rate limiter expires. This is how rate-limiter is done in BullMQ now, but we had to sacrifice group keys, which is now a feature of the Pro version but properly implemented with virtual queues.

So all in all, I am not sure this could be merged even if it will solve your particular case, it may actually even be the case you will suffer from what I exposed above if you get high loads in the future.

andreas:
@manast, thanks for the additional context. I could be mistaken, but I think there are key differences from the proposed implementation to the previous versions:

1. The version of the rate limiter prior to #1212 appears to only delay jobs to the next limiter window when `jobCounter >= maxJobs`. Here `jobCounter` refers to the number of jobs processed in the current window. This implementation exhibits the problems you mention about rate limited jobs being rate limited over and over again (issue #1110).
2. After applying #1212, rate limited jobs are "spaced out" based on the number jobs processed in the current time window, i.e. `(jobCounter - maxJobs) * duration / max`. I assume this mitigates the re-processing problem of #1110.
3. After applying #1816, rate limited jobs are "spaced out" based on the _total_ number of rate limited jobs and it adds the drift compensation factor, i.e. `1.1 * (numLimitedJobs - maxJobs) * duration / max`.

While it may appear that the proposed solution is similar "version 1", note that it uses the _total_ number of rate limited jobs for delaying, not just the jobs processed in the current time window. This is a key difference, which I believe avoids the problem re-processing problem of #1110: jobs will be "spaced out" like for version 3, they are just delayed to the beginning of a limiter window, rather than being "distributed" within the windows. This has the benefit of avoiding unnecessary delays and better utilizing the full rate limit.

In summary, I believe the proposed solution has the qualities I mentioned in my above comment _and_ it avoids the problem of processing rate limited jobs over and over again. That being said, you're the expert, so I may very well have missed something 🙂

manast:
A couple of tests related to this change are failing.

andreas:
Thanks, I'll take a look!

PR #2587, opened 2023-04-26
Fix: ts declaration

GuyBerko:
The metrics option and the getMetrics function are missing from the typescript declaration which causes ts errors in projects that use bull. plus the backoff strategyOptions are named "option" in the PATTERNS and in the REFERENCE but in the ts declaration they are called "strategyOptions" which also causes ts to be unhappy.


manast:
Thanks for the contribution!

PR #2622, opened 2023-07-04
Fix return type of getJobCountByTypes

Benjythebee:
The return type of `getJobCountByTypes(types: JobStatus[] | JobStatus)` is invalid;
It is a number, not a JobCounts;

Another proof is that function `count()` uses `getJobCountByTypes` and yet returns a number;


manast:
Thanks for the PR. Indeed it was wrong in the typings.

PR #2634, opened 2023-07-29
queue.add: deep clone opts #2633

### LINKED ISSUE #2633: Side effect causing jobs to have same `jobId`
## Description
In case the same `repeat` object is reused across several jobs, all the jobs result in having the same `jobId` in redis, hence only one of the created jobs gets executed (and has a proper entry in redis). 
Subjectively, I, as a user of the library, consider this a bug.

See #2634 for resolution

```
'use strict'

const Bull = require('bull')
const redis = require('redis')
const { promisify } = require('util')

const testSchedule = { cron: '*/10 * * * *' }

async function main () {
  const client = redis.createClient()
  const hgetAsync = promisify(client.hget).bind(client)
  const keysAsync = promisify(client.keys).bind(client)
  const queue = new Bull('my-shiny-queue-1')

  await Promise.all([
    queue.add(
      'my-shiny-job-1',
      {},
      {
        jobId: 'my-shiny-job-1',
        repeat: testSchedule,
        removeOnComplete: true
      }
    ),
    queue.add(
      'my-shiny-job-2',
      {},
      {
        jobId: 'my-shiny-job-2',
        repeat: testSchedule,
        removeOnComplete: true
      }
    )
  ])

  const myShinyJobsList = await queue.getRepeatableJobs()
  console.log('job confs returned by bull', myShinyJobsList)

  const qKeys = await keysAsync('bull:my-shiny-queue-1:repeat:*')
  const realRedisData = await Promise.all(qKeys.map(k => hgetAsync(k, 'opts')))
  console.log('repeat conf from redis:')
  for (const rawStr of realRedisData) {
    const obj = JSON.parse(rawStr)
    console.log('key: %s, jobId: %s', obj.repeat.key, obj.repeat.jobId)
  }
}

main()
  .catch(console.error)
  .finally(() => process.exit())
```

## Bull version
4.10.2

## Additional information

script output:
```
job confs returned by bull [
  {
    key: 'my-shiny-job-2:my-shiny-job-2:::*/10 * * * *',
    name: 'my-shiny-job-2',
    id: 'my-shiny-job-2',
    endDate: null,
    tz: null,
    cron: '*/10 * * * *',
    every: null,
    next: 1690590600000
  },
  {
    key: 'my-shiny-job-1:my-shiny-job-1:::*/10 * * * *',
    name: 'my-shiny-job-1',
    id: 'my-shiny-job-1',
    endDate: null,
    tz: null,
    cron: '*/10 * * * *',
    every: null,
    next: 1690590600000
  }
]
repeat conf from redis:
key: my-shiny-job-1:my-shiny-job-1:::*/10 * * * *, jobId: my-shiny-job-2
key: my-shiny-job-2:my-shiny-job-2:::*/10 * * * *, jobId: my-shiny-job-2
```

Although queue returns correct data, the real data in redis is incorrect.

This happens due to the fact that the shared `repeat` object is mutated.

batrudinych:
fixes #2633 

p.s.: cloning only `repeat` when calling the `nextRepeatableJob` would also work if full deep `opts` clone is not an option.

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -732,7 +732,7 @@ Queue.prototype.add = function(name, data, opts) {
     data = name;
     name = Job.DEFAULT_JOB_NAME;
   }
-  opts = { ...this.defaultJobOptions, ...opts };
+  opts = _.cloneDeep({ ...this.defaultJobOptions, ...opts });
 
   opts.jobId = jobIdForGroup(this.limiter, opts, data);
 


manast:
Thanks for the fix!

PR #2640, opened 2023-08-03
fix: add mising getMetrics type

alko89:
It seems typings are missing the getMetrics method definition.

https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queuegetmetrics


manast (approved):
Looks good, thanks.

PR #2641, opened 2023-08-05
chore: update ubuntu build version

alko89:
Ubuntu 18.04 is deprecated. Didn't test this, but since the action doesn't use any os specific actions, like installing packages, this should work just fine.

image


manast:
Thanks!

Issue #1414, opened 2019-08-04
Queue close does not close connections if it is called when is redis offline [BUG]


misos1:
## Description

Seems when is redis offline then bull is trying to reconnect and in this state queues cannot be closed. When is queue closing it should cancel reconnecting. Also seems if are queues created when is redis offline they will never actually connect after is redis run.

## Minimal, Working Test code to reproduce the issue.
``` node js
let Queue = require("bull");
let queue = new Queue("queue");
queue.process(_job => console.log("started"));
queue.on("error", err => console.log("queue error:", err));
queue.close();
```

When is redis running this test code gracefully ends. When is redis not running then this will emit bunch of errors and never ends even after I run redis again:
```
queue error: Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1056:14) {
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379
}
```
I also tried to run it without `queue.close` while was redis running, then turn off redis and call `queue.close` and then turn on redis again and it does not closes.

When I try to run queues, then turn off redis it will print error messages, then turn on redis again and call `queue.close` it will gracefully end.

## Bull version
3.10.0

theclive:
I am experiencing something similar, if I subscribe to the client events and pass in my own instance of ioredis, upon `queue.close()`, I experience the following debug output:

```
2019-10-08T14:05:20.693Z - debug:    Redis queue: Disconnected from Redis
2019-10-08T14:05:20.693Z - debug:    Redis queue: Connection to Redis has been terminated, and no more reconnections will be made
2019-10-08T14:05:20.697Z - debug:    Redis queue: Connected to Redis
```

Here is the code for creating my queue connection:
```javascript
    const connection = new RedisParam(getConnectionOptionsParam(logMsgPrefix))

    connection.on('connect', () => {
      logger.log('debug', `${logMsgPrefix}: Connected to Redis`)
    })

    connection.on('error', error => {
      logger.log(
        'error',
        `${logMsgPrefix}: Error in Redis client: %s`,
        error.message,
        { error }
      )
    })

    connection.on('end', () => {
      logger.log(
        warnOnDisconnect ? 'warn' : 'debug',
        `${logMsgPrefix}: Connection to Redis has been terminated, and no more reconnections will be made`
      )
    })

    connection.on('close', () => {
[...]

PR #2673, opened 2023-10-14
fix: catch pause errors when closing

manast:
should address issue #1414


misos1:
Maybe you could just use `finally` instead of `then` in appropriate places.

manast:
maybe, but I do not want to make more changes than the strictly necessary as Bull has reached EOL.

PR #2675, opened 2023-10-19
feat: add missing extendLock definition and its docs

Smankusors:
feat: add missing extendLock definition
also:
* update lockKey definition with the source code
* add missing references for job locks documentation


Smankusors:
Eh, please don't.

manast:
This looks good to me, however the danger with it is that people will think that they should normally use these methods, which basically should only be used internally unless you know very well what you are doing.

PR #2685, opened 2023-11-30
fix(docs): correct misuse of Error constructor

sdjdd:
Correct the misuse of the Error constructor in the documentation. The second parameter should not be a string.
Link: https://nodejs.org/api/errors.html#class-error


manast:
Thanks!

PR #2716, opened 2024-03-25
fix: avoid race condition when reconnecting

martinslota:
# Background

We have been seeing occasional `Connection is aborted` errors when pausing a Bull queue that is backed by Redis in cluster mode. The investigation led to [this fix in `ioredis`](https://github.com/redis/ioredis/pull/1864). After deploying that fix, the errors became a lot more rare, but they still pop up from time to time - sometimes days apart.

The debug logs from `ioredis` led me to believe that the issue may be around how `bull` [re-establishes the blocking connection when `pause()` is called](https://github.com/OptimalBits/bull/blob/60fa88f08637f0325639988a3f054880a04ce402/lib/queue.js#L868-L871). Specifically, the promise returned by `redisClientDisconnect` may get resolved before [this disconnect() call](https://github.com/OptimalBits/bull/blob/60fa88f08637f0325639988a3f054880a04ce402/lib/queue.js#L562) is executed, causing `ioredis` to abort its already ongoing connection attempt.

Attempts to reliably reproduce the bug have led to [this repository](https://github.com/martinslota/bull-pause-resume-bug) where `ioredis` is intentionally modified to slightly delay the resolution of the promise returned from the Redis cluster client's `quit()` method in order to provoke the bug into occurring. All that is then necessary is to create a queue and [repeatedly pause/resume it](https://github.com/martinslota/bull-pause-resume-bug/blob/6e27cc86e7c48cf4574f7f59a3fe23801ff6e0ee/index.js#L31-L39).

# Change

Avoids calling `client.disconnect()` within `redisClientDisconnect` if the promise is already resolved at that point.


martinslota:
I'd love to keep this pull request alive because the bug it fixes is affecting our production deployments and it generates quite a bit of noise.

@manast I apologize for cold-tagging you here - I'd appreciate any pointers or suggestions regarding how I could take this pull request further.

martinslota:
I’m still hopeful that this bug fix will eventually get some attention from a maintainer.

In the meantime, I will make sure to deploy the fix as part of our code to gain additional confidence. I will report again when it has been out for a while.

manast:
@martinslota have you checked how connections are handled in BullMQ? I have spent quite a lot of time trying to iron out all connection issues, including workarounds for ioredis bugs when blocking commands gets disconnected: https://github.com/taskforcesh/bullmq/blob/master/src/classes/redis-connection.ts

martinslota:
@manast No, I haven't - thank you for the reference! 🙇 

We have just adopted this small fix in our code and it will be going to production today.

Would it be possible to backport the connection fixes to Bull? Or would your recommendation be to migrate to BullMQ? We have not explored what it would take to do but I think having fewer random connection errors could make it worth the effort.

manast:
> Would it be possible to backport the connection fixes to Bull? Or would your recommendation be to migrate to BullMQ? We have not explored what it would take to do but I think having fewer random connection errors could make it worth the effort.

Actually no as the way connections are handled in Bull is quite different so it is not trivial to do.

martinslota:
We have had this fix in production for around 2 months now and it has really helped reduce the amount of non-actionable Redis connection errors during shutdown (when we pause Bull queues).

Perhaps it could make sense to get the fix merged.

Issue #2723, opened 2024-04-17
Stuck should not be separate from JobStatus


jdforsythe:
This `JobStatus | 'stuck'` requires us to use that in our codebase.

https://github.com/OptimalBits/bull/blame/60fa88f08637f0325639988a3f054880a04ce402/index.d.ts#L261

```ts
const stateProms: Promise<JobStatus | 'stuck'>[] = jobs.map((j) => j.getState());
```

However, 'stuck' is one of the valid job statuses, according to the docs and code, so why isn't it just a part of the `JobStatus` union?

https://github.com/OptimalBits/bull/blob/develop/index.d.ts#L358-L371

Then our code can be simplified:

```ts
const stateProms: Promise<JobStatus>[] = jobs.map((j) => j.getState());
```

manast:
There is no such thing as "stuck" status in Bull. When a job stalls, it goes back to wait, it cannot be in any other state other than the ones provided by the typescript definition.

jdforsythe:
@manast 

The Reference says otherwise:

https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#jobgetstate

As does the presence of this function:

https://github.com/OptimalBits/bull/blob/develop/lib/job.js#L425-L429

As well as the `getState()` function:

https://github.com/OptimalBits/bull/blob/develop/lib/job.js#L435-L459

jdforsythe:
@manast because we have to keep adding it to our code to get the typing correct. We'd be fine if it were removed completely. We don't care about `stuck` but we're forced to add it to the type every time we get a job state.

I'd be happy to submit a PR to remove it, but what *should* `getState()` return if it's not in any of the lists?

manast:
A job must be in some list otherwise it would be a bug in bull, or the job has been removed and is not in the queue at all anymore.

jdforsythe:
@manast I submitted a PR to remove "stuck":

https://github.com/OptimalBits/bull/pull/2725

PR #2725, opened 2024-04-19
fix: remove stuck as a job state

jdforsythe:
Removes `stuck` as a valid job state.

Removes the function to check, removes the return of 'stuck' in `getState()`, updates docs and tests to remove "stuck".`


jdforsythe:
According to #2723 "stuck" is not a valid state and this should be merged to remove it

@manast

PR #2730, opened 2024-05-04
fix(job): validate jobKey in updateProgress and update

roggervalf:
Using updateProgress and update method after a job is removed, re-add a jobKey but only with data or progress attribute


manast (approved):
Looks good.

PR #2731, opened 2024-05-04
fix(scripts): remove from stalled after moving to wait or delayed

roggervalf:


manast (approved):
LGTM, this seems like a quite serious bug actually.

Issue #1034, opened 2018-08-27
Debounced Jobs


tophep:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%240%20Funded-%2300A156.svg" />](https://issuehunt.io/r/OptimalBits/bull/issues/1034)

## Description
Let's say I want to queue a job every time one of my users updates their profile. But if a user makes many rapid updates I don't want to queue a bunch of jobs that repeat the same work. I'd rather wait until the user is done with the flurry of updates.

This was [previously discussed here](https://github.com/OptimalBits/bull/issues/855#issuecomment-387118635) but it was unrelated to the main thread and there was no resolution.

A simple solution could be to use a custom Id with a delay. When another job needs to be queued, delete the old job and re-queue with delay again. The caveat (touched on in thread above) is when the job is active and being processed and another job needs to be queued. @manast Any thoughts on how to handle this edge case? cc @lxcid

Separately, would be convenient to increase the delay of existing jobs (instead of delete + enqueue).

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/OptimalBits/bull/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

manast:
Well I guess that should be like any other debouncing, i.e. if the job already started it will continue working, and the job added after that should be either ignored or queued.

sebasmagri:
Any chance to look at this again?

This would solve a very common use case I have. A possible solution I've been looking at is to have deduplication set up on `jobId` + `state`, so that if there's a job already `waiting`, there will not be new `waiting` jobs for the same `jobId`, but if there's a task already `active` for the same `jobId` we can still create a task that will be executed after that.

This requires concurrence to be 1, however, and doesn't consider failure/completion states. Also would require some additional checks.

Have also looked at a double queue setup, one for debouncing and the second for the actual processing, but still sounds very hacky.

So a _native_ solution for this would be perfect.

Thoughts? Any direction in which we could help?

shivgarg5676:
Did you guys find any good solution to this problem? I am facing the same situation where the same job can be queued multiple times and I need to process one. Currently, I am storing all the jobs in the Redis myself and batching them up and sending one job in the queue after 10 seconds. But a native solution could be better.

[...]

PR #2760, opened 2024-07-30
feat(job): support debouncing

roggervalf:
ref https://github.com/OptimalBits/bull/issues/1034


lib/commands/obliterate-2.lua
```diff
@@ -26,10 +30,11 @@ local function getZSetItems(keyName, max)
     return rcall('ZRANGE', keyName, 0, max - 1)
 end
 
-local function removeJobs(parentKey, keys)
```
roggervalf, line 29: parentKey is wrong, this is an issue in this script that is being fixed in this pr

Issue #2325, opened 2022-03-18
`redisOptsFromUrl` does not set `tls` to true for `rediss://` urls


skylander86:
## Description

When working with TLS encrypted redis connections (e.g., Elasticache), the Redis connection fails to connect because the `tls` attribute on the redis option during connecting is not set.

## Minimal, Working Test code to reproduce the issue.

```javascript
const q = new Queue("helloworld", "rediss://:password@hostname:6379/5")

console.log(q.client.status)  // this is stuck at `connect` state
```

## Bull version

4.70

vpetlu:
+1
@skylander86 did you find a workaround?

vpetlu:
> +1 @skylander86 did you find a workaround?

Found a way to specify the tls flag after digging into the queue constructor:
```
const myFirstQueue = new Queue('my-third-queue', 'rediss://<url>',
    { redis: { tls: true, enableTLSForSentinelMode : false } }
  );
```

craigotis:
For anyone finding this, if you're on Bull 4, it's worth noting that while Bull doesn't properly handle `rediss://` for TLS connections, it _does_ spread the query parameters into the options object. Making code changes is one approach, but you can also add:
```
...?tls=true
```
To the end of your connection string, and assuming you're passing it directly into Bull (ie. not using the separate host/username/password options), it should pass that into IORedis and connect successfully.

emanuelegorga:
Hey @craigotis , I am on Bull 4.9.0 and unfortunately the `redis://xxxx.com?tls=true` did not work out. Any idea why that might be the case? 🤔

craigotis:
@emanuelegorga In what ways does it fail? Can you try with both `rediss://` _and_ `?tls=true`?

emanuelegorga:
It hangs when trying to enqueue the first job. The same behaviour happens locally if I purposely use an invalid host which will result in Bull not being able to connect to Redis. 
I am sure about that the Redis host I am passing to the `production` environment is correct, so I am assuming that the only problem left can only be the TLS connection also considering that locally it works just fine when passing the valid local Redis instance.

One thing that makes me doubt about the `tls=true` parameter is the documentation on Redis here: https://github.com/luin/ioredis#tls-options

As I understand, the parameter is supposed to be taken and passed as option to the Redis instance, however, the documentation shows that the options related to TLS are supposed to go within an object instead of a boolean value like `tls=true`.

craigotis:
[...]

PR #2776, opened 2024-10-04
feat(queue): handle redis url starting with rediss scheme by specifying tls as empty object

Otanikotani:
We believe this is a potential in house fix for https://github.com/OptimalBits/bull/issues/2325 . 
Our setup:
- AWS Elasticache 6.2.6 with TLS enabled.
- Nestjs app that uses `bull`. 
- The configuration for Redis is set through a Redis url which looks like: `rediss://:banana@master.use1.cache.amazonaws.com?tls=true`

We have been experiencing intermittent 
```
Error: Connection is closed.
    at close (/usr/src/app/node_modules/ioredis/built/redis/event_handler.js:189:25)
    at TLSSocket.<anonymous> (/usr/src/app/node_modules/ioredis/built/redis/event_handler.js:156:20)
    at Object.onceWrapper (node:events:634:26)
    at TLSSocket.emit (node:events:531:35)
    at TLSSocket.emit (/usr/src/app/node_modules/dd-trace/packages/datadog-instrumentations/src/net.js:69:25)
    at node:net:339:12
    at TCP.done (node:_tls_wrap:657:7)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```

We think that the following issue is somehow related: https://github.com/redis/ioredis/issues/1628. We do not experience "Connection is closed." error if we disable TLS on our Elasticache Redis. 

When we switched from using a Redis url to providing Redis options like this:
```typescript
const opts: RedisOptions = {
      host: this.host,
      port: this.port,
      db: this.database,
      password: this.password,
};
if (this.scheme === 'rediss') {
      opts.tls = {};
}
if (this.username) {
      opts.username = this.username;
}
return opts;
```
the error went away. So we would like to bring the same logic to the `getOptsFromUrl` function. If the protocol is [rediss](https://www.iana.org/assignments/uri-schemes/prov/redis) then the `tls: {}` is added to the Redis options. 

Kudos to:
@hype08
@joshgrift

### DIFF
--- a/lib/queue.js
+++ b/lib/queue.js
@@ -346,6 +346,10 @@ function redisOptsFromUrl(urlString) {
       }
     }
 
+    if (redisUrl.protocol && redisUrl.protocol.startsWith('rediss')) {
+      redisOpts.tls = {};
+    }
+
     if (redisUrl.query) {
       redisOpts = { ...redisOpts, ...redisUrl.query };
     }
--- a/test/test_queue.js
+++ b/test/test_queue.js
@@ -223,6 +223,16 @@ describe('Queue', () => {
       return queue.close();
     });
 
+    it('creates a queue using the supplied redis url that contains rediss protocol', () => {
+      const queue = new Queue('custom', {
+        redis: 'rediss://abc:123@127.2.3.4:1234/1'
+      });
+
+      expect(queue.client.options.tls).to.be.eql({});
+
+      return queue.close();
+    });
+
     it('creates a queue using the supplied redis host', () => {
       const queue = new Queue('custom', { redis: { host: 'localhost' } });
 


mbackermann:
We should definitely have this option going forward

lib/queue.js
```diff
@@ -346,6 +346,10 @@ function redisOptsFromUrl(urlString) {
       }
     }
 
+    if (redisUrl.protocol && redisUrl.protocol.startsWith('rediss')) {
+      redisOpts.tls = {};
+    }
```
Stasmo, line 351: Good change. Default should be to use TLS for `rediss` protocol, and you can still disable that with query string options like `tls=false`.

Issue #2782, opened 2024-11-01
Encountering RangeError [ERR_BUFFER_OUT_OF_BOUNDS] When Creating Jobs in Bull on Node.js v23


aryamohanan:
## Description
We are encountering an error in Bull on Node.js v23. while attempting to create jobs, which results in the following error message:

> RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: "length" is outside of buffer bounds  
>     at Buffer.utf8Write (node:internal/buffer:1066:13)  
>     at Packr.encodeUtf8 (node_modules/msgpackr/dist/node.cjs:1202:18)  
>     at pack (node_modules/msgpackr/dist/node.cjs:1491:15)  
>     at Packr.pack.encode (node_modules/msgpackr/dist/node.cjs:1300:6)  
>     at Object.addJob (node_modules/bull/lib/scripts.js:40:7)  
>     at addJob (node_modules/bull/lib/job.js:83:18)  
>     at node_modules/bull/lib/job.js:97:14 {  
>   code: 'ERR_BUFFER_OUT_OF_BOUNDS'  
> }

The error appears to originate from the [msgpackr](https://www.npmjs.com/package/msgpackr) library, which Bull utilizes for data serialization.

## Bull Version
4.16.3

## Node.js Version
v23

## Additional Information
This issue seems to be resolved in msgpackr version 1.1.2. You can find more details in the [commit](https://github.com/kriszyp/msgpackr/commit/6b18367f1f6883550452991c7170cc31aa4d1d9d#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R4). 
Please refer to the related issue [here](https://github.com/kriszyp/msgpackr/issues/144).

PR #2783, opened 2024-11-01
fix: bump msgpackr to 1.1.2 to resolve ERR_BUFFER_OUT_OF_BOUNDS error

aryamohanan:
This PR updates the `msgpackr` dependency to version 1.1.2 in order to resolve a RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: "length" is outside of buffer bounds error that occurs during job creation. The error originates from the msgpackr library, which is used by Bull for data serialization. 
Upgrading to msgpackr version 1.1.2 includes a fix that addresses this issue.

ref https://github.com/OptimalBits/bull/issues/2782

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "get-port": "^5.1.1",
     "ioredis": "^5.3.2",
     "lodash": "^4.17.21",
-    "msgpackr": "^1.10.1",
+    "msgpackr": "^1.11.2",
     "semver": "^7.5.2",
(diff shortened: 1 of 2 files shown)


roggervalf:
Thank you. For reference looks like  https://github.com/kriszyp/msgpackr/pull/151 this was the fix

manast:
We should also add NodeJS 23 then to the test node matrix.

PR #2785, opened 2024-12-18
fix: upgrade cron-parser dependency for Luxon CVE-2023-22467

carvilsi:
Upgrades `cron-parser` dependency for Luxon: Inefficient Regular Expression Complexity vulnerability [CVE-2023-22467](https://github.com/advisories/GHSA-3xq5-wjfh-ppjc)


manast:
Thanks for the fix!

PR #2823, opened 2026-09-20
fix(job): emit Job instances from pattern-based remove

### LINKED ISSUE #2115: 'removed' event inconsistent parameter
According to REFERENCE.md:
```
.on('removed', function (job) {
  // A job successfully removed.
});
```

In job.js this event is emited with job id here:

```
Job.remove = async function(queue, pattern) {
  await queue.isReady();
  const removed = await scripts.removeWithPattern(queue, pattern);
  removed.forEach(jobId => queue.emit('removed', jobId));
};
```

And with the full Job object here:

```
Job.prototype.remove = function() {
  const queue = this.queue;
  const job = this;

  return queue.isReady().then(() => {
    return scripts.remove(queue, job.id).then(removed => {
      if (removed) {
        queue.emit('removed', job);
      } else {
        throw new Error('Could not remove job ' + job.id);
      }
    });
  });
};

```

dyk1454683243-sudo:
## Summary

Fixes #2115 — the `removed` event was inconsistent:

- `Job.prototype.remove` emitted a **Job instance**
- `Job.remove` / `Queue#removeJobs` (pattern remove) emitted a **job id string**

`REFERENCE.md` and `RemovedEventCallback` document a Job instance. Both paths now emit that shape.

Pattern remove snapshots the Redis job hash in `removeJobs` before `DEL`, then reconstructs a `Job` via `Job.fromJSON` so listeners still receive `id`, `name`, and `data`.

## Changes

- `lib/commands/removeJobs-8.lua`: `HGETALL` before delete; return job hashes as JSON
- `lib/scripts.js` / `lib/job.js`: decode hashes and emit `Job` instances
- Docs/types: document that both remove paths emit a `Job`
- Tests: regression coverage for `Job#remove` and `Queue#removeJobs`

## Notes

This is a bugfix against the documented API. Anyone already treating pattern-remove payloads as strings will need to use `job.id`.

Locked jobs are still skipped and do not emit.

Fixes #2115


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

